PR #8, opened 2021-07-19
Word encoding and comparator

### LINKED ISSUE #7: Add the word encoding
### What's wrong

We currently have some word encoding methods almost ready.

To make sure they are well defined, we should add a script to feed some values and test the idea.

### How are we fixing it

Add Python script and some test cases.

ChihChengLiang:
### What's wrong

Fix #7 

### How are we fixing it

We make the [word encoding](https://hackmd.io/@liangcc/zkvmbook/https%3A%2F%2Fhackmd.io%2FQUiYS3MnTu29s62yg9EtLQ) and the [comparator](https://hackmd.io/6jzBI4nmRR-JvSG2TFHrDw) spec executable and tested


src/encoding/comparator.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    assert len(a8s) == len(b8s) == 32
+    assert len(result) == 16
+
+    # Before we do any comparison, the previous result is "equal"
+    result = result[:] + [0]
+
+    for i in reversed(range(0, 32, 2)):
+        a16 = a8s[i] + 256 * a8s[i + 1]
+        b16 = b8s[i] + 256 * b8s[i + 1]
```
ChihChengLiang, line 43: Thank @han0110 for catching the bit shift bug, which is fixed in https://github.com/appliedzkp/zkevm-specs/pull/8/commits/5d500a57f19aff9cf436329c584f93cd9b911d57

han0110 (approved):

PR #9, opened 2021-07-20
add addition checks

ChihChengLiang:
### What's wrong

Adding addition arithmetics

PR #11, opened 2021-07-20
Tableless addition checks

ChihChengLiang:
### What's wrong

Turns out we don't even need table lookup to do 8-bit addition like what we did in #9

### How are we fixing it

Replace the table lookup with the simple linear combinations

### DIFF
--- a/src/encoding/addition.py
+++ b/src/encoding/addition.py
@@ -11,6 +11,11 @@ def check_add(
     sum8s: Sequence[U8],
     carry32: Sequence[bool],
 ):
+    """
+    Check the addition of two length 32 8-bit chunks a8s and b8s
+    Items of a8s and b8s should already be verified to be 8 bits before.
+    Overflow is allowed to match the evm behavior
+    """
     assert len(a8s) == len(b8s) == len(sum8s) == 32
     assert len(carry32) == 32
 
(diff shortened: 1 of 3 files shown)


src/encoding/addition.py
```diff
@@ -0,0 +1,23 @@
...
+):
+    assert len(a8s) == len(b8s) == len(sum8s) == 32
+    assert len(carry) == 32
+
+    # Before we add anything yet, the carry bit should be 0
+    # We allow the last carry bit to be 1 to support the overflow case
+    carry = [0] + carry[:]
+
+    for i in range(32):
+        assert carry[i + 1] * (carry[i + 1] - 1) == 0, "carry should be 0 or 1"
```
han0110, line 22: Should we check from `0..31`?
```suggestion
        assert carry[i] * (carry[i] - 1) == 0, "carry should be 0 or 1"
```

ChihChengLiang: Since we add a [0] to carry at line 19 so that the carry have length 33 and thus we check 1..32. 

But that's confusing indeed, I shouldn't mutate the input argument carry. Let me fix it.

han0110: Sorry I just noticed on line 76 which makes carry to length 32 and totally missed line 19... It makes sense now

han0110 (approved):

PR #12, opened 2021-07-20
add commitment check

ChihChengLiang:
### What's wrong

We haven't described the commitment check yet, which checks 
- if an array x8s is an opening of a u256 commitment and 
- if elements in the x8s are all 8 bits.

### How we are fixing it

- Adding the RangeTable for 8-bit value check
- add check_commitment
- and add testings

### DIFF
--- a/src/encoding/lookup.py
+++ b/src/encoding/lookup.py
@@ -6,7 +6,7 @@ class LookupTable:
     rows: Set[int]
     random: int
 
-    def __init__(self, columns: Sequence[str], random: int = 123) -> None:
+    def __init__(self, columns: Sequence[str], random: int = 5566) -> None:
         self.columns = set(columns)
         self.rows = set()
         self.random = random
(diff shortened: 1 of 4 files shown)


src/encoding/lookup.py
```diff
@@ -6,7 +6,7 @@ class LookupTable:
     rows: Set[int]
     random: int
 
-    def __init__(self, columns: Sequence[str], random: int = 123) -> None:
+    def __init__(self, columns: Sequence[str], random: int = 5566) -> None:
```
ChihChengLiang, line 9: The small value causes a collision in the compression.

For example, these two rows have the same commitment.

- a = 123, b = 0, compression = 123 * 1 + 0 * 123**1
- a = 0, b = 1, compression = 0 * 1 + 1 * 123**1

barryWhiteHat: I think if i is derieved from x8s then it should be impossible to have a collision using the values in the table.

PR #18, opened 2021-08-09
add barry's edit

ChihChengLiang:
Elaborate on why we can't do 256 bits directly


ChihChengLiang:
Post merging review welcomed

PR #31, opened 2021-08-11
[WIP] State and EVM proof

han0110:
The WIP PR aims to spec out the state proof and evm proof.

### DIFF
--- a/specs/state-proof.md
+++ b/specs/state-proof.md
@@ -36,22 +36,23 @@ class AccountNonceRecord:
 
 def account_noncee_constraint(prev: AccountNonceRecord, cur: AccountNonceRecord):
     # address should be valid
-    assert(range_lookup(cur.address, 160))
+    assert range_lookup(cur.address, 160)
     # rw should be a Write
-    assert(cur.rw == RW.Write)
+    assert cur.rw == RW.Write
 
+    # grouping
     if prev is not None:
         # address should increase (non-strcit)
-        assert(cur.address >= prev.address)
+        assert cur.address >= prev.address
 
     if prev is None or cur.address != prev.address:
         # TODO: verify the nonce exist in previous state trie root
         pass
     elif cur.address == prev.address:
         # global counter should increase
-        assert(cur.global_counter > prev.global_counter)
+        assert cur.global_counter > prev.global_counter
         # nonce can only increase by 1
-        assert(cur.nonce == prev.nonce + 1)
+        assert cur.nonce == prev.nonce + 1
 ```
 
 ### `AccountBalance`
@@ -66,24 +67,25 @@ class AccountBalanceRecord:
 
 def account_balance_constraint(prev: AccountBalanceRecord, cur: AccountBalanceRecord):
     # address should be valid
-    assert(range_lookup(cur.address, 160))
+    assert range_lookup(cur.address, 160)
     # rw should be a Read or Write
-    assert(cur.rw == RW.Read or cur.rw == RW.Write)
+    assert cur.rw == RW.Read or cur.rw == RW.Write
 
+    # grouping
     if prev is not None:
         # address should increase (non-strcit)
-        assert(cur.address >= prev.address)
+        assert cur.address >= prev.address
 
     if prev is None or cur.address != prev.address:
         # TODO: verify the nonce exist in previous state trie root
         pass
     elif cur.address == prev.address:
         # global counter should increase
-        assert(cur.global_counter > prev.global_counter)
+        assert cur.global_counter > prev.global_counter
 
         if cur.rw == RW.Read:
             # balance should be consistent to previous one
-            assert(cur.balance == prev.balance)
+            assert cur.balance == prev.balance
 ```
 
 ### `AccountCodeHash`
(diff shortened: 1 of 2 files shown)


specs/state-proof.md
```diff
@@ -0,0 +1,351 @@
+# State Proof
+
+In EVM the interpreter has ability to do any random read-write access to data like account balance, account storage, or stack and memory in current scope, but it's hard for a circuit to keep tracking these data to ensure their consistency from time to time. So the state proof helps EVM proof to check all the random read-write access records are valid, throgh grouping them by their unique index first, and then sorting them by accessing timestamp. We call the accessing timestamp **global counter**, which counts the number of access records and also serves as an unique identifier for a record.
```
ChihChengLiang, line 3: The timestamp sounds like date-time related. What do you think we rephrase "accessing timestamp" to "order of access"?

ChihChengLiang: Should we also define `global counter` as a variable somewhere and give it an `int` type?

specs/state-proof.md
```diff
@@ -0,0 +1,351 @@
...
+| [`AccountBalance`](#AccountBalance)   | `{address}`       | Account's balance                        |
+| [`AccountCodeHash`](#AccountCodeHash) | `{address}`       | Account's code hash                      |
+| [`AccountStorage`](#AccountStorage)   | `{address}.{key}` | Account's storage as a key-value mapping |
+| [`CallState`](#CallState)             | `{id}.{enum}`     | Call's internal state                    |
+| [`CallStateStack`](#CallStateStack)   | `{id}.{index}`    | Call's stack as a encoded word array     |
+| [`CallStateMemory`](#CallStateMemory) | `{id}.{index}`    | Call's memory as a byte array            |
+
+The concatenation of **Target** and **Index** becomes the unique index for data. Each record will be attached with a global counter as their accessing timestamp, and the records are constraint to be in group by their unique index first and to be sorted by their global counter increasingly. Given the access to previous record, each target has their custom constraints, for example, `AccountNonce` always increase by 1, and values in `CallStateMemory` should fit in 8-bit.
```
ChihChengLiang, line 21: > each target has their custom constraints

Is the "custom constraints" here refers to the one we use in the circuit? Or do we want to say each target has its specific format and rules to update?

ChihChengLiang: > The concatenation of **Target** and **Index** becomes the unique index for data

We might want to define the concrete concatenation inputs and outputs later

specs/state-proof.md
```diff
@@ -0,0 +1,351 @@
...
+## Circuit Constraints
+
+The following part describes the custom constraints for each target by a python script. There are some common helper class and function:
+
+```python
+def range_lookup(value: int, range: str):
+    result = re.search('^([(\[])(.+),\s+(.+)([)\]]$)', range)
+    min = eval(result.group(2)) + (1 if result.group(1) == '(' else 0)
+    max = eval(result.group(3)) - (1 if result.group(4) == ')' else 0)
+    return min <= value and value <= max
```
ChihChengLiang, line 32: Are we defining the `range_lookup` function instead of using something like `assert min < value <= max` because we want to emphasize that we have to perform table lookup for this check?
An alternative is explicitly saying all `assert min < value <= max` check requires a table lookup then use that style everywhere.

ChihChengLiang (commented):
First round of review

specs/state-proof.md
han0110: "Order of Access" sounds great!

Do you think we should have a file for all terminology or we separate the terminology in the file who defines them?

specs/state-proof.md
```diff
@@ -0,0 +1,351 @@
...
+```python
+class AccountBalance(Record):
+    fields = ['address', 'balance']
+
+
+def account_balance_constraint(prev: AccountBalance, cur: AccountBalance):
+    diff = cur - prev
+
+    # rw should be a Read or Write
+    assert cur.rw == RW.Read or cur.rw == RW.Write
```
ChihChengLiang, line 116: nitpick
```suggestion
    assert cur.rw in (RW.Read, RW.Write)
```

specs/state-proof.md
han0110: > each target has their custom constraints

We want to say each target has its specific format and rules to update. Maybe we should say:

> each target has their own format and rules to update.

specs/state-proof.md
ChihChengLiang: the latter sounds better.

specs/state-proof.md
han0110: > We might want to define the concrete concatenation inputs and outputs later

For sure, I left it ambiguous because it could be encoded with tricky optimization. Let me add a reminder here.

specs/state-proof.md
han0110: I was trying use the latter, but when value is allowed to be `0`, `min` becomes `-1`, which seems not intuitive. Maybe we could have convention `assert min <= value < max`, which seems to be more intuitive just like the range in most language.

specs/state-proof.md
```diff
@@ -0,0 +1,351 @@
...
+        # address should increase (non-strcit)
+        # TODO: check this by 160-bit range lookup or other more efficient method
+        assert diff.address >= 0
+
+    if prev is None or diff.address != 0:
+        # TODO: verify the nonce exist in previous state trie root or initialized to 0
+        pass
+    elif diff.address == 0:
+        # global counter should increase
+        assert range_lookup(diff.global_counter, '(0, 2**28)')
```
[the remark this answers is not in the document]
han0110: Yeah, that's the maximum degree of the polynomial if we are using BN254. Let me do a table for these magic values.

ChihChengLiang (approved):
Looks good to me as the first draft. We could add future PRs to complete the draft.

han0110:
Merge first draft for further work in parallel

PR #39, opened 2021-09-24
Add POP opcode spec

scroll-dev:


specs/stack_ops.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Constraints
+   1. opId = OpcodeId(0x50)
+   2. state transition:  
+      gc + 1
+      stack_pointer + 1
+      pc + 1
+      gas + 2
+
```
ChihChengLiang, line 15: Do we need to add the read-write to the bus mapping here?

DreamWuGit: we don't need  any bus mapping lookups as no operand required for Pop operation, as well the poped value will not fetch and discard directly in this "opPop" logic.

ChihChengLiang: Yeah, that makes sense. Chatted with @han0110, as he describe it that the pop is analogously resizing the array to `len - 1` and leave the poped value there. The poped value can no longer be accessed in evm because it is out of the stack now.

ChihChengLiang (approved):
LGTM, thank you for the PR!

PR #43, opened 2021-09-29
Refactor the python package and separate encoding from opcode

scroll-dev:
Try to do some refactor first so that the later opcode spec implementation could be more clear.

Open to any suggestion. @ChihChengLiang


gaswhat:
Close this PR and will recreate a new one.

PR #46, opened 2021-09-30
Add BYTE opcode spec

Brechtpd:


src/zkevm_specs/opcode/byte.py
```diff
@@ -0,0 +1,33 @@
...
+    assert len(v8s) == len(i8s) == len(r8s) == 32
+
+    # Any index value >= 256 always returns all zeros
+    msb_sum_zero = sum(i8s[1:]) == 0
+
+    # Byte 0:
+    r = r8s[0]
+    for i in range(32):
+        selected = i8s[0] == i
+        r = r - (v8s[i] * msb_sum_zero * selected)
```
ChihChengLiang, line 20: Why we are having `msb_sum_zero` here? Is this a trick to avoid creating another polynomial to check `msb_sum_zero`?

ChihChengLiang (commented):
The first round of review, added some questions

specs/opcode/26BYTE.md
```diff
@@ -0,0 +1,27 @@
...
+    - pc + 1
+    - gas + 3
+3. lookups: 3 busmapping lookups
+    - `index` is at the top of the stack
+    - `value` is at the new top of the stack
+    - `result` is at the new top of the stack
+
+## Exceptions
+
+1. stack underflow: the stack is empty or only contains a single value
```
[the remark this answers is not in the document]
Brechtpd: Oh interesting, was not aware it was checked like this. I just assumed that if you pop 2 values of the stack and 1 back on there can never be an overflow, but seems like that doesn't matter.

src/zkevm_specs/opcode/byte.py
Brechtpd: This is done here because only the LSB value of the index is used to check if the byte should be copied. So the LSB value of the index may be [0..32] which would mean we have to copy the byte value, but of course if any of the other index byte values are non-zero the index is actually a much bigger value so we should never copy any of the bytes. So `msb_sum_zero` just checks if any of those byte values are zero or non-zero once, and the we use it here as an extra condition for the byte copy.

specs/opcode/26BYTE.md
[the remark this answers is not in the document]
ChihChengLiang: Oh sorry, I think I misunderstood the Geth code, so the overflow check seems unnecessary.

src/zkevm_specs/opcode/byte.py
ChihChengLiang: I see in your circuit PR https://github.com/appliedzkp/zkevm-circuits/pull/106 you apply IsZero gadget to `msb_sum_zero` to make it 0 or 1.
But you need an extra column for IsZero gadget of msb_sum_zero`.

What do you think we do the following? Does it achieve the same result?

```python
# 1 poly (no new column for `msb_sum_zero`)
assert sum(i8s[1:]) == 0
...

    r = r - (v8s[i] * selected)  # (no `msb_sum_zero` here)
```

ChihChengLiang: Ah sorry, I see the problem with my proposed solution.
So in EVM, the non-zero MSB in `index` is still valid and will not raise any exception. It only causes the copied `result` to be zero.
We should add this note to the spec.

specs/opcode/26BYTE.md
```diff
@@ -0,0 +1,27 @@
+# BYTE op code
+
+## Procedure
+
+The `BYTE` opcode retrieves a single byte from a value. This is done by popping two values, first the `index` and then the `value`, from the stack. The byte at position `index`, with the position starting at `0` at the MSB in `value`, will be copied to the LSB of `result`. `result` is then pushed on the stack.
+
```
ChihChengLiang, line 6:
```suggestion
If the MSB in `index` is non-zero, the `result` will be 0.
```

ChihChengLiang (approved):
LGTM. Suggest adding the case of MSB of `index` to the spec.

src/zkevm_specs/opcode/byte.py
```diff
@@ -0,0 +1,33 @@
...
+    assert(r == 0)
```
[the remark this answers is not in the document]
Brechtpd: `r` was initialized to `r8s[0]`, and the selected byte was subtracted from this value, so the check was just that result == selected_byte (or 0 when no byte was selected).
I actually wrote the check in a more clear way like `assert(r == r8s[0])`  in the circuits but did not think to do it here as well, so changed it to that here now as well.

miha-stopar (approved):

PR #47, opened 2021-10-09
Bitwise: xor or and

Lawliet-Chan:


specs/opcode/18XOR.md
```diff
@@ -0,0 +1,50 @@
+# XOR op code
+
+## Procedure  
+
+A stack initalize empty with stack pointer to 1024, pop operation can only happen when stack is not empty, and it will increase by 1 of stack pointer.
```
miha-stopar, line 5: Same comments as above.

specs/opcode/17OR.md
```diff
@@ -0,0 +1,49 @@
+# OR op code
+
+## Procedure  
+
+A stack initalize empty with stack pointer to 1024, pop operation can only happen when stack is not empty, and it will increase by 1 of stack pointer.  
+
+The or-ed value will be dropped directly without no any more checking & utilizing.
```
miha-stopar, line 7: Same comments as above for AND.

specs/opcode/16AND.md
```diff
@@ -0,0 +1,50 @@
...
+    - gas + 3
+
+
+## Exceptions
+
+1. stack underflow: when stack is empty
+2. gas out: remaining gas is not enough   
+
+```python
+class Stack():
```
miha-stopar, line 25: Is this code needed here? Also, I am not sure about pop and push implementation? There is a code for Stack in 50POP.md which has the pop and push implementation as it's in Geth. 

BTW @ChihChengLiang : should we move code for Stack in 50POP.md into src folder?

ChihChengLiang: Yes, I think that's a good idea.

PR #52, opened 2021-10-14
Entrypoint for EVM circuit and specification for `BEGIN_TX`

### LINKED ISSUE #29: EVM errors
The way our zkevm handles errors is important to clarity. Different opcodes deal with that in different ways and have to be aware of different error cases. For example add opcode needs to be aware of two possible underflows on the stack and OOG after the operation has been completed. Defining this and having a consistent way to throw errors seems useful. 

https://github.com/appliedzkp/zkevm-circuits/pull/26 already tries to address these issues. But its still important to define them clearly 

https://hackmd.io/T6rSea43RTOhXre5sYY37A is a list of such error cases we need to be aware of.

han0110:
This PR aims to:

- Provide a entrypoint `verify_steps` to better describe EVM circuit.
- Naively setup all used lookup tables' interface (except `keccak256_table`).
- Naively enumerate all possible execution result. (resolves #29)
- Spec `BEGIN_TX` to verify the beginning of a transaction.
- Implement `ADD` and `PUSH` using newly introduced `Instruction` as sample code.
- Take #66 into consideration in `bytecode_table` for further implementation.

Some changes have been made including:

- Bus mapping is split into multiple tables corresponding their provider. In the end we can still merge them into same lookup table with some selector, but it depends on the dimension and desired capacity.
- `global_counter` is renamed to `rw_counter` cause only writable data are time-aware.
- Error cases is separated from success cases to avoid redundant implementation and test. See [`ExecutionResult`](https://github.com/appliedzkp/zkevm-specs/pull/52/files#diff-f467252436c0df869ff222eba7983f188b21dcbafc989820fbd5a994846eccdcR4) for all possible cases.
- Introduce `Instruciton` to simulate both verifier (constrain some relation) and prover (calculate some non-qudratic relation).


src/zkevm_specs/evm/step.py
```diff
@@ -0,0 +1,432 @@
...
+                self.call.opcode_source,
+                self.call.program_counter + offset,
+            ])
+
+    def stack_pop_lookup(self) -> int:
+        value = self.r_lookup(RWTableTag.Stack, [
+            self.core.call_id,
+            self.call.stack_pointer + self.stack_pointer_diff
+        ])[0]
+        self.stack_pointer_diff += 1
```
DreamWuGit, line 308: do we need to move "self.stack_pointer_diff += 1" to the beginning  of current method ?

han0110: When pushing data to stack, we move (decrease) the `stack_pointer` first, then write the data. So when popping the value, we should do it reversely, read the data first, then move (increase) the `stack_pointer`, right?

DreamWuGit: yeah, agree!

src/zkevm_specs/evm/execution_result/call.py
```diff
@@ -0,0 +1,156 @@
...
+from .execution_result import ExecutionResult
+
+
+def call(curr: Step, next: Step, r: int, opcode: Opcode):
+    # Verify opcode
+    assert opcode == Opcode.CALL
+
+    # Verify depth
+    depth = curr.call_context_lookup(CallContextTag.Depth)
+    curr.fixed_lookup(FixedTableTag.Range1024, [depth])
```
DreamWuGit, line 15: I think valid depth maximum can be  1024 according to https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L172  , the look up range should be Range1025 , pls. check it !Thx

han0110: Thanks for catching this! Depth starts from `1` and would always be propagated by caller, so let's check it's in range by `curr.fixed_lookup(FixedTableTag.Range1024, [depth - 1])`.

src/zkevm_specs/evm/table.py
```diff
@@ -0,0 +1,199 @@
...
+        [(FixedTableTag.StackOverflow, opcode, stack_pointer, 0) for (opcode, stack_pointer) in stack_overflow_pairs()] +
+        [(FixedTableTag.OOGConstant, opcode, gas, 0) for (opcode, gas) in oog_constant_pairs()]
+    )
+    tx_table: Set[Tuple[
+        int,  # tx_id
+        int,  # tag
+        int,  # index (or 0)
+        int,  # value
+    ]]
+    bytecode_table: Set[Tuple[
```
DreamWuGit, line 163: what do you think of add "is_code" to bytecode_table ,which indicate whether it is true code or not ? seems common requirement  for three scenarios regarding byte code table.

han0110: Here is my consideration:
If we don't want to explicitly gather (copy) opcode into a single table from different sources, for opcode fetching, we might have 3 sources in different situation:
- When contract interaction, we lookup `bytecode_table` to read bytecode
- When contract creation in root call, we lookup `tx_table` to read transaction's calldata
- When contract creation in internal call, we lookup `rw_table` to read caller's memory

With an annotation `is_code` will be useful for handling opcode `JUMP*`, however, the source could other than `bytecode_table`, so only annotating it in `bytecode_table` seems not a general solution, we also need to annotate such information in `tx_table` and `rw_table`, and the latter one is more tricky since the memory would not always be layouted adjacently.

So currently I prefer to just look back 32 opcodes when handling `JUMP*`, what do you think?

src/zkevm_specs/evm/execution_result/call.py
```diff
@@ -0,0 +1,156 @@
...
+            curr.bytes_range_lookup(callee_gas_left - gas, 8)
+            callee_gas_left = gas
+    else:
+        assert is_capped
+
+    next_gas_left = curr.call.gas_left - base_gas_cost - callee_gas_left
+
+    # TODO: Handle precompile
+    if is_empty_code_hash:
+        assert result == 1
```
DreamWuGit, line 84: what does the result mean , call method result ?

src/zkevm_specs/evm/execution_result/call.py
```diff
@@ -0,0 +1,156 @@
...
+    is_account_empty = is_zero_nonce and is_zero_balance and is_empty_code_hash
+    base_gas_cost = 100 \
+        + is_cold_access * 2500 \
+        + is_account_empty * 25000 \
+        + has_value * 9000 \
+        + memory_gas_cost
+
+    gas_available = curr.call.gas_left - base_gas_cost
+    one_64th_available_gas = le_to_int(curr.allocate_byte(8))
+    curr.fixed_lookup(FixedTableTag.Range64, [gas_available - 64 * one_64th_available_gas])
```
DreamWuGit, line 66: need to assert the lookup ?

tests/evm/test_begin_tx.py
```diff
@@ -0,0 +1,113 @@
...
+    main, ExecutionResult, CoreState, CallState, Step,
+    Tables, TxTableTag, CallContextTag, RWTableTag,
+)
+from zkevm_specs.util import fp_inv, linear_combine, keccak256, EMPTY_CODE_HASH
+
+
+# TODO: Parametrize r, tables
+def test_begin_tx():
+    r = 1
+    bytecode = bytes.fromhex('00')
```
DreamWuGit, line 11: I understand begin_tx is a flag we set in circuit side as a placeholder, there is no explicitly care in evm , if so the call state and others  i.e. allocations . should be empty at this step , right ?

src/zkevm_specs/evm/execution_result/call.py
han0110: It's the value pushed to caller's stack after the call to tell if the call succeeds, see [here](https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L665-L670).

src/zkevm_specs/evm/execution_result/call.py
han0110: The assertion is already done inside the function.

tests/evm/test_begin_tx.py
han0110: Yeah, the `call_state` will not be used and no opcode will be fetched. However, we still need allocation to verify a bunch of stuff, like verifying if it could buy intrinsic gas, transfer ether, callee has code or not, etc...

src/zkevm_specs/evm/table.py
DreamWuGit: I think looking back 32 opcodes seems has problem,  the back 32th byte can be code or data as well, also need to check it.   thus looking back deeper requires,   do you think it is problem ?

src/zkevm_specs/evm/execution_result/call.py
```diff
@@ -0,0 +1,152 @@
...
+        if result and not is_persistent:
+            assert rw_counter_end_of_reversion == callee_rw_counter_end_of_reversion
+            assert callee_state_write_counter == \
+                curr.call.state_write_counter + curr.state_write_counter_diff
+
+        curr.assert_step_transition(
+            next,
+            rw_counter_diff=curr.rw_counter_diff,
+            execution_result_not=ExecutionResult.BEGIN_TX,
+            call_id=curr.core.rw_counter,
```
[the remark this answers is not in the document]
icemelon: Got it. That makes sense.

src/zkevm_specs/evm/table.py
han0110: But `PUSH*` is up to `PUSH32`, why do we need to look back more than 32 opcodes?

DreamWuGit: let's say see example:  push32  00 01 02 03 ... 1f
translate to byte code sequence is: 
7f  00 01. 02...  1f
the '7f' can also be data segment as it is maybe in previous push* section recursively.   right ?

han0110: Ohh, you are right. Never thought about such false alarm case... Maybe we need to annotate `is_code` for them. Let me think about this and convert it into an issue for further discussion.

Could we add `is_code` in `bytecode_table` in further PR for more discussion?

DreamWuGit: sure !

src/zkevm_specs/evm/execution_result/call.py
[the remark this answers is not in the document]
icemelon: Is it possible that there is no rw in stack or memory in a call?

tests/evm/test_begin_tx.py
icemelon: Do we need to assign an opcode id to BEGIN_TX?

src/zkevm_specs/evm/execution_result/call.py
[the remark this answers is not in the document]
han0110: For an internal call, in the end it needs to read out caller's context and restore it to next step, so there must be some rw for it.

tests/evm/test_begin_tx.py
han0110: The idea here is try to define our own finite state machine to emulate EVM, because the opcode in EVM cost very differently between each other, and the error cases make even simple opcode complicated. So I try to enumerate all kind of possible execution result of a step, and repeat such step to emulate EVM.

Then `BEGIN_TX` is just one kind of execution result which doesn't consume opcode, there might be some pre-compiles do so.

src/zkevm_specs/evm/execution_result/call.py
```diff
@@ -0,0 +1,156 @@
...
+    # Verify depth
+    depth = curr.call_context_lookup(CallContextTag.Depth)
+    curr.fixed_lookup(FixedTableTag.Range1024, [depth - 1])
+
+    # Gas needs full decompression due to EIP 150
+    bytes_gas = curr.decompress(curr.stack_pop_lookup(), 32, r)
+    bytes_callee_address = curr.decompress(curr.stack_pop_lookup(), 32, r)
+    bytes_value = curr.decompress(curr.stack_pop_lookup(), 32, r)
+    bytes_cd_offset = curr.decompress(curr.stack_pop_lookup(), 32, r)
+    bytes_cd_length = curr.decompress(curr.stack_pop_lookup(), 5, r)
```
icemelon, line 22: Why limit the length to 5 bytes? I see in the Geth, the length uses Uint64 type. So it can be up to 8 bytes. Though it's impossible to have such large data, I feel it's better to be consistent?

han0110: In EVM the `memory_address` could be up to <code>2<sup>37</sup>-32</code> in successful cases<sup>[1]</sup>. I think decompressing every `memory_address` into 5 bytes is reasonable, although it then could be up to <code>2<sup>40</sup>-1</code> which leads to inequivalent to current EVM when gas limit is raised to very large value <code>≈2<sup>55</sup></code>, but I don't think it's possible.

The reason `cd_offset` is decompressed into 32 bytes is because the behavior of EVM, which allows `offset` to be any value if the `length` is `0`.

[1]: https://github.com/ethereum/go-ethereum/blob/master/core/vm/gas_table.go#L33-L40

src/zkevm_specs/evm/step.py
```diff
@@ -0,0 +1,432 @@
...
+
+        # Verify next_memory_size_rd is correct
+        if has_rd_length:
+            assert sum(bytes_rd_offset[5:]) == 0
+        self.fixed_lookup(FixedTableTag.Range32, [32 * next_memory_size_rd - has_rd_length * (rd_offset + rd_length)])
+
+        # Verify next_memory_size == \
+        #   max(self.call.memory_size, next_memory_size_cd, next_memory_size_rd)
+        assert next_memory_size in [self.call.memory_size, next_memory_size_cd, next_memory_size_rd]
+        self.bytes_range_lookup(next_memory_size - self.call.memory_size, 4)
```
icemelon, line 385: Why do we use 4 bytes here while cap the length at 5 bytes? We should also use variables to represent these constants to make them more meaningful and consistent.

han0110: Agree, I should mention why we store memory size in word and why it should fit in 4 bytes.

The reason to store memory size in word is because memory expansion always expands to multiple of word, and it makes the quadratic gas cost easier to calculate (we don't need to divide it by 32 everytime).

The reason to use 4 bytes is still due to the quadratic gas cost, the memory size in word could be up to <code>2<sup>32</sup>-1</code>, which doesn't overflow `u64` when doing square (mimicking the behavior of current EVM).

src/zkevm_specs/evm/step.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def stack_push_lookup(self) -> int:
+        self.stack_pointer_diff -= 1
+        return self.w_lookup(RWTableTag.Stack, [
+            self.core.call_id,
+            self.call.stack_pointer + self.stack_pointer_diff
+        ])[0]
+
+    def assert_sufficient_constant_gas(self, opcode: Opcode) -> int:
+        next_gas_left = self.call.gas_left - OPCODE_INFO_MAP[opcode].constant_gas
```
icemelon, line 319: The `constant_gas` is a negative number. So here should use `+`?

src/zkevm_specs/evm/step.py
icemelon: Got it.

ChihChengLiang (approved):
I did not give a thorough look, just reading through the previous review discussions. If there's nothing more to add to this PR, we can merge and unblock the future PR.

DreamWuGit (approved):
LGTM!

PR #53, opened 2021-10-19
LT & GT opcode spec

scroll-dev:
Add the python code and doc for LT and GT opcode. Also refactor the test code.


specs/opcode/10LT_11GT.md
```diff
@@ -0,0 +1,28 @@
+# LT & GT opcode
+## Procedure
+
+The `LT` and `GT` opcode compares the top two values on the stack, and push the
+result (0 or 1) back to the stack.
+
```
ChihChengLiang, line 6: I'd like to see a short description on

1. How do we use `c8s === b8s - a8s` to check `b8s > a8s`
2. How do we use `swap` to reuse the same gadget for GT and LT.

So that the reviewer or the auditor can have a rough idea before diving into the Python code.

src/zkevm_specs/opcode/lt_gt.py
```diff
@@ -0,0 +1,76 @@
...
+@is_circuit_code
+def check_lt(
+    a8s: Sequence[U8],
+    b8s: Sequence[U8],
+    result8s: Sequence[U8],
+    c8s: Sequence[U8],
+    swap: bool,
+    carry: bool,
+):
+    assert not swap
```
ChihChengLiang, line 62: I would suggest renaming `swap` to something like `is_gt` so that it doesn't collide with the `SWAPX` opcode

src/zkevm_specs/opcode/lt_gt.py
```diff
@@ -0,0 +1,76 @@
...
+def lt_gt_common(
+    a8s: Sequence[U8],
+    b8s: Sequence[U8],
+    result8s: Sequence[U8],
+    c8s: Sequence[U8],
+    carry: bool,
+):
+    """
+    Check if c8s equals to b8s - a8s.
+    When result is 1, the last carry of the addition of a8s and c8s should be 0,
```
miha-stopar, line 14: Is this properly formulated? Don't we have only one carry?

miha-stopar (commented):
LGTM! However, would it make sense to include the EQ opcode as well if https://github.com/appliedzkp/zkevm-circuits/pull/120 is merged?

PR #54, opened 2021-10-21
Fix typo and syntax

NoCtrlZ:
Hi there.
I found some typo so I corrected.
I would be grateful if you could confirm.
Thank you!


ChihChengLiang (approved):
Looks good! Nice catch!

PR #72, opened 2021-11-29
Add formating and linting

ChihChengLiang:
### Problem

We are seeing review feedbacks on formating more. We can add automation to this.

### Solution

- Add formatting tools and checks
- Add Black to format Python
- Add mdformat to format markdown. Note: I like prettier more since it formats markdown and other configs. It also has more editor supports. Unfortunately, it's a nodejs package, which makes the project setup more complicated. So we add a Makefile fmt command to mitigate the inconvenience in editor formatting.
- Add Makefile to create a similar devex in zkevm-circuit


miha-stopar (approved):
Nice, really helpful!

PR #74, opened 2021-12-04
Bytecode circuit spec

Brechtpd:
Implementation: https://github.com/appliedzkp/zkevm-circuits/pull/170

The keccak lookup table format is not final yet so may need additional changes for this as well when that happens (currently uses the WIP format that is the most convenient here).

(I'm not sure if the byte needs to be constrained in the circuit because it should already be indirectly be range limited by the keccak circuit, but kept it for now. If not necessary the circuit could be simplified with the removal of the `byte_push_size` column).


han0110 (approved):
LGTM! Spec is really clear, nice work!

Though I was thinking the `UnrolledBytecode` would implement the `table_assignments` instead of using that of `Bytecode`, I actually built it for temporary usage just like the `rw_table` and `tx_table`, they should all be replaced with actual circuit.

If so, I think column `is_code` could be synthetic expression like `(not row.q_first and not prev_row.is_final) * is_zero(prev_row.push_data_left) + (row.q_first or prev_row.is_final))`. But it also seems to add complexity to save nothing.

Let's discuss more and then integrate them in the future PRs.

Brechtpd:
> Though I was thinking the `UnrolledBytecode` would implement the `table_assignments` instead of using that of `Bytecode`, I actually built it for temporary usage just like the `rw_table` and `tx_table`, they should all be replaced with actual circuit.

Ah I see. So are all columns are always accessible when used in other circuits? I don't know how the table data is actually shared between the circuits in practice.

han0110:
> So are all columns always accessible when used in other circuits

Yes, that's what I was thinking to have a proof of concept, but maybe not directly accessible, instead the dependency circuit to output synthetic table expressions as lookup table for dependent circuit.

I am also not sure how we end up integrate different circuits together, but this is the most straightforward approach I think, and that's why I created the [`trait LookupTable`](https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/table.rs#L8-L10) for EVM circuit to easily switch between mock tables or real tables that have been verified in another layout.

Brechtpd:
Any resources on what those "synthetic table expressions" would be, I have no idea what to think of them. If there aren't any that's fine, I'll just wait on your implementation and see. :)

The only method I know of to verifiably share data between circuits is by using public inputs, so I thought something like the shared data would be put into both circuits and that data would be accumulated somehow (e.g. similarly to how the hash data is accumulated) and verified against the same public input value in both circuits. But it looks like there's something smarter possible.

han0110:
> Any resources on what those "synthetic table expressions" would be

Sorry I should not create the weird terms, I just want to emphasize that, usually we use lookup argument to some fixed columns as public lookup table, but it doesn't restrict us to use lookup argument to advice columns as private lookup table or even the linear combination of advice columns with fixed columns (selectors), as long as they are already verified to have some relation (e.g. the hash of unrolled bytes are correct).

For example, the `bytecode_table` EVM circuit expects would look like: `[hash, index, byte, is_code]`, and all of them are already advice columns in Bytecode circuit, it should be able to lookup to such advice columns directly (we might also need to multiply 4 of them with `not padding`).

But actually the `is_code` is just the `is_zero` expression generated by `push_data_left` and `push_data_left_inv`, it doesn't actually need to be there as a column to be lookup-ed by EVM circuit, Bytecode circuit only needs to generate such expression.

> The only method I know of to verifiably share data between circuits is by using public inputs, so I thought something like the shared data would be put into both circuits and that data would be accumulated somehow (e.g. similarly to how the hash data is accumulated) 

I think the main reason to have `hash_rlc` is to reduce the complexity to use Keccak circuit, originally I was thinking we need to lookup u64 to u64 on every rouund and make sure all bytes are absorbed. But with "random linear combination magic", we can make sure the bytes between the 2 tables are synchronised, then we don't need to worry about composition from bytes to u64 in Bytecode circuit side.

So for me it's more like a serialization of data transmission between circuit instead of make them verifiable, lookup would be the one ensure they are verifiable. As for other cases we try to share data between circuits, most of them are place well in a row, so we don't need such special serialization.

specs/bytecode-proof.md
```diff
@@ -0,0 +1,57 @@
...
+| Column                | Description                                                        |
+| --------------------- | ------------------------------------------------------------------ |
+| `q_first`             | `1` on the first row, else `0`                                     |
+| `q_last`              | `1` on the last row, else `0`                                      |
+| `hash`                | The keccak hash of the bytecode                                    |
+| `index`               | The position of the byte in the bytecode                           |
+| `byte`                | The byte data for the current position                             |
+| `is_code`             | `1` if the byte is code, `0` if the byte is PUSH data              |
+| `push_data_left`      | The number of bytes that still need to follow for PUSH data        |
+| `hash_rlc`            | The accumulator containg the current and previous bytes            |
```
miha-stopar, line 16: typo: containg

specs/bytecode-proof.md
```diff
@@ -0,0 +1,57 @@
...
+
+| Column                | Description                                                        |
+| --------------------- | ------------------------------------------------------------------ |
+| `q_first`             | `1` on the first row, else `0`                                     |
+| `q_last`              | `1` on the last row, else `0`                                      |
+| `hash`                | The keccak hash of the bytecode                                    |
+| `index`               | The position of the byte in the bytecode                           |
+| `byte`                | The byte data for the current position                             |
+| `is_code`             | `1` if the byte is code, `0` if the byte is PUSH data              |
+| `push_data_left`      | The number of bytes that still need to follow for PUSH data        |
```
miha-stopar, line 15: I was a bit confused at first about `push_data_left` naming, I thought it's some info about how much push data are to the left (direction), it's totally clear though from the description, not sure why I though this :). However, perhaps you can add a couple of words like: how many bytes of data are still left behind this position (might be even named push_data_behind, but it's really not important, you can ignore me :)).

specs/bytecode-proof.md
```diff
@@ -0,0 +1,57 @@
...
+| `hash_rlc`            | The accumulator containg the current and previous bytes            |
+| `hash_length`         | The bytecode length                                                |
+| `byte_push_size`      | The number of bytes pushed for the current byte                    |
+| `is_final`            | `1` if the current byte is the last byte of the bytecode, else `0` |
+| `padding`             | `1` if the current row is padding, else `0`                        |
```
miha-stopar, line 20: Perhaps a couple of words why padding is needed, it's because lookup into keccak_table fails for zero rows, right?

miha-stopar (approved):
Very nice! Just some minor comments, you can ignore them :).

specs/bytecode-proof.md
Brechtpd: Interesting way to read `left` here, did not even think about that. :) Changed the description a bit more to what you described.

specs/bytecode-proof.md
Brechtpd: Yeah didn't want to make assumptions, added this: 
> Explicit padding is added to this circuit to be able to fully fill the circuit with valid data without depending on the keccak circuit to either hash this padding data or support looking up e.g. all zero data.

miha-stopar: Cool, thanks!

Issue #75, opened 2021-12-06
handle block related op codes 


DreamWuGit:
there is kind of op codes referring to block properties , i.e. Blockhash, Coinbase,  gas limit, Difficulty, Timestamp etc.   except `Blockhash`  requires Keccak256 hash operation, others behave similarly:
 getting target property from block context and push to stack , also with some type conversion  sometimes. 

for `Blockhash`  need separated handling regarding to Keccak256 circuit.
for others ,  my intuitive approach is constructing block context (or called properties )table,  which contains such above properties, and carefully deal with each data type in circuit , by looking up this table to constraint them.   additional to add these properties into tests'  block section for witness  feeding.

appreciate to hear more ideas or comment here :)

barryWhiteHat:
This makes sense. I think that we should start by building a table (block_constants_table) of these constants in the verifier contract similar to [here](https://github.com/appliedzkp/zkevm-circuits/issues/122)

From there we have two choices 
1. In the evm proof get the value from that table, then add a push to the bus mapping that tells the start circuit to do that push.  
2. In the evm proof pass a generic tag ("coinbase") and in the state proof convert that to the correct address by a lookup to block_constants_table. 

@DreamWuGit what do you think about the two proposal above ? Does my explanation of the earlier part match your understanding ?

han0110:
> for Blockhash need separated handling regarding to Keccak256 circuit.

I think we would have built block hashes in layer1 contract for fast retrieval instead of calculating previous block hashes on the fly. Does it make sense to store previous 256 block hashes in layer1 contract @barryWhiteHat?

Also In my WIP [`call-and-stop`](https://github.com/han0110/zkevm-specs/tree/feature/call-and-stop) branch, I was trying to propose the similar approach, to put block context with transaction and use lookup to read the value, because they are both built from public input. I would separate it out as a independent PR to unblock further block context related work if the approach is acceptable (also after #52 is merged).

In normal case, such static data could be retrieve by copy constraint, but since there is `BLOCKHASH`, which reads `previous_256_block_hashes[stack.pop()]`, which depends on dynamic witness and needs lookup to handle.

[...]

PR #76, opened 2021-12-08
[WIP] Follow `EIP1559` to calculate correct transaction gas price

han0110:
This PR aims to add the missing block constant functionality of `Instruction` and implement `EIP1559` on `BeginTx`.

Currently the block constant can be used directly, just like using the copy constraint in halo2. But the opcode `BLOCKHASH` requires to access up to previous 256 block hashes, which would be expensive to copy 256 value at each step. See more discussion on #75.


han0110:
Close in favor of #77

PR #77, opened 2021-12-08
Add `block_lookup` in `Instruction` for block constant access

han0110:
This PR aims to add the missing block constant functionality of `Instruction`.

It adds a `block_table` to store all static block constant and up to previous 256 block hashes for block constant lookup. The table in real implementation could share the same instance columns with `tx_table` once they are properly separate by selectors.

This approach targets a straightforward implementation, in the future we could also adopt `EIP2935` mentioned in #75 to totally remove `block_table`.

### DIFF
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -221,14 +221,14 @@ class Tables:
 
     # Each row in BlockTable contains:
     # - tag
-    # - block_number_or_zero
+    # - block_number_or_zero (meaningful only for BlockHash, will be zero for other tags)
     # - value
-    block_table: Set[Array4]
+    block_table: Set[Array3]
 
     # Each row in TxTable contains:
     # - tx_id
     # - tag
-    # - index_or_zero
+    # - index_or_zero (meaningful only for CallData, will be zero for other tags)
     # - value
     tx_table: Set[Array4]
 
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -1,7 +1,7 @@
 from typing import Sequence, Set, Tuple
 from enum import IntEnum, auto
 
-from ..util import Array4, Array4, Array8
+from ..util import Array3, Array4, Array8
 from .execution_state import ExecutionState
 from .opcode import (
     invalid_opcodes,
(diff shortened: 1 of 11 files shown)


src/zkevm_specs/evm/table.py
```diff
@@ -202,6 +219,12 @@ class Tables:
         ]
     )
 
+    # Each row in BlockTable contains:
+    # - tag
+    # - block_number_or_zero
```
DreamWuGit, line 224: block number is zero seems make no sense as it begins from 1.

src/zkevm_specs/evm/typing.py
```diff
@@ -1,7 +1,53 @@
...
+        self.gas_limit = gas_limit
+        self.block_number = block_number
+        self.time = time
+        self.difficulty = difficulty
+        self.base_fee = base_fee
+        self.previous_block_hashes = previous_block_hashes
+
+    def table_assignments(self, rlc_store: RLCStore) -> Sequence[Array3]:
+        return [
+            (BlockContextFieldTag.Coinbase, 0, self.coinbase),
```
DreamWuGit, line 41: use self.block_number here ?  `(BlockContextFieldTag.Coinbase, 0, self.block_number), 
(BlockContextFieldTag.GasLimit, 0, self.block_number),` the following does the same .

src/zkevm_specs/evm/table.py
han0110: This is only for `BlockContextFieldTag.BlockHash` usage, I have added some more comments in [`c52c2f1`](https://github.com/appliedzkp/zkevm-specs/pull/77/commits/c52c2f16158aaa629906a38ff5d0448b3b115a76)

src/zkevm_specs/evm/typing.py
han0110: The `block_number_or_zero` is only meaningful for `BlockContextFieldTag.BlockHash`, and will be `0` for other tags. The reason to have `block_number_or_zero` is because prover can index one of 256 block hashes by stack value, so we build such friendly lookup table for it to use.

DreamWuGit: I guess you assume block table  only have one block, my first idea is table look like:    
| block_number |tag   | value|
| -------    |   ---- | -----  |
| 1000704| hash    | 0x29..38 |
| 1000704| coinbase  | 0x5B3..C4|. 

which can hold multiple blocks,   it would be useful for aggregate block proofs, let me know your thinking here

han0110: Yes, I assume in the same proof we are in the same block. If in the future we need to verify multiple block in a proof, I think we only need to add another `block_number` column just like what you propose.

DreamWuGit: OK, make sense to me for now .

DreamWuGit (approved):
LGTM Thanks！

icemelon (approved):
LGTM

icemelon:
Could we merge this first? We have a `COINBASE` spec pr that depends on this one.

ChihChengLiang (approved):
approve to unblock other tasks.

PR #78, opened 2021-12-08
Use legacy transaction format and defer support of dynamic gas price

han0110:
This PR aims to change the transaction format to legacy one, and defer support of dynamic gas price transaction introduced by `EIP1559` to the future work for simplicity.

It also does:

- Extend `add_word` to `add_words` to match current circuit implementation
- Add `CallDataGasCost` in `tx_table` for lookup to calculate the correct intrinsic gas cost for legacy tx

### DIFF
--- a/src/zkevm_specs/evm/instruction.py
+++ b/src/zkevm_specs/evm/instruction.py
@@ -268,7 +268,7 @@ class Instruction:
         self,
         tag: RWTableTag,
         inputs: Sequence[int],
-        is_persistent: bool = True,
+        is_persistent: bool,
     ) -> Array8:
         assert tag.write_only_persistent()
 
@@ -281,8 +281,8 @@ class Instruction:
         self,
         tag: RWTableTag,
         inputs: Sequence[int],
-        is_persistent: bool = True,
-        rw_counter_end_of_reversion: int = 0,
+        is_persistent: bool,
+        rw_counter_end_of_reversion: int,
     ) -> Array8:
         assert tag.write_with_reversion()
 
@@ -341,8 +341,8 @@ class Instruction:
         self,
         account_address: int,
         account_field_tag: AccountFieldTag,
-        is_persistent: bool = True,
-        rw_counter_end_of_reversion: int = 0,
+        is_persistent: bool,
+        rw_counter_end_of_reversion: int,
     ) -> Tuple[int, int]:
         row = self.state_write_with_reversion(
             RWTableTag.Account,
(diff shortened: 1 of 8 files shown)


src/zkevm_specs/evm/instruction.py
```diff
@@ -74,32 +74,8 @@ def constrain_equal(self, lhs: int, rhs: int):
...
-
-        result, carry = self.add_word(value_with_gas_fee, sender_balance)
-        self.constrain_equal(sender_balance_prev, result)
-        self.constrain_zero(carry)
-
-        result, carry = self.add_word(value, receiver_balance_prev)
-        self.constrain_equal(receiver_balance, result)
-        self.constrain_zero(carry)
+    def constrain_sufficient_gas_left(self, gas_left: int):
+        self.int_to_bytes(gas_left, N_BYTES_GAS)
```
ChihChengLiang, line 78: Is the intention of this function an underflow check?

han0110: Yes

ChihChengLiang: What do you think we rename `constrain_sufficient_gas_left ` to something like `constrain_not_underflow` to reflect the purpose?

han0110: Sounds good to me. Renamed in [`6b37c0b`](https://github.com/appliedzkp/zkevm-specs/pull/78/commits/6b37c0b7f5e755c13f44642b8f04f27c13483ee9)

ChihChengLiang (approved):
LTGM

src/zkevm_specs/evm/typing.py
```diff
@@ -55,8 +67,7 @@ class Transaction:
     id: int
     nonce: U64
     gas: U64
-    gas_tip_cap: U256
-    gas_fee_cap: U256
+    gas_price: U256
```
DreamWuGit, line 70: does `gas_price` really need `U256` size ?

han0110: I am not sure actually, I basically tried to follow those types used in `geth`. Some of them like `nonce` and `gas` have less bits than the one specified in yellow paper, but others match.

So I was thinking there might be some reason for `geth` to have `big.Int` for `gas_price`, so I keep it as `U256` here.

DreamWuGit: if so,  does `N_BYTES_GAS = 8` of param.py need to change accordingly ?

tests/evm/test_begin_tx.py
```diff
@@ -42,7 +46,7 @@ def test_begin_tx(tx: Transaction, result: bool):
     block = Block()
     caller_balance_prev = rlc_store.to_rlc(int(1e20), 32)
     callee_balance_prev = rlc_store.to_rlc(0, 32)
-    caller_balance = rlc_store.to_rlc(int(1e20) - (tx.value + tx.gas * tx.gas_fee_cap), 32)
+    caller_balance = rlc_store.to_rlc(int(1e20) - (tx.value + tx.gas * tx.gas_price), 32)
     callee_balance = rlc_store.to_rlc(tx.value, 32)
 
     bytecode = Bytecode("00")
```
DreamWuGit, line 52: I understand for begin_tx ,the byte code `bytecode = Bytecode("00")` here doesn't matter for any input , right ?

src/zkevm_specs/evm/typing.py
han0110: `N_BYTES_GAS` is for `gas` itself, tho it could have value up to 256 bits in yellow paper, but it's used as `uint64` in `geth`, so `N_BYTES_GAS = 8`.

While `gas_price` is the unit that maps `gas` to real `ether`, which also could have value up to 256 bits, and used as `big.Int` in `geth`, so we try not to change it in case there is some edge cases that haven't came to our minds.

tests/evm/test_begin_tx.py
han0110: Here is a decision I haven't figured it out yet, thanks for reminding me, I will convert this into a issue ASAP.

The decision is, if `callee.code_hash = keccak256('')` (e.g. `callee` is not yet a contract or precompiled), and when there is call to it, we have 2 options:

1. Only do ether transfer, and not dive into another call.
2. Do ether transfer, and prepare the call context just like we are diving into another call. The first executed opcode will always be `STOP` because `program_counter == code_length` (which is not yet implemented).

The first option is the exact behavior of `geth` ([ref](https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L219-L220)), but it introduces a branch if `callee.code == keccak256('')`, which requires some more constraints.

The second option is somehow weird, but it doesn't need extra constraint of `BeginTx` or `CALL` in the first option, by sacrificing an extra step, which I think is okay to do.

So the testing now have `Bytecode("00")`, which is actually having the exact behavior of the second option, to temporarily avoid this decision.

src/zkevm_specs/evm/typing.py
DreamWuGit: OK to me for now! Thx

tests/evm/test_begin_tx.py
DreamWuGit: yeah, we can think more later, I can approve this pr first for now.

DreamWuGit (approved):

HAOYUatHZ (approved):
LGTM as well!

PR #81, opened 2021-12-23
spec for opcode MUL

scroll-dev:
Mostly follows the spec doc in https://hackmd.io/HL0QhGUeQoSgIBt2el6fHA with some bug fixes


specs/opcode/02MUL.md
```diff
@@ -0,0 +1,96 @@
+# MUL op code
```
ed255, line 1: I'm having some difficulty reading the latex equations and tables.  It seems that there are extra double `\` everywhere, where only one is required for latex.  For example, if I copy the document into hackmd this is what I see https://hackmd.io/b9qqwnwnTpCxJlrsv1D58w?both
What tool did you use to render this markdown?

HAOYUatHZ: I agree with u, I think it's our mistake.

HAOYUatHZ: @noel2004

src/zkevm_specs/opcode/mul.py
```diff
@@ -0,0 +1,56 @@
...
+    y8s: Sequence[U8],
+    t0: U256,
+    t1: U256,
+    t2: U256,
+    t3: U256,
+    v0: Sequence[U8],
+    v1: Sequence[U8],
+):
+    assert len(a8s) == len(x8s) == len(y8s) == 32
+    assert len(v0) == len(v1) == 9
```
ed255, line 18: `specs/opcode/02MUL.md` specifies that v0 and v1 are between 0 and 2^64.  But I notice that here you assert that v0 and v1 are 9 elements of 8 bits, which means between 0 and 2^72.  Am I missing something?

noel2004: t1 is the sum of two 128bit products from limbs so they would excceed 128bit, which lead to the possibility that v0 would excceed 64bit. So does v1 and in fact it should not exceed 2^66 instead of 2^64. We can relax the constraint to make v0/v1 not exceed 2^72 or 9 bytes because everything should be ok only if v0 (or v1) * 2^128 not exceed the prime field we used (for bn254 it is about 253bit).

I should update the narration from its reference in the following patch.

src/zkevm_specs/opcode/mul.py
```diff
@@ -0,0 +1,56 @@
...
+            a_idx, x_idx = a_id, total_idx - a_id
+            if len(a64s) >= a_idx + 1:
+                tmp_a = a64s[a_idx]
+            else:
+                tmp_a = 0
+            if len(x64s) >= x_idx + 1:
+                tmp_x = x64s[x_idx]
+            else:
+                tmp_x = 0
+            cir_t[total_idx] += tmp_a * tmp_x
```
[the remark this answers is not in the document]
noel2004: I think hard-coded would be better and the subscripts are just instinctive indicator. I will update it

src/zkevm_specs/opcode/mul.py
```diff
@@ -0,0 +1,56 @@
...
+        v0m += u8 * (2 ** (8 * i))
+
+    for i, u8 in enumerate(v1):
+        assert 0 <= u8 <= 255
+        v1m += u8 * (2 ** (8 * i))
+
+    assert t0 == cir_t[0]
+    assert t1 == cir_t[1]
+    assert t2 == cir_t[2]
+    assert t3 == cir_t[3]
```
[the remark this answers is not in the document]
noel2004: Ok. So we just put the hard-coded calcultions in the previous comment here and compare them to t0~t3 directly. And for the following lines we can just use t0 ~ t3.

specs/opcode/02MUL.md
noel2004: I will check it. I have simply copied the exported md file from hackmd before.

src/zkevm_specs/opcode/mul.py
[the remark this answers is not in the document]
han0110: Yes, use `cir_t[*]` directly.

specs/opcode/02MUL.md
noel2004: It is the lint / fmt procedure `mdformat --number` adding extra slash into the embedded latex code. It seems we can not use latex syntax inside the documents or it can not pass the lint checking. So I have to provide linking of the references only.

src/zkevm_specs/opcode/mul.py
```diff
@@ -0,0 +1,43 @@
...
+    for i, u8 in enumerate(v1):
+        assert 0 <= u8 <= 255
+        v1m += u8 * (2 ** (8 * i))
+
+    assert t0 == a64s[0] * x64s[0]
+    assert t1 == a64s[0] * x64s[1] + a64s[1] * x64s[0]
+    assert t2 == a64s[0] * x64s[2] + a64s[1] * x64s[1] + a64s[2] * x64s[0]
+    assert t3 == a64s[0] * x64s[3] + a64s[1] * x64s[2] + a64s[2] * x64s[1] + a64s[3] * x64s[0]
+    assert v0m * (2 ** 128) == t0 + t1 * (2 ** 64) - y64s[0] - y64s[1] * (2 ** 64)
+    assert v1m * (2 ** 128) == v0m + t2 + t3 * (2 ** 64) - y64s[2] - y64s[3] * (2 ** 64)
```
han0110, line 38: Sorry for the ambiguity, what I meant in previous comment is actually to remove the constraint and requirement of `t*` as concrete witness, and use `t*` as the expression of sum of limbs product.

```suggestion
    t0 = a64s[0] * x64s[0]
    t1 = a64s[0] * x64s[1] + a64s[1] * x64s[0]
    t2 = a64s[0] * x64s[2] + a64s[1] * x64s[1] + a64s[2] * x64s[0]
    t3 = a64s[0] * x64s[3] + a64s[1] * x64s[2] + a64s[2] * x64s[1] + a64s[3] * x64s[0]
    assert v0m * (2 ** 128) == t0 + t1 * (2 ** 64) - y64s[0] - y64s[1] * (2 ** 64)
    assert v1m * (2 ** 128) == v0m + t2 + t3 * (2 ** 64) - y64s[2] - y64s[3] * (2 ** 64)
```

noel2004: get it, thx

specs/opcode/02MUL.md
noel2004: Finally I turn the original reference into a plain markdown format with unicode symbols.

han0110 (approved):
The updated doc looks great to me!

specs/opcode/02MUL.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Procedure
+
+We followed the approch of [Multi-Limbs Multiplication](https://hackmd.io/HL0QhGUeQoSgIBt2el6fHA), with some notes:
+
+In notation *c = a ⋅ b*, layout could be like the below table if we got 32 columns to use. In table, *a₀, ⋯, a₃₁* as well as *b* and *c* are evm words on stack through bus mapping lookup. Each limb in evm word will be range lookup check to fit 8-bit.
```
ChihChengLiang, line 7: Suggest highlighting the endianness.
```suggestion
In notation *c = a ⋅ b*, layout could be like the below table if we got 32 columns to use. In table, *a₀, ⋯, a₃₁* as well as *b* and *c* are big-endian evm words on stack through bus mapping lookup. Each limb in evm word will be range lookup check to fit 8-bit.
```

tests/test_mul.py
```diff
@@ -0,0 +1,45 @@
...
+        for a_id in range(0, total_idx + 1):
+            a_idx, b_idx = a_id, total_idx - a_id
+            if len(a64s) >= a_idx + 1:
+                tmp_a = a64s[a_idx]
+            else:
+                tmp_a = 0
+            if len(b64s) >= b_idx + 1:
+                tmp_b = b64s[b_idx]
+            else:
+                tmp_b = 0
```
ChihChengLiang, line 32: nitpick for brevity
```suggestion
            tmp_a = a64s[a_idx] if len(a64s) >= a_idx + 1 else 0
            tmp_b = b64s[b_idx] if len(b64s) >= b_idx + 1 else 0
```

ChihChengLiang (commented):
The final request for change before merging

ChihChengLiang (approved):
LGTM

PR #83, opened 2022-01-03
spec for opcode COINBASE

scroll-dev:


specs/opcode/41COINBASE.md
```diff
@@ -0,0 +1,39 @@
...
+
+## EVM behavior
+
+The `COINBASE` opcode loads an `address` (20 bytes of data) from block context.
+then push the `address` to the stack.
+
+## Circuit behavior
+
+1. construct block context table
+2. do bussmapping lookup for stack write operation
```
ChihChengLiang, line 15:
```suggestion
2. do busmapping lookup for stack write operation
```

tests/evm/test_coinbase.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.mark.parametrize("opcode, address", TESTING_DATA)
+def test_coinbase(opcode: Opcode, address: bytes):
+    rlc_store = RLCStore()
+
+    coinbase_address = rlc_store.to_rlc(address)
+
+    bytecode = Bytecode(f"{opcode.hex()}00")
+    bytecode_hash = rlc_store.to_rlc(bytecode.hash, 32)
+    # block = Block(coinbase_address, int(15e6), 10, 0, 0, int(1e9), [])
```
ChihChengLiang, line 29: Forgetting this commented-out code?
```suggestion
```

ChihChengLiang (approved):
LGTM, added some nitpicks

src/zkevm_specs/evm/execution/block_coinbase.py
```diff
@@ -0,0 +1,19 @@
...
+from ..table import BlockContextFieldTag
+from ..opcode import Opcode
+
+
+def coinbase(instruction: Instruction):
+    opcode = instruction.opcode_lookup(True)
+    instruction.constrain_equal(opcode, Opcode.COINBASE)
+    address = instruction.stack_push()
+    # in real circuit also check address raw data is 160 bit length (20 bytes)
+    # check block table for coinbase address
```
han0110, line 11: I assume the block table will somehow be built by verifier (or at least fed into circuit as public input somewhere), so I think EVM circuit could take that it fits in 160 bits.

src/zkevm_specs/evm/execution/__init__.py
```diff
@@ -5,5 +5,6 @@
 from .push import *
 from .jump import *
 from .jumpi import *
+from .block_coinbase import *
```
han0110, line 8: Could we name it after `ExecutionState` (e.g. `coinbase.py`) ?

DreamWuGit: for `coinbase` case is ok, but I consider there are others block's op code (`time`, `number` etc), if use file name directly like `time.py`, `number.py` , easily cause ambiguous meaning comparing to system files.
So I prefer to keep prefix for all , like `block_coinbase.py`, `block_time.py`, `block_number.py` etc.

src/zkevm_specs/evm/execution/block_coinbase.py
DreamWuGit: agree with that, in circuit already use `Address` type to represent it.

tests/evm/test_coinbase.py
DreamWuGit: thanks for catching it .

src/zkevm_specs/evm/execution/__init__.py
han0110: I see, it makes sense to me.

CPerezz:
Should we merge this @han0110 ?

src/zkevm_specs/evm/execution/block_coinbase.py
```diff
@@ -0,0 +1,19 @@
...
+    instruction.constrain_equal(address, instruction.block_lookup(BlockContextFieldTag.Coinbase))
```
han0110, line 12: If we want to keep field `coinbase` in table as actual value instead of random linear combination, we need to do a base conversion here (from `256` to `r`).

```suggestion
    instruction.constrain_equal(
        address,
        instruction.bytes_to_rlc(instruction.int_to_bytes(instruction.block_lookup(BlockContextFieldTag.Coinbase), 20)),
    )
```

tests/evm/test_coinbase.py
```diff
@@ -0,0 +1,68 @@
...
+
+@pytest.mark.parametrize("opcode, address", TESTING_DATA)
+def test_coinbase(opcode: Opcode, address: bytes):
+    rlc_store = RLCStore()
+
+    coinbase_address = rlc_store.to_rlc(address)
+
+    bytecode = Bytecode(f"{opcode.hex()}00")
+    bytecode_hash = rlc_store.to_rlc(bytecode.hash, 32)
+    block = Block(coinbase_address)
```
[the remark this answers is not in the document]
DreamWuGit: I see `Block.table_assignments` will not do rlc internally  in https://github.com/appliedzkp/zkevm-specs/blob/master/src/zkevm_specs/evm/typing.py#L54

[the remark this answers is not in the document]
han0110: Sorry for being ambiguous, what I mean is `Block` always takes actual values as input, the conversion will be done in `table_assignments`. Currently the field `coinbase` is actual value in table, so we will need to do that by constraint.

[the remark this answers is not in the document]
DreamWuGit: understood !  will update it .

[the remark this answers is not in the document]
DreamWuGit: fixed in 164f7de. thanks

src/zkevm_specs/evm/execution/block_coinbase.py
DreamWuGit: fixed in 164f7de. thanks

han0110 (approved):

PR #84, opened 2022-01-06
Calculate the correct number for block hash in table assignment

han0110:


DreamWuGit (approved):
matches logic in https://github.com/ethereum/go-ethereum/blob/master/core/evm.go#L86L88

ChihChengLiang (approved):
LGTM

PR #91, opened 2022-01-14
Add `SLOAD` & `SSTORE` spec

scroll-dev:


HAOYUatHZ:
I am not certain about the following part but happy to refactor:

For gas_refund, currently I always read, but write only persist. Should I make gas_refund "read_only_persist"? so that we can save 1 lookup.

HAOYUatHZ:
close for now. will split into 2 PRs after discussing with Han.

PR #105, opened 2022-01-25
Specs for opcode TIMESTAMP (0x42)

scroll-dev:


DreamWuGit (approved):
LGTM, need to fix build error!

src/zkevm_specs/evm/execution/block_timestamp.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def timestamp(instruction: Instruction):
+    opcode = instruction.opcode_lookup(True)
+    instruction.constrain_equal(opcode, Opcode.TIMESTAMP)
+    timestamp = instruction.stack_push()
+    # in real circuit also check timestamp raw data is 64 bit length (8 bytes)
+    # check block table for timestamp
+    instruction.constrain_equal(
+        timestamp,
```
ed255, line 13: I believe this timestamp requires an RLC "decoding" to match against the value in the block table, like in the implementation: https://github.com/appliedzkp/zkevm-circuits/pull/307/files#diff-797b8e94ba7b72b822bd3d28b64e46ba96d03825cc23dbaad24f55f636aed18bR43-R45

ed255 (changes requested):

src/zkevm_specs/evm/execution/block_timestamp.py
z2trillion: I updated the table assignment of the timestamp to match the implementation here: https://github.com/appliedzkp/zkevm-circuits/blob/d0c04afd8a7b46c4a0451a57eb203d3102d5c7be/zkevm-circuits/src/evm_circuit/witness.rs#L72

PR #106, opened 2022-01-26
add `sload` & `sstore` specs

scroll-dev:
based on:
+ https://eips.ethereum.org/EIPS/eip-2200
+ https://eips.ethereum.org/EIPS/eip-2929


HAOYUatHZ:
just realize that I forget to take call_context_lookup into rw_counter. will update accordingly.

HAOYUatHZ:
> just realize that I forget to take call_context_lookup into rw_counter. will update accordingly.

fixed

HAOYUatHZ:
updated according to https://github.com/appliedzkp/zkevm-specs/pull/108#discussion_r802652988

HAOYUatHZ:
I change lookup for CalleeAddress from TxContextLookup to CallContextLookup

but not sure whether it's the right way?

HAOYUatHZ:
conflict with current master.

Trying to resolve

src/zkevm_specs/evm/execution/storage.py
```diff
@@ -0,0 +1,113 @@
...
+
+    is_warm_new, is_warm = instruction.add_account_storage_to_access_list_with_reversion(
+        tx_id, callee_address, storage_key, is_persistent, rw_counter_end_of_reversion
+    )
+
+    gas_refund, gas_refund_prev = instruction.tx_refund_write_with_reversion(
+        tx_id, is_persistent, rw_counter_end_of_reversion
+    )
+    gas_refund_new = gas_refund_prev
+    if value_prev != value_new:
```
han0110, line 75: I am thinking we could use `is_equal` flag instead of branches, which better reflects circuit, and also better reflects  yellowpaper's long description. But it's okay to just keep as it now, since there are already a lot thing hard to reflect circuit's behavior.

src/zkevm_specs/evm/execution/storage_gas.py
```diff
@@ -0,0 +1,6 @@
+COLD_SLOAD_COST = 2100
+WARM_STORAGE_READ_COST = 100
+SLOAD_GAS = 100
+SSTORE_SET_GAS = 20000
+SSTORE_RESET_GAS = 2900
+SSTORE_CLEARS_SCHEDULE = 15000
```
han0110, line 6: Could we place all parameters to `util/param.py`?

src/zkevm_specs/evm/instruction.py
```diff
@@ -528,6 +544,27 @@ def sub_balance_with_reversion(
...
+    ) -> Tuple[FQ, FQ, FQ, FQ]:
+        row = self.state_write_with_reversion(
+            RWTableTag.AccountStorage,
+            [account_address, storage_key],
+            is_persistent,
+            rw_counter_end_of_reversion,
+            state_write_counter,
+        )
+        return row[-4], row[-3], row[-2], row[-1]
+
```
han0110, line 567: I think it makes more sense to treat `tx_id` as input (as mentioned above), and we don't need to return it.

```suggestion
    def account_storage_write_with_reversion(
        self,
        account_address: int,
        storage_key: int,
        tx_id: int,
        is_persistent: bool,
        rw_counter_end_of_reversion: int,
        state_write_counter: Optional[int] = None,
    ) -> Tuple[FQ, FQ, FQ]:
        row = self.state_write_with_reversion(
            RWTableTag.AccountStorage,
            [account_address, storage_key, Tables._, Tables._, tx_id],
            is_persistent,
            rw_counter_end_of_reversion,
            state_write_counter,
        )
        return row[-4], row[-3], row[-1]

```

src/zkevm_specs/evm/instruction.py
```diff
@@ -528,6 +544,27 @@ def sub_balance_with_reversion(
         self.constrain_zero(carry)
         return balance, balance_prev
 
+    def account_storage_read(self, account_address: int, storage_key: int) -> Tuple[FQ, FQ, FQ, FQ]:
+        row = self.rw_lookup(RW.Read, RWTableTag.AccountStorage, [account_address, storage_key, 0])
+        return row[-4], row[-3], row[-2], row[-1]
```
han0110, line 549:
```suggestion
    def account_storage_read(self, account_address: int, storage_key: int, tx_id: int) -> FQ:
        row = self.rw_lookup(RW.Read, RWTableTag.AccountStorage, [account_address, storage_key, 0, Tables._, Tables._, tx_id])
        return row[-4]
```

han0110 (commented):
Most part looks great! Thanks for handling such opcode with the msot complicated gas calculation.

src/zkevm_specs/evm/execution/storage.py
HAOYUatHZ: opened an issue: https://github.com/appliedzkp/zkevm-specs/issues/123

I will work on it as my following PR

ChihChengLiang (approved):
LGTM

PR #109, opened 2022-01-28
Change field value from int to field numbers

scroll-dev:
Currently values in the Fp field all use int type in Python. There two drawbacks about this:
- There is potential overflow problem with int value
- Writing expression on these values requires to use `fp_add`, `fp_mul`, etc. It is quite annoying and makes the expression hard to understand.

Therefore, this PR replaces the values in the Fp field by `bn128_FQ` in the py-ecc package.

### DIFF
--- a/src/zkevm_specs/evm/instruction.py
+++ b/src/zkevm_specs/evm/instruction.py
@@ -5,7 +5,8 @@ from typing import Optional, Sequence, Tuple, Union
 from ..util import (
     Array4,
     Array10,
-    FpNum,
+    FQ,
+    IntOrFQ,
     RLC,
     MAX_N_BYTES,
     N_BYTES_MEMORY_ADDRESS,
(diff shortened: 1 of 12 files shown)


ed255 (approved):
LGTM!

src/zkevm_specs/evm/instruction.py
```diff
@@ -280,28 +284,33 @@ def rlc_to_int_exact(self, rlc: RLC, n_bytes: int) -> int:
...
     def int_to_rlc(self, value: int, n_bytes: int) -> RLC:
         return RLC(value, self.randomness, n_bytes)
 
     def bytes_to_int(self, value: Sequence[int]) -> int:
         assert len(value) <= MAX_N_BYTES, "Too many bytes to composite an integer in field"
-
         return int.from_bytes(value, "little")
 
-    def range_lookup(self, value: int, range: int):
+    def bytes_to_fq(self, value: Sequence[int]) -> FQ:
```
ChihChengLiang, line 297: should this be bytes?
```suggestion
    def bytes_to_fq(self, value: bytes) -> FQ:
```

icemelon: Updated

PR #110, opened 2022-01-29
Enforce `ExecutionState` transition

han0110:
This PR aims to add missing `ExecutionState` transition constraint. There are 3 of `ExecutionState`s which need to be protected to avoid non-halting opcodes trying to halt unexpectedly. They are:

- `BeginTx` - Can only from `EndTx`
- `EndTx` - Can only from `ExecutionState` which halts or `BeginTx`
- `EndBlock` - Can only from `EndTx` or `EndBlock`

Also it adds some missing `ErrorOutOfGas` cases and remove some unused information.


src/zkevm_specs/evm/instruction.py
```diff
@@ -100,6 +101,14 @@ def constrain_bool(self, value: int):
     def constrain_gas_left_not_underflow(self, gas_left: int):
         self.bytes_range_lookup(gas_left, N_BYTES_GAS)
 
+    def constrain_execution_state_transition(self):
+        if self.next.execution_state == ExecutionState.BeginTx:
+            assert self.curr.execution_state == ExecutionState.EndTx
+        elif self.next.execution_state == ExecutionState.EndTx:
+            assert self.curr.execution_state.halts() or self.curr.execution_state == ExecutionState.BeginTx
```
DreamWuGit, line 108: what is the case of  the `BeginTx:` direct follows by `EndTx` ?

han0110: When `callee` has empty code. But we could also remove this branch if we decide to always dive into call, even if code is empty.

src/zkevm_specs/evm/main.py
```diff
@@ -15,24 +15,26 @@ def verify_steps(
     begin_with_first_step: bool = False,
     end_with_last_step: bool = False,
 ):
-    # TODO: Enforce general ExecutionState transition constraint
+    # For the last step, the next step is meaningless
+    if end_with_last_step:
+        steps += [None]
+
+    for idx in range(len(steps) - 1):
+        curr, next = steps[idx], steps[idx + 1]
```
ChihChengLiang, line 23: We could do something like
```suggestion
    for curr, _next in zip(steps, steps[1:]):
```
This works even if `steps` has 1 element.

ChihChengLiang (approved):
LGTM

src/zkevm_specs/evm/instruction.py
```diff
@@ -107,7 +108,24 @@ def constrain_bool(self, num: FQ):
...
+        if curr == ExecutionState.EndTx:
+            assert next in [ExecutionState.BeginTx, ExecutionState.EndBlock]
+        elif curr == ExecutionState.EndBlock:
+            assert next == ExecutionState.EndBlock
+
+        # Negation ExecutionState transition constraint for rest ones
+        if next == ExecutionState.BeginTx:
+            assert curr == ExecutionState.EndTx
+        elif next == ExecutionState.EndTx:
+            assert curr.halts() or curr == ExecutionState.BeginTx
```
ed255, line 124: I'm wondering if this should be:
```python
assert (curr.halts() and curr.is_root) or curr == ExecutionState.BeginTx)
```

For example, if we're in a call at depth 3 and we RETURN/STOP, `curr.halts()` is true, but the execution continues to the call at depth 2.

src/zkevm_specs/evm/table.py
```diff
@@ -67,21 +63,8 @@ def table_assignments(self) -> Sequence[Array4]:
...
-        elif self == FixedTableTag.StackOverflow:
-            return [
-                (self, opcode, stack_pointer, 0)
-                for opcode, stack_pointer in stack_underflow_pairs()
-            ]
-        elif self == FixedTableTag.StackUnderflow:
-            return [
-                (self, opcode, stack_pointer, 0) for opcode, stack_pointer in stack_overflow_pairs()
-            ]
         else:
```
ed255, line 66: Since you remove these tags from the fixed table, what will be the strategy to prove that that an opcode can be "invalid", be a "state write" or cause a "stack overflow/underflow"?

Also, with this deletion the functions `invalid_opcodes`, `state_write_opcodes`, `stack_underflow_pairs` and `stack_overflow_pairs` are no longer used.  Will they be used elsewhere?  Or can they be removed?

src/zkevm_specs/evm/instruction.py
han0110: The checks here is more like "all possible `ExecutionState` transiton" when we are defining the state transition diagram.

When one can transit to multiple, its constraint should make sure the transition is reasonable. For example, in #100, we constrain something like this ([reference](
https://github.com/appliedzkp/zkevm-specs/pull/100/files#diff-09a605ab107c1b57880e784cd75a70e238f1ebf114bd4bf1d535671bff929f14R15-R17)):

```python
# Go to EndTx only when is_root
is_to_end_tx = instruction.is_equal(instruction.next.execution_state, ExecutionState.EndTx)
instruction.constrain_equal(instruction.curr.is_root + is_to_end_tx, 2 * instruction.curr.is_root * is_to_end_tx)
```

src/zkevm_specs/evm/table.py
han0110: > Since you remove these tags from the fixed table, what will be the strategy to prove that that an opcode can be "invalid", be a "state write" or cause a "stack overflow/underflow"?

I am thinking we can put all these stuff in `Fixed::ResponsibleOpcode`, and make use of the unused fourth column as auxiliary data. I was planning to add them in the following PR, but let me add them now to make it clear.

For invalid opcodes, the `ExecutionState::ErrorInvalidOpcode` should be responsible for them (weird but I think it somehow makes sense).

For state write opcodes, we only use them when `ExecutionState::ErrorWriteProtection`, so it's responsible for them.

For stack overflow/underflow, we can reduce to a single state `ExecutionState::ErrorStack`, with the `stack_pointer` in fourth column.

> Also, with this deletion the functions `invalid_opcodes`, `state_write_opcodes`, `stack_underflow_pairs` and `stack_overflow_pairs` are no longer used. Will they be used elsewhere? Or can they be removed?

I am going to add those pairs mentioned above into `Fixed::ResponsibleOpcode` sub-table, then these fns will be used.

src/zkevm_specs/evm/main.py
han0110: I just found we still need the `idx`, so I keep the current version. Also in #116, we can't extend the `steps` with a `None`, so iterating on range seems easier.

src/zkevm_specs/evm/table.py
```diff
@@ -63,25 +53,15 @@ def table_assignments(self) -> Sequence[Array4]:
             return [(self, lhs, rhs, lhs ^ rhs) for lhs, rhs in product(range(256), range(256))]
         elif self == FixedTableTag.ResponsibleOpcode:
             return [
-                (self, execution_state, opcode, 0)
+                (self, execution_state, opcode, aux, opcode, aux)
```
ed255, line 56:
```suggestion
                (self, execution_state, opcode, aux)
```
Typo maybe?

ed255:
A part from the possible typo mentioned in https://github.com/appliedzkp/zkevm-specs/pull/110#discussion_r815743749 the rest LGTM!

PR #115, opened 2022-02-10
upgrade and pin dependencies

ChihChengLiang:
### Problem

The package black has a new update released, causing some confusion. The CI checks the formatting with the new version but the developers' local old versions disagree with the formatting.

### Solution

We need to pin the versions of the packages to prevent CI from using a new version that we haven't explicitly upgraded.

Note: Run `make install` after this PR is merged.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ where = src
 
 [options.extras_require]
 test =
-    pytest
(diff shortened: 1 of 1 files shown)


HAOYUatHZ (approved):

ed255 (approved):
LGTM!

PR #117, opened 2022-02-14
Change line-length from 120 to 100

ChihChengLiang:
### Problem

120 is too long, change it to 100 to be consistent with the circuit repo.


tests/evm/test_begin_tx.py
```diff
@@ -97,7 +112,18 @@ def test_begin_tx(tx: Transaction, callee: Account, result: bool):
         bytecode_table=set(callee.code.table_assignments(randomness)),
         rw_table=set(
             [
-                (1, RW.Read, RWTableTag.CallContext, 1, CallContextFieldTag.TxId, 0, tx.id, 0, 0, 0),
+                (
```
ed255, line 115: Do you think it's acceptable to skip the format check for test vectors? Like so: https://github.com/appliedzkp/zkevm-specs/blob/45842edfcdeda4744cf5414728c4d32a4a677473/tests/test_state_circuit.py#L27

I think having this exception can make the test code more readable.

ed255 (approved):
LGTM! Please take a look at the comment for a discussion about formatting skipping for test vectors.

tests/evm/test_begin_tx.py
ChihChengLiang: That sounds good! Will do.

PR #118, opened 2022-02-14
Redo spec of state circuit

ed255:
- Update the state proof document
- Write the spec of the state circuit in python from scratch following
  the current `rw_table` design.
- Add a minimal state circuit test
- List of changes:
    - Renamed key and aux columns so start with index 0, in order to
      make it more readable when working with code were we have arrays
      called `keys`.  This way `key0` == `keys[0]`.
    - Add a new key column so that we have a cloumn exclusive for
      address keys, which allows us to reuse the key order constraints
      for addresses for all operations.
    - Add the boilerplate for the missing operations
    - Remove value2 column (used for `valuePrev`) in some Write
      operations

### DIFF
--- a/specs/state-proof.md
+++ b/specs/state-proof.md
@@ -1,376 +1,47 @@
 # State Proof
 
-The state proof helps EVM proof to check all the random read-write access records are valid, through grouping them by their unique index first, and then sorting them by order of access. We call the order of access `GlobalCounter`, which counts the number of access records and also serves as an unique identifier for a record. When state proof is generated, the `BusMapping` is also produced and will be shared to EVM proof as a lookup table.
+The state proof helps EVM proof to check all the random read-write access records are valid, through grouping them by their unique index first, and then sorting them by order of access. We call the order of access `ReadWriteCounter`, which counts the number of access records and also serves as an unique identifier for a record. When state proof is generated, the `BusMapping` is also produced and will be shared to EVM proof as a lookup table.
 
 ## Random Read-Write Data
 
-State proof maintains the read-write part of [random accessible data](./evm-proof.md#Random-Accessible-Data) of EVM proof:
+State proof maintains the read-write part of [random accessible data](./evm-proof.md#Random-Accessible-Data) of EVM proof.
 
-| Target                                | Index             | Description                              |
-| ------------------------------------- | ----------------- | ---------------------------------------- |
-| [`AccountNonce`](#AccountNonce)       | `{address}`       | Account's nonce                          |
-| [`AccountBalance`](#AccountBalance)   | `{address}`       | Account's balance                        |
-| [`AccountCodeHash`](#AccountCodeHash) | `{address}`       | Account's code hash                      |
-| [`AccountStorage`](#AccountStorage)   | `{address}.{key}` | Account's storage as a key-value mapping |
-| [`CallState`](#CallState)             | `{id}.{enum}`     | Call's internal state                    |
-| [`CallStateStack`](#CallStateStack)   | `{id}.{index}`    | Call's stack as a encoded word array     |
-| [`CallStateMemory`](#CallStateMemory) | `{id}.{index}`    | Call's memory as a byte array            |
+The operations recorded in the state proof are:
 
-The concatenation of **Target** and **Index** becomes the unique index for data. Each record will be attached with a `GlobalCounter`, and the records are constraint to be in group by their unique index first and to be sorted by their `GlobalCounter` increasingly. Given the access to previous record, each target has their own format and rules to update, for example, `AccountNonce` always increase by 1, and values in `CallStateMemory` should fit in 8-bit.
+- `Memory`: Call's memory as a byte array
+- `Stack`: Call's stack as RLC-encoded word array
+- `Storage`: Account's storage as key-value mapping
+- `CallContext`: Context of a Call
+- `Account`: Account's state (nonce, balance, code hash)
+- `TxRefund`: Value to refund to the tx sender
+- `TxAccessListAccount`: State of the account access list
+- `TxAccessListAccountStorage`: State of the account storage access list
+- `AccountDestructed`: State of destruction of an account
 
-## Constants
+Each operation uses diferent parameters for indexing.  See [RW Table](./tables.md#rw_table) for the complete details.
 
-| Name                 | Value              | Description                                                                               |
-| -------------------- | ------------------ | ----------------------------------------------------------------------------------------- |
-| `MAX_U8`             | `2**8 - 1 (255)`   | Maximum value of u8 (byte)                                                                |
-| `MAX_STACK_INDEX`    | `2**10 - 1 (1023)` | Maximum index of stack (EVM has maximum stack size 1024)                                  |
-| `MAX_GLOBAL_COUNTER` | `2**25 - 1`        | Maximum number of `GlobalCounter`, which is also the maximum degree of polynomial allowed |
+The concatenation of all table keys becomes the unique index for data. Each record will be attached with a `ReadWriteCounter`, and the records are constraint to be in group by their unique index first and to be sorted by their `ReadWriteCounter` increasingly. Given the access to previous record, each target has their own format and rules to update, for example, values in `Memory` should fit in 8-bit.
 
 ## Circuit Constraints
 
-The following part describes the custom constraints for each target by a python script. There are some common helper class and function:
-
-```python
-class RW(Enum):
-    Read: int = 0
-    Write: int = 1
-
-    def __sub__(self, _):
-        return None
-
-
-class Record:
-    fields: Tuple[str]
-    values: object
-
-    def __init__(self, **kwargs) -> None:
-        self.values = {}
-        self.fields = tuple(["global_counter", "rw"] + self.fields)
-        for field in self.fields:
-            self.values[field] = kwargs[field]
-            setattr(self, field, self.values[field])
-
-    def __sub__(self, rhs):
-        if rhs is None:
-            return self
-        return type(self)(
-            **{field: self.values[field] - rhs.values[field] for field in self.fields}
-        )
-
-
-class ReadWriteGate:
-    group_fields: Tuple[(str, Callable)]
-
-    def constraint(self, prev, cur):
-        diff = cur - prev
-
-        is_first_row = prev is None
-        is_first_row_in_group = is_first_row or any(
-            [getattr(diff, field) != 0 for field, _ in self.group_fields]
-        )
-
-        # grouping
-        if not is_first_row:
-            for field, group_fn in self.group_fields:
-                assert group_fn(cur, prev, diff)
-                if getattr(diff, field) != 0:
-                    break
-
-        # rw should be valid
-        assert cur.rw in (RW.Read, RW.Write)
-
-        # global counter should increase in group
-        if not is_first_row_in_group:
-            assert 0 < diff.global_counter <= MAX_GLOBAL_COUNTER
-
-        type(self).constraint_every_row(cur)
-        type(self).constraint_in_group(prev, cur, diff, is_first_row_in_group)
-
-    @abstractclassmethod
-    def constraint_every_row(cur):
-        raise NotImplemented
-
-    @abstractclassmethod
-    def constraint_in_group(prev, cur, diff, is_first_row_in_group):
-        raise NotImplemented
-```
-
-### `AccountNonce`
-
-| Field     | Description     |
-| --------- | --------------- |
-| `address` | Account address |
-| `nonce`   | Account nonce   |
-
-```python
-class AccountNonce(Record):
-    fields = ["address", "nonce"]
-
-
-class AccountNonceGate(ReadWriteGate):
-    group_fields = [
-        # TODO: check address by 160-bit range lookup or other more efficient method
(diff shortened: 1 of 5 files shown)


src/zkevm_specs/state.py
```diff
@@ -0,0 +1,540 @@
...
+    assert row.keys[3] == 0
+
+    # 1. First access for a set of all keys
+    #
+    # We add an extra write to set the value of the state in previous block, with rwc=0.
+    #
+    # When the set of all keys changes (first access of storage (address, key))
+    # - It must be a WRITE
+    if not all_keys_eq(row, row_prev):
+        assert row.is_write == 1 and row.rw_counter == 0
```
miha-stopar, line 146: This is still to be finished, right? Like adding constraints for previous and current value (to enable revert).

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,540 @@
...
+    # 6. Read consistency
+    #
+    # When a row is READ
+    # AND When all the keys are equal in two consecutive a rows:
+    # - The corresponding value must be equal to the previous row
+    if row.is_write == 0 and all_keys_eq(row, row_prev):
+        assert row.value == row_prev.value
+
+    # 7. Padding rows
+    # TODO: Figure out if we really need padding rows
```
miha-stopar, line 309: Hm, I forgot all the reasons for padding, but one of the problems without padding would probably be to ignore the constraints about the increasing counter because the address doesn't change in empty rows.

miha-stopar (commented):
Nice!! Only a couple of nitpicks and questions.

src/zkevm_specs/state.py
ed255: @han0110 mentioned that we can remove the value_prev column.  See here https://hackmd.io/VCGteyZEQFKUjHA_A72f1Q for the current documentation of the state circuit design and here https://github.com/appliedzkp/zkevm-specs/pull/94#discussion_r788627623 for a relevant discussion.  So in this PR there's a single value column.

In the spec from this PR, there's a constraint for all READ rows which don't change keys, for which the value must equal the previous value; and no constraint for WRITE rows.  I think that works fine for reverts, as a revert does a WRITE and from the state circuit perspective, this write can be anything.  I think it will be the EVM circuit that enforces that the writes from a revert restore the correct value.

@miha-stopar What constraints in particular were you thinking about?

specs/state-proof.md
```diff
@@ -1,376 +1,47 @@
...
+values we use range checks of the difference between the consecutive cells,
+with the help of fixed lookup tables.  Since we use lookup tables to prove
+correct ordering, for every column that must be sorted we need to define the
+maximum value it can contain (which will correspond to the fixed lookup table
+size); this way, two consecutive cells in order will have a difference that is
+found in the table, and a reverse ordering will make the difference to wrap
+around to a very high value (due to the field arithmetic), causing the result
+to not be in the table.
 
+The exact list of constraint is documented in detail as comments in the python
```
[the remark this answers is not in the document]
ed255: fixed, thanks!

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,540 @@
...
+    assert min_val <= x.n and x.n <= max_val
+
+
+@is_circuit_code
+def check_memory(row: Row, row_prev: Row):
+    get_call_id = lambda row: row.keys[1]
+    get_mem_addr = lambda row: row.keys[2]
+
+    # 0. Unused keys are 0
+    assert row.keys[2] == 0
```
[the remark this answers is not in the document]
ed255: thanks for spotting this!  This assert was correct, but the definition of `get_mem_addr` was wrong as it should be key3.  Fixed!

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,540 @@
...
+
+class Row(NamedTuple):
+    """
+    State circuit row
+    """
+
+    # fmt: off
+    rw_counter: FQ
+    is_write: FQ # boolean
+    keys: Tuple[FQ, FQ, FQ, FQ, FQ] # key2 is 160bit Address.  key4 is RLC encoded
```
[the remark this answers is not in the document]
ed255: I've added a little description with a link to the table.  It's hard to tell what each column is because it changes per operation, so I added the table link which contains the complete information.

ed255:
Another doubt I have for the state circuit is how to extract the number of operations that have happened, so that the prover can't introduce extra operations that are not used in the EVM circuit.
1. What forbids adding 2 rows of different operations with the same `rw_counter`?
2. How does the EVM circuit know the number of operations (the highest rwc_counter value)?

We must take into account that for State Tree operations we're introducing artificial writes with `rw_counter=0` which must not be accounted for, so the number of rows is different than the number of operations.

One idea that comes to my mind is to add an advice column that counts the number of rows with `rw_counter != 0`.  Then in the padding rows, constraint that `rw_counter == 0`.  This way the EVM circuit can do a lookup to any padding row to read the number of operations from this new column.

What do people think of this?

barryWhiteHat:
> What forbids adding 2 rows of different operations with the same rw_counter?

NOthing but the below check confirms that every row has to be used by the evm proof. 

> How does the EVM circuit know the number of operations (the highest rwc_counter value)?

I was thinking to count the number of elements in the RW table in the zkevm and then the l1 smartcontract confirm that the polynomial contains exactly that many elements. I know Han has been thinking about this more so maybe has a better idea :)

src/zkevm_specs/state.py
miha-stopar: I see, wasn't up to date. Thanks!

han0110:
> How does the EVM circuit know the number of operations (the highest rwc_counter value)?

Originally Barry proposes to use degree check (by underlying PCS) on the `rw_table`, but I think that it needs prover to provide an shuffle argument to argue that `(low_degree_rw_table || zero_padding) == shuffle(rw_table)`, because we might have some artificial row that exist in between meaningful rows that `rw_counter != 0`. (e.g. to prove `[(rw_counter=1, ...), (rw_counter=0, ...), (rw_counter=30, ...)]` have only 2 meaningful elements, prover needs to provide `[(rw_counter=1, ...), (rw_counter=30, ...)]`, and verifier verifies it has degree 1, and the padded version of it is also a shuffle of the `rw_table` generated by State circuit).

My naive thought is to let both State circuit and EVM circuit to consume the same public input `total_rw`, and the former verifies it matches the sum of row that `rw_counter != 0`, the latter verifies it matches the latest step's `rw_counter`. I think this approach is easier to implement (already supported by current `halo2`), but not sure if it's sound as the previous approach or not.

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,544 @@
...
+        )
+
+    keys = get_keys_in_order(row)
+    keys_prev = get_keys_in_order(row_prev)
+    keys_eq = True
+    for i in range(len(keys) - 1):
+        keys_eq = keys_eq and keys[i] == keys_prev[i]
+        if keys_eq:
+            key_diff = keys[i + 1] - keys_prev[i + 1]
+            assert_in_range(key_diff, 0, MAX_KEY_DIFF)
```
han0110, line 294: Because getting `is_eq = keys[i] == keys_prev[i]` requires an inverse, perhaps we can recomposition `keys` into larger limbs and get the chain of `(is_gt, is_eq)` of `key_diff`, and finally require transition of the chain to be `is_eq -> is_eq or is_gt` and `is_gt -> any`.

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,544 @@
...
+
+    # 5. RWC is monotonically strictly increasing for a set of all keys
+    #
+    # When all the keys are equal in two consecutive a rows:
+    # - The corresponding rwc must be strictly increasing.
+    if all_keys_eq(row, row_prev):
+        rw_diff = row.rw_counter - row_prev.rw_counter
+        assert_in_range(rw_diff, 0, MAX_KEY_DIFF)
```
han0110, line 302: If "monotonically strictly", should the `min` here be `1`?

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,544 @@
...
+MAX_MEMORY_ADDRESS = 2**32 - 1
+MAX_STACK_PTR = 1023
+
+
+class Tag(IntEnum):
+    """
+    Tag used as first key in the State Circuit Rows to "select" the operation target.
+    """
+
+    Start = 1
```
han0110, line 18: Is `Start` aim for avoiding the "previous row wraps around" issue? It seems we can leverage this as padding (prover adds as many `Start` rows in the beginning with `rw_counter = 0` as padding).

han0110 (commented):
All look great, only with some outdated docs are wrong and should be removed.

barryWhiteHat:
> My naive thought is to let both State circuit and EVM circuit to consume the same public input total_rw, and the former verifies it matches the sum of row that rw_counter != 0, the latter verifies it matches the latest step's rw_counter. I think this approach is easier to implement (already supported by current halo2), but not sure if it's sound as the previous approach or not.

At the point when we are implmenting the l1 verifier we can consider these two options. For now as long as we have a solution we are confident will work i think we are good to go.

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,544 @@
...
+    if not all_keys_eq(row, row_prev):
+        assert row.is_write == 1 and row.rw_counter == 0
+
+    # 2. nonce
+    if get_field_tag(row) == AccountFieldTag.Nonce:
+        get_nonce = lambda row: row.value
+        # When all the keys are equal in two consecutive a rows:
+        # - if WRITE, Nonce increases by 1
+        if all_keys_eq(row, row_prev) and row.is_write == 1:
+            assert get_nonce(row) - get_nonce(row_prev) == 1
```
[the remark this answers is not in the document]
han0110: Maybe not, because in EVM circuit the constraint already does that explicitly, like:

https://github.com/appliedzkp/zkevm-specs/blob/d7bfc7ccfea8ffd2c38f7ebb5fe655cf84f9eda1/src/zkevm_specs/evm/execution/begin_tx.py#L33

src/zkevm_specs/state.py
```diff
@@ -0,0 +1,544 @@
...
+    #
+    # Constraints that affect all rows, no matter which Tag they use
+    #
+
+    # 0. Tags transition
+    #
+    # Tag monotonically increases for all values except for Start
+    if row.tag() != Tag.Start:
+        tag_diff = row.tag() - row_prev.tag()
+        assert_in_range(tag_diff, 0, MAX_KEY_DIFF)
```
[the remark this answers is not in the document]
ed255: Can we be sure that all blocks will use all tags always?
If that's not the case, then there could be an increment of more than 1.  Assuming that was the case, I just went to use `MAX_KEY_DIFF` to avoid introducing a new range lookup table.  I'm not familiar with the cost of A vs B:
- A. 1 big fixed lookup table with 2 lookups happening
- B. 1 big fixed lookup table with 1 lookup + 1 small fixed lookup table with 1 lookup

src/zkevm_specs/state.py
ed255: Yes you're right!  I made a mistake there.

src/zkevm_specs/state.py
ed255: > Is `Start` aim for avoiding the "previous row wraps around" issue? 

Yes exactly!

> It seems we can leverage this as padding (prover adds as many `Start` rows in the beginning with `rw_counter = 0` as padding).

That's an interesting idea!  So we pad the beginning rows with Start as necessary, and also use Start to not check rules that would fail due to the wrap around with the end of the table.

src/zkevm_specs/state.py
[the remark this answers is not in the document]
han0110: I see, it makes sense to me now.

Since we already have different range lookup, we would somehow have tag for fixed table to reuse the fixed columns. If we do so, then cost of A vs B is same.

src/zkevm_specs/state.py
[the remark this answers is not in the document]
ed255: Ah, in that case this constraint can be just removed then?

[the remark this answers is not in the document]
han0110: Yeah, and I guess for balance and storage we don't have special constraint, so perhaps all of them could be omitted.

src/zkevm_specs/state.py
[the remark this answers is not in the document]
ed255: I have added constraints for each key to be in the expected range (including key0 = tag) in https://github.com/appliedzkp/zkevm-specs/compare/b145f36feaab9db33f41c4cd14251dc961b7609d..0258fe56111ed78f653b02a51d08093ec05958d0

src/zkevm_specs/state.py
ed255: From a discussion with Han, we can reuse the Start tag as padding (padding at the beginning instead of at the end)

ed255:
I have added tests for situations that don't pass the constraints.

Should I update `tables.py` to remove the `value_prev` column in the `rw_table`?

han0110 (approved):
LGTM!

miha-stopar (approved):
LGTM! Thanks!

PR #125, opened 2022-02-18
add logs spec 

DreamWuGit:
introduction see https://hackmd.io/fj9yRzrfRkyKgcrw5xBAgg?view

change following items:
add TxLog field tag of RW table
add logs instruction with test
add gas util for tests which involves memory expansion gas use case (mostly copied from src/zkevm_specs/opcode/mload_mstore.py for now)
add internal `CopyToLog` gadget spec according to circuit implementation


src/zkevm_specs/evm/execution/log.py
```diff
@@ -0,0 +1,60 @@
...
+    contract_address = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
+    instruction.constrain_equal(contract_address, instruction.tx_log_lookup(TxLogFieldTag.Address))
+
+    # omit block number constraint even it is set within op code explicitly, because by default the circuit only handle
+    # current block, otherwise, block context lookup is required.
+    # calculate dynamic gas cost
+    next_memory_size, memory_expansion_cost = instruction.memory_expansion_constant_length(mstart, msize)
+    dynamic_gas = LOG_STATIC_GAS * (opcode - Opcode.LOG0) + 8 * msize + memory_expansion_cost
+
+    rw_counter_diff = 5 + 2 * (opcode - Opcode.LOG0) + 2 * msize
```
ChihChengLiang, line 50: Can we add some comments to explain the number?

src/zkevm_specs/util/gas.py
```diff
@@ -0,0 +1,41 @@
...
+def div(
+    value: U256,
+    divisor: U64,
+) -> Tuple[U256, U256]:
+    quotient = value // divisor
+    remainder = value % divisor
+    return (quotient, remainder)
+
+
+def memory_expansion(
```
ChihChengLiang, line 22: No actions for the PR, just a note for other reviewers:
This part of the code seems duplicated with the one in [mload_mstore](https://github.com/appliedzkp/zkevm-specs/blob/28b299fed41e8a1180aef8033290c94702ff649f/src/zkevm_specs/opcode/mload_mstore.py#L78). There's also a similar one in the `instruction.py`.
IIUC, unlike the other places, where the memory_expansion is used as circuit code, this particular memory_expansion function is only used as utilities in the tests.
We could merge them together somehow in other future PR.

tests/evm/test_log.py
```diff
@@ -0,0 +1,187 @@
...
+            + construct_data_rows(3 + 2 * topic_count + msize, data, randomness)
+            + [
+                # demo rows of one topic with two data
+                # (3, RW.Read, RWTableTag.Stack, 1, 1022, 0, RLC(topic1, randomness, 32), 0, 0, 0),
+                # (4, RW.Read, RWTableTag.Memory, 1, 11, 0, 10, 0, 0, 0),
+                # (5, RW.Read, RWTableTag.Memory, 1, 12, 0, 20, 0, 0, 0),
+                # write tx log with topic and data
+                # (6, RW.Write, RWTableTag.TxLog, 0, 0, TxLogFieldTag.Topics, RLC(topic1, randomness, 32), 0, 0, 0),
+                # (7, RW.Write, RWTableTag.TxLog, 0, 0, TxLogFieldTag.Data, 10, 0, 0, 0),
+                # (8, RW.Write, RWTableTag.TxLog, 0, 1, TxLogFieldTag.Data, 20, 0, 0, 0),
```
ChihChengLiang, line 71: Are these intended to use or should they be removed?

tests/evm/test_log.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@pytest.mark.parametrize("topics, mstart, msize", TESTING_DATA)
+def test_log(topics: list, mstart: U64, msize: U64):
+    randomness = rand_fp()
+    block = Block()
+
+    data = []
+    for i in range(msize):
+        data.append(rand_range(255))
```
ChihChengLiang, line 41: nitpick
```suggestion
    data = [rand_range(255) for _ in range(msize)]
```

ChihChengLiang (approved):
LGTM modulo comments + rebase.

src/zkevm_specs/util/gas.py
DreamWuGit: yes,  but method in `instruction.py` can not use in the test module,  adding it into `util` enable all related tests can take use of it directly.

tests/evm/test_log.py
DreamWuGit: it is not using but for illustrating what's the `RW` table look like after adding `TxLog`.  let me know if you think it is non sense

src/zkevm_specs/evm/execution/log.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def log(instruction: Instruction):
+    opcode = instruction.opcode_lookup(True)
+    # constrain op in [log0, log4] range
+    instruction.range_lookup(opcode - Opcode.LOG0, 5)
+
+    # pop `mstart`, `msize` from stack
+    mstart = instruction.rlc_to_int_exact(instruction.stack_pop(), 8)
+    msize = instruction.rlc_to_int_exact(instruction.stack_pop(), 8)
```
icemelon, line 14: this is dynamic size. So you need to use multiple steps to copy the memory, similar to `calldatacopy`.

icemelon (changes requested):

src/zkevm_specs/util/gas.py
ChihChengLiang: makes sense, the current approach sounds good.

tests/evm/test_log.py
ChihChengLiang: Keeping them seems fine 👍

src/zkevm_specs/evm/execution/log.py
DreamWuGit: adding in specs/opcode/A0_A4LOG.md

src/zkevm_specs/evm/execution/log.py
DreamWuGit: thanks , fixed in [93e57b2](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/93e57b2ef33976b72f57bec7b0ed1b4fb43046f0)

DreamWuGit:
rebasing and refactoring done accordance with latest circuit behavior,  @icemelon @ChihChengLiang  could you have a look again ?  thanks

src/zkevm_specs/evm/table.py
```diff
@@ -212,6 +218,27 @@ class CallContextFieldTag(IntEnum):
...
+    # State circuit directly.
+    Address = auto()  # address of the contract that generated the event
+    Topics = auto()  # list of topics provided by the contract
+    Data = auto()  # log data in bytes
+    BlockNumber = auto()  # block number which logs are belong to
+    TxHash = auto()  # hash of the transaction
+    TxIndex = auto()  # index of the transaction in the block
+    # The Removed field is true if this log was reverted due to a chain reorganisation.
+    # it is intermediate value, hence not count for state.
+    # Removed = auto()
```
ChihChengLiang, line 237: Do we mean to remove `Removed`?

src/zkevm_specs/evm/execution/copy_to_log.py
```diff
@@ -0,0 +1,46 @@
+from ...util import FQ, N_BYTES_MEMORY_SIZE
+from ..execution_state import ExecutionState
+from ..instruction import Instruction, Transition
+from ..step import CopyToLogAuxData
+from ..table import RW, TxLogFieldTag
+from ..util import BufferReaderGadget
+
+
+MAX_COPY_BYTES = 71
```
ChihChengLiang, line 9: Why is it 71 bytes? Can we comment this?

ChihChengLiang (commented):
I added some quick feedback on the new changes.

src/zkevm_specs/evm/table.py
```diff
@@ -18,6 +18,7 @@ class FixedTableTag(IntEnum):
     table.
     """
 
+    Range5 = auto()  # value, 0, 0
```
CPerezz, line 21: Why do we need a RangeCheck of [0,4]?? I'm not seeing it.

src/zkevm_specs/util/gas.py
```diff
@@ -0,0 +1,43 @@
+from ..encoding import U8, U64, U128, U256
+from typing import Sequence, Tuple
+
+G_MEM = 3
```
CPerezz, line 4: Doc constants please.

src/zkevm_specs/evm/table.py
DreamWuGit: `Removed` field is an Intermediate variable to event subscriber, which only be true during chain reorganisation.  you can refer to https://github.com/ethereum/go-ethereum/blob/master/core/types/log.go#L53L55.  and blog https://ethereum.stackexchange.com/questions/77871/web3-ethereum-log-subscriptions-removed-field
in other words, it will always be false when it is in the block time. so by default I think we don't need to care about it as we focus on the final block.  does it make sense ?

src/zkevm_specs/evm/execution/copy_to_log.py
DreamWuGit: it the same to `memory_copy` , move it to param.py  and make it common use for both `copy_to_log`&`memory_copy` in [87f20da](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/87f20da4503b3f17e7fc865186616092570d9415)

specs/opcode/A0_A4LOG.md
```diff
@@ -0,0 +1,77 @@
...
+## EVM behavior
+
+1. LogN first pops two elements `mStart`, `mSize` from stack, then
+   pops `N` topics from stack.
+2. Copy memory data from `mStart` address with `mSize` length
+3. Construct log entry and append to log collections of state
+
+## Circuit behavior
+
+Because the `msize` is dynamic, it requires multiple step slots to fully verify the `CALLDATACOPY`.
```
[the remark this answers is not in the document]
DreamWuGit: updated circuit behavior in [5f690db](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/5f690dbd63cd83b87a9f406c2ee14682c04921b7) to make it more clear.  `dynamic` means one or more `CopyToLog` gadgets used.

src/zkevm_specs/evm/execution/calldatacopy.py
```diff
@@ -41,6 +41,7 @@ def calldatacopy(instruction: Instruction):
         instruction.constrain_equal(next_aux.src_addr_end, call_data_length + call_data_offset)
         instruction.constrain_equal(next_aux.from_tx, instruction.curr.is_root)
         instruction.constrain_equal(next_aux.tx_id, tx_id)
+        instruction.constrain_equal(next_aux.bytes_left, length)
```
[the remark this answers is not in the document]
DreamWuGit: make sense !

src/zkevm_specs/evm/step.py
```diff
@@ -40,6 +40,10 @@ class StepState:
     memory_size: FQ
     state_write_counter: FQ
 
+    # log index of current block, this field maybe moved if we find them
+    # not often used.
+    log_index: int
```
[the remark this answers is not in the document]
DreamWuGit: I understand `aux_data` fields are used as constant parameter,  `log_index` will change per any logs step(log0--log4), putting in step state we can use `constrain_step_state_transition` constrain its change, let me know if you also prefer that.

src/zkevm_specs/evm/table.py
DreamWuGit: putting all logs (log0--log4) codes into one `log` code, to this range check to ensure `log` code within range.

CPerezz: Ohhh I see.
Hmmm I guess in the Rust impl we will probably use const-generics. But not sure that it will mean any change for that case.

So far looks correct to have this check here within this implementation.

specs/opcode/A0_A4LOG.md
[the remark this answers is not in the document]
CPerezz: Thanks a lot!! Sorry to be annoying. But the clearer, simpler and easier we write all this, the easier will be for others to get up to speed with it :)

src/zkevm_specs/evm/execution/calldatacopy.py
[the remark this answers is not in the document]
CPerezz: Thanks a lot! <3

src/zkevm_specs/evm/step.py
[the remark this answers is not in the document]
CPerezz: But `aux_data` is defined as:
```python
# Auxilary witness data needed by gadgets
aux_data: Any
```

So either the comment on the code is wrong, or the values are not constant as this contains "witness data".

WDYT @DreamWuGit @ed255 ?

src/zkevm_specs/evm/table.py
```diff
@@ -297,6 +324,11 @@ def tx_lookup(self, inputs: Sequence[int]) -> Array4:
         assert len(inputs) <= 4
         return _lookup("tx_table", self.tx_table, inputs)
 
+    def tx_log_lookup(self, inputs: Sequence[int]) -> Array4:
+        assert len(inputs) <= 10
```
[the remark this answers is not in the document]
DreamWuGit: tx log are in RW table with width 10, but  found this method is not used , removing it .

src/zkevm_specs/evm/step.py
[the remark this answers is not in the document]
DreamWuGit: oh,  I understood  you mean put `log_index` into `aux_data`,   them `aux_data` fields are constant witnesses, like   `CopyToMemoryAuxData`,  `CopyToLogAuxData:
    src_addr: FQ
    bytes_left: FQ
    src_addr_end: FQ`
```
class CopyToMemoryAuxData:
    src_addr: FQ
    dst_addr: FQ
    bytes_left: FQ
    src_addr_end: FQ
    from_tx: FQ
    tx_id: FQ.     
........
class CopyToLogAuxData:
    src_addr: FQ
    bytes_left: FQ
    src_addr_end: FQ
```. 
it will be hard to check if the 'log_index' change correctly when execution.

File: src/zkevm_specs/evm/table.py (hunk 6 above)
Comment (ChihChengLiang) [reply src/zkevm_specs/evm/table.py:237]: That makes sense. Since it's always false inside the circuit, we are safe to remove `Removed`?
```suggestion
```

DreamWuGit: yeah, I will remove it , thanks

src/zkevm_specs/evm/step.py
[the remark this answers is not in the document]
CPerezz: Hmmm I see. 
No worries then. I was just trying to figure out if there was a way to not need new structures. Or fit them inside ones that we already use to reduce complexity :)

ed255 (commented):
A part from some notes I added, I would like to discuss the possibility of revert:  If a log is generated during a call that reverts, will the log be reverted? (that is, removed from the logs of the block).  If so, what's the strategy to handle this case?  Two ideas come to my mind:
1. If `not is_persistent`, all log operations are reverted.  This would be very inconvenient because we need to "play back" all the `CopyToLog` steps, and backtrack the `log_index`
2. If `not is_persistent`, just don't generate the logs operations.  I think that the `LOG*` opcodes don't have any side effects observable within the execution context in the current block right?  So that should be doable.

Please let me know what you think.

src/zkevm_specs/evm/step.py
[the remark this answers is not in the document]
DreamWuGit: @ed255 thanks  it can be a way, but so far I didn't come out how to exactly constrain it in the state circuit side, so put it in the step for now.  glad to hear more ideas!

src/zkevm_specs/evm/table.py
```diff
@@ -212,6 +218,27 @@ class CallContextFieldTag(IntEnum):
...
+    """
+
+    # The following are write-only data inside a transaction, they will be written in
+    # State circuit directly.
+    Address = auto()  # address of the contract that generated the event
+    Topics = auto()  # list of topics provided by the contract
+    Data = auto()  # log data in bytes
+    BlockNumber = auto()  # block number which logs are belong to
+    TxHash = auto()  # hash of the transaction
+    TxIndex = auto()  # index of the transaction in the block
```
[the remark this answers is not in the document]
DreamWuGit: for `BlockNumber` I add comment in `log.py` as following:   
 `omit block number constraint even it is set within op code explicitly, because by default the circuit only handle
    # current block, otherwise, block context lookup is required`. 
Yes,  for `TxIndex ` `TxHash`,  I am planning to remove it as well as `Removed` field since we use `log_index` to identify definite log entry,  then the tx info(tx_hash, tx_id) are also fixed.

src/zkevm_specs/util/gas.py
icemelon: this constant is already define at [here](https://github.com/appliedzkp/zkevm-specs/blob/master/src%2Fzkevm_specs%2Futil%2Fparam.py#L70)

src/zkevm_specs/evm/execution/log.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # check contract_address in CallContext & TxLog
+    # use call context's  callee address as contract address
+    contract_address = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
+    instruction.constrain_equal(contract_address, instruction.tx_log_lookup(TxLogFieldTag.Address))
+    # check not static call
+    instruction.constrain_equal(0, instruction.call_context_lookup(CallContextFieldTag.IsStatic))
+
+    # constrain topics in stack & logs
+    for i in range(int(opcode) - Opcode.LOG0):
```
icemelon, line 25: you should expand this part. The current for loop didn't reflect the logic in the circuit code

src/zkevm_specs/util/param.py
```diff
@@ -76,3 +76,9 @@
 SSTORE_SET_GAS = 20000
 SSTORE_RESET_GAS = 2900
 SSTORE_CLEARS_SCHEDULE = 15000
+
+# log static gas
+LOG_STATIC_GAS = 375
```
icemelon, line 81:
```suggestion
GAS_COST_LOG = 375
```

and move to above section, like line 46

src/zkevm_specs/util/gas.py
```diff
@@ -0,0 +1,43 @@
+from ..encoding import U8, U64, U128, U256
```
icemelon, line 1: I think we should rename this file to `testing.py` to reduce the confusion as it only provides utility functions for testing.

src/zkevm_specs/evm/instruction.py
```diff
@@ -359,6 +364,11 @@ def tx_context_lookup(
     def tx_calldata_lookup(self, tx_id: FQ, index: FQ) -> FQ:
         return self.tables.tx_lookup([tx_id, TxContextFieldTag.CallData, index])[3]
 
+    def tx_log_lookup(self, field_tag: TxLogFieldTag, index: int = 0) -> Union[int, RLC]:
+        return self.rw_lookup(RW.Write, RWTableTag.TxLog, [self.curr.log_index, index, field_tag])[
```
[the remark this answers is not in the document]
DreamWuGit: the `index` also stands for log `Data` order, so I don't change it to either `topic_index` or `data_index`

src/zkevm_specs/util/gas.py
DreamWuGit: make sense , will update !

DreamWuGit: updated in [0912af8](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/0912af8352fad807c9c70e00c0cd89c1e7532ded)

src/zkevm_specs/evm/table.py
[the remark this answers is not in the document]
DreamWuGit: removed in [0912af8](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/0912af8352fad807c9c70e00c0cd89c1e7532ded)

src/zkevm_specs/util/gas.py
DreamWuGit: updated in [0912af8](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/0912af8352fad807c9c70e00c0cd89c1e7532ded)

src/zkevm_specs/evm/table.py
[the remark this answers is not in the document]
DreamWuGit: removed in [0912af8](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/0912af8352fad807c9c70e00c0cd89c1e7532ded)

src/zkevm_specs/evm/table.py
DreamWuGit: removed in [0912af8](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/0912af8352fad807c9c70e00c0cd89c1e7532ded)

DreamWuGit:
> A part from some notes I added, I would like to discuss the possibility of revert: If a log is generated during a call that reverts, will the log be reverted? (that is, removed from the logs of the block). If so, what's the strategy to handle this case? Two ideas come to my mind:
> 
> 1. If `not is_persistent`, all log operations are reverted.  This would be very inconvenient because we need to "play back" all the `CopyToLog` steps, and backtrack the `log_index`
> 2. If `not is_persistent`, just don't generate the logs operations.  I think that the `LOG*` opcodes don't have any side effects observable within the execution context in the current block right?  So that should be doable.
> 
> Please let me know what you think.

@ed255  thanks for you mention reverting,  logs will be also reverted,  but something different with other kind of storage (i.e. account storage, which can go back to pre or default value if reverting happens), . logs don't have pre or default value ,  just were discarded (at this point, similar to Stack ,Memory type).  my thinking is similar to your 2nd idea,  logs will not affect execution context, but only result in part of receipts root of block in successful( non revert) condition .  this is the reason why current implementation don't explicit handle reverting case.   Glad to know if you have any concern !

src/zkevm_specs/evm/execution/log.py
DreamWuGit: updated in [69f93be](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/69f93be980b266701a44d3070e0fc8000dceadb5) thanks

src/zkevm_specs/util/gas.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Sequence, Tuple
+
+G_MEM = 3
+
+
+def memory_size(
```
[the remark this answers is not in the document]
DreamWuGit: updated in [69f93be](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/69f93be980b266701a44d3070e0fc8000dceadb5) thanks

src/zkevm_specs/evm/table.py
```diff
@@ -212,6 +218,27 @@ class CallContextFieldTag(IntEnum):
     StateWriteCounter = auto()
 
 
+class TxLogFieldTag(IntEnum):
```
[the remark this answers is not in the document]
ed255: @DreamWuGit I want to share some alternative ideas for the usage of the rw_table.  This is different from your design but maybe you find something interesting.  Feel free to use some of the ideas or none at all!


| 0 *rwc*  | 1 *isWrite* | 2 *Key0 (Tag)* | 3 *Key1* | 4 *Key2* | 5 *Key3*    | 6 *Key4*    | 7 *Value0* | 8 *Value1* | 9 *Aux0* | 10 *Aux1*  |
| -------- | ----------- | -------------- | -------- | -------- | ----------- | ----------- | ---------  | ---------- | -------- | ---------- |
|          |             |                |          |          |             |             |            |            |          |            |
|          |             |                |          |          | *TxLogTag*  |             |            |            |          |            |
| $counter | true        | TxLog          | $logID   |          | Address     |             | $value     |            |          |            |
| $counter | true        | TxLog          | $logID   |          | TxID        |             | $value     |            |          |            |
| $counter | true        | TxLog          | $logID   |          | Topic0      |             | $value     |            |          |            |
| $counter | true        | TxLog          | $logID   |          | Topic1      |             | $value     |            |          |            |
| $counter | true        | TxLog          | $logID   |          | Topic2      |             | $value     |            |          |            |
| $counter | true        | TxLog          | $logID   |          | Topic3      |             | $value     |            |          |            |
|          |             |                |          |          |             |             |            |            |          |            |
| $counter | true        | TxLogData      | $logID   |          | $byteIndex  |             | $value     |            |          |            |

Here are the changes:
1. Added TxID tag
2. Instead of specifying the topic number using an aux column, use 4 different tags for each topic
3. Add a new RwTag for the content of the log, this way it has the same shape as the Memory rows.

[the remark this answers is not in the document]
DreamWuGit: :)   thanks,  for topics ,  the topic count  is not fixed, from 0 to 4 (from `log0` to `log4`),  topics are reading from stack,  using fixed 4 will cause different behavior with stack operations.   I see your `logID ` as a synonym for `log_index`.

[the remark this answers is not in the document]
ed255: > :) thanks, for topics , the topic count is not fixed, from 0 to 4 (from `log0` to `log4`), topics are reading from stack, using fixed 4 will cause different behavior with stack operations.

In the proposed table I was thinking that only the required Topic number would be used.  Like this:
```python 
    for i in range(4):
        if i < topic_count:
            topic_is_zeros[i] = 0
            topic = instruction.stack_pop()
            instruction.constrain_equal(topic, instruction.tx_log_lookup(TxLogFieldTag.Topic0 + i))
```
But maybe that complicates other things.

> I see your `logID ` as a synonym for `log_index`.  

Yes, exactly! I used `logID` in the example as an equivalent, because we're using `*ID` everywhere (like TxID, CallID, etc.)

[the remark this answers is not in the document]
DreamWuGit: > This tag is not defined yet in https://github.com/appliedzkp/zkevm-specs/blob/master/specs/tables.md.
> 
> I would appreciate this PR also modifies this file to reflect the existance of this Tag and it's usage.

updated in [cb6f40b](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/cb6f40ba0eaf2db317a57a885739aeedc8e9985e) thanks!

src/zkevm_specs/evm/step.py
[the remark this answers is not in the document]
CPerezz: Well, maybe it's crazy but we could track it as an Operation. As we do with Stack/Memory/Storage read/write ops and add a section for it in the `rw_table`. 

On that way you should be able to access from both sites as the tables are the same.

WDYT?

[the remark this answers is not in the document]
DreamWuGit: the original concern here is removing `log_index` from step, instead solely rely on state circuit to constrain validity by ed255,  I  would agree almost all what you said above :）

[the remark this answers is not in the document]
icemelon:
```suggestion
    # log index of current block, this field maybe moved if we find them
    # not often used.
    log_id: FQ
```

specs/tables.md
```diff
@@ -96,6 +101,10 @@ Type sizes:
...
 | $counter | $isWrite    | Memory                     | $callID  |          | $memoryAddress             |             | $value     | 0          | 0        | 0               |
 | $counter | $isWrite    | AccountStorage             |          | $address |                            | $storageKey | $value     | $valuePrev | $txID    | $CommittedValue |
+|          |             |                            |          |          |                            |             |            |            |          |                 |
+| $counter | true        | TxLog                     |$log_index | 0        |  $address                  |             | $value     | 0          | 0        | 0               |
+| $counter | true        | TxLog                     |$log_index | $index   |  $Topics                   |             | $value     | 0          | 0        | 0               |
+| $counter | true        | TxLog                     |$log_index | $index   |  $Data                     |             | $value     | 0          | 0        | 0               |
```
icemelon, line 107:
```suggestion
| $counter | true        | TxLog                     |$logID | 0        |  Address                  |             | $value     | 0          | 0        | 0               |
| $counter | true        | TxLog                     |$logID | $topicIndex   |  Topic                   |             | $value     | 0          | 0        | 0               |
| $counter | true        | TxLog                     |$logID | $byteIndex   |  Data                     |             | $value     | 0          | 0        | 0               |
```

specs/tables.md
```diff
@@ -45,9 +45,14 @@ Type sizes:
...
 > - **storageKey**: field size, RLC encoded (Random Linear Combination)
-> - **value, valuePrev**: variable size, depending on Key 0 (Tag) and Key 3 where appropiate.
-> - **Key2** is reserved for Ethereum Address key type
-> - **Key4** is reserved for RLC encoded key type
+
+- **value, valuePrev**: variable size, depending on Key 0 (Tag) and Key 3 where appropiate.
+- **Key2** is reserved for Ethereum Address key type
+- **Key4** is reserved for RLC encoded key type
+- **log_index**, 8 bytes
+- **topics**, field size, RLC encoded (Random Linear Combination)
```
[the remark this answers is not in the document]
icemelon: need to update here according to the comment below

specs/tables.md
icemelon: and also need to add an additional row for `TopicSize` indicating the number of topics for this log

src/zkevm_specs/util/param.py
```diff
@@ -43,6 +43,8 @@
 GAS_COST_COPY = 3
 # Gas cost of non-creation transaction
 GAS_COST_TX = 21000
+# log static gas
```
icemelon, line 46:
```suggestion
# Constant gas cost of LOG
```

src/zkevm_specs/evm/step.py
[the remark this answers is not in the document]
DreamWuGit: open an issue #141 to track state circuit part of work related to `Log` section!

specs/tables.md
DreamWuGit: updated to use suggestion change for renaming.   For  `TopicSize` at the beginning I add it , but I don't see its usage in circuit as circuit can handle dynamic topic count, so I remove it later,  let me know if you still think it is needed !

icemelon: As per offline discussion, I agree that `TopicSize` is not necessary at this moment.

DreamWuGit:
@ed255 @CPerezz per discussion in  issue https://github.com/appliedzkp/zkevm-specs/issues/141, also chatted with Han,  it make sense to make those change  in incoming state circuit PR,  Want to know if you have other concerns here,  if no, we can consider to merge this at the moment .

CPerezz:
@DreamWuGit fine for me. Make sure an issue is filled to track this tech-debt 😁

specs/tables.md
```diff
@@ -45,9 +45,14 @@ Type sizes:
...
 > - **Memory -> value, valuePrev**: 1 byte
 > - **storageKey**: field size, RLC encoded (Random Linear Combination)
-> - **value, valuePrev**: variable size, depending on Key 0 (Tag) and Key 3 where appropiate.
-> - **Key2** is reserved for Ethereum Address key type
-> - **Key4** is reserved for RLC encoded key type
+
+- **value, valuePrev**: variable size, depending on Key 0 (Tag) and Key 3 where appropiate.
+- **Key2** is reserved for Ethereum Address key type
+- **Key4** is reserved for RLC encoded key type
+- **logID**, 8 bytes
```
ed255, line 52: Could you please specify if logID is unique per block?  And maybe add a note that this is not yet stable (as I think in the end logID will be unique per Tx, so that we know the log position within the receipt).

tests/evm/test_logs.py
```diff
@@ -0,0 +1,317 @@
...
+
+    verify_steps(
+        randomness=randomness,
+        tables=tables,
+        steps=steps,
+    )
+
+
+# helper to construct topics rows of RW table
+def construct_topic_rws(gc: U256, sp: int, topics: list, randomness: int):
```
ed255, line 295: I think you mean `rw_counter` instead of `gc` right?

src/zkevm_specs/evm/execution/log.py
```diff
@@ -0,0 +1,67 @@
...
+    mstart = instruction.rlc_to_fq_exact(instruction.stack_pop(), 8)
+    msize = instruction.rlc_to_fq_exact(instruction.stack_pop(), 8)
+
+    # check contract_address in CallContext & TxLog
+    # use call context's  callee address as contract address
+    contract_address = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
+    instruction.constrain_equal(contract_address, instruction.tx_log_lookup(TxLogFieldTag.Address))
+    # check not static call
+    instruction.constrain_equal(0, instruction.call_context_lookup(CallContextFieldTag.IsStatic))
+
```
ed255, line 24: The only thing I find missing in this implementation is a read to CallContext -> IsPersistent, and then using that as a dynamic selector to disable (when is_persistent = 0):
1. All the lookups to TxLog rows in rw_table (and their associated increase in rw_counter)
2. The constraint that the next step must be `ExecutionState.CopyToLog`

ed255 (changes requested):
I've added a few notes.  Most of them are small.  The only big one is the one related to handling of the opcode when `is_persistent = 0`.

After these points are addressed I agree that we can move on and address the update to the state circuit (and rw table) in a future PR :)

specs/tables.md
DreamWuGit: yes, logID will be unique per Tx.

specs/tables.md
```diff
@@ -96,6 +101,10 @@ Type sizes:
...
 | $counter | $isWrite    | Memory                     | $callID  |          | $memoryAddress             |             | $value     | 0          | 0        | 0               |
 | $counter | $isWrite    | AccountStorage             |          | $address |                            | $storageKey | $value     | $valuePrev | $txID    | $CommittedValue |
+|          |             |                            |          |          |                            |             |            |            |          |                 |
+| $counter | true        | TxLog                     |$logID | 0        |  Address                  |             | $value     | 0          | 0        | 0               |
+| $counter | true        | TxLog                     |$logID | $topicIndex   |  Topic                   |             | $value     | 0          | 0        | 0               |
+| $counter | true        | TxLog                     |$logID | $byteIndex   |  Data                     |             | $value     | 0          | 0        | 0               |
```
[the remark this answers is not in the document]
DreamWuGit: yeah, will do it in PR #159

src/zkevm_specs/evm/execution/log.py
DreamWuGit: added in [a9b4e08](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/a9b4e08f53226f01a0310de80b2366a6794129e9)

src/zkevm_specs/evm/table.py
```diff
@@ -212,6 +218,20 @@ class CallContextFieldTag(IntEnum):
...
+    Tag for RWTable lookup with tag TxLog, which is used to index specific
+    field of TxLog.
+    """
+
+    # The following are write-only data inside a transaction, they will be written in
+    # State circuit directly.
+    Address = auto()  # address of the contract that generated the event
+    Topic = auto()  # list of topics provided by the contract
+    Data = auto()  # log data in bytes
+    BlockNumber = auto()  # block number which logs are belong to
```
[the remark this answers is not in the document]
DreamWuGit: removed

tests/evm/test_logs.py
DreamWuGit: yes, changed to rw_counter in [a9b4e08](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/a9b4e08f53226f01a0310de80b2366a6794129e9)

src/zkevm_specs/evm/execution/log.py
ed255: Thanks!  I believe there's a more optimized way to handle the `is_persistent = False`: also skip the Memory reads as well as skipping all the `CopyToLog` steps.  Nevertheless I think the current implementation is correct so I'm OK with it!

src/zkevm_specs/evm/execution/log.py
```diff
@@ -0,0 +1,76 @@
...
+    instruction.step_state_transition_in_same_context(
+        opcode,
+        rw_counter=Transition.delta(instruction.rw_counter_offset),
+        program_counter=Transition.delta(1),
+        stack_pointer=Transition.delta(2 + opcode - Opcode.LOG0),
+        state_write_counter=Transition.delta(1),
+        dynamic_gas_cost=dynamic_gas,
+        memory_size=Transition.to(next_memory_size),
+        log_id=Transition.delta(1),
```
ed255, line 75: One last thing: log_id should only be `delta(1)` when `is_persistent = 1`.
Maybe you can do `log_id=Transition.delta(is_persistent)`

ed255 (commented):
I've added one last comment for the transition of log_id.  After that is addressed, everything will look good to me :)

tests/evm/test_logs.py
```diff
@@ -0,0 +1,317 @@
...
+            code_source=bytecode_hash,
+            program_counter=0,
+            stack_pointer=1015,
+            memory_size=mstart,
+            gas_left=dynamic_gas,
+            log_id=0,
+            state_write_counter=0,
+        )
+    ]
+    rws = [
```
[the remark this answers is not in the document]
ed255: @DreamWuGit requested to address this in https://github.com/appliedzkp/zkevm-specs/pull/159

src/zkevm_specs/evm/execution/log.py
DreamWuGit: Ah,  right , good catch!  will update it .

DreamWuGit: updated in [c2f1e79](https://github.com/appliedzkp/zkevm-specs/pull/125/commits/c2f1e79faa999e32bd3c28a799c4ca69fb892815) as well as test!

ed255 (approved):
LGTM! Great work!

ed255:
@icemelon do you consider the changes you requested i this PR resolved? (currently your review status is "requested changes")

icemelon (approved):
LGTM

PR #131, opened 2022-02-24
use constraints for `sload` & `sstore` gas

HAOYUatHZ:
close https://github.com/appliedzkp/zkevm-specs/issues/123

### DIFF
--- a/src/zkevm_specs/evm/execution/storage.py
+++ b/src/zkevm_specs/evm/execution/storage.py
@@ -114,16 +114,13 @@ def sstore(instruction: Instruction):
     instruction.constrain_equal(gas_refund, gas_refund_new)
 
     warm_case_gas = instruction.select(
-        instruction.is_equal(value_prev, value),
+        instruction.is_equal(value_prev, value)
+        or (not instruction.is_equal(original_value, value_prev)),
         SLOAD_GAS,
         instruction.select(
-            instruction.is_equal(original_value, value_prev),
-            instruction.select(
-                instruction.is_zero(original_value),
-                SSTORE_SET_GAS,
-                SSTORE_RESET_GAS,
-            ),
-            SLOAD_GAS,
+            instruction.is_zero(original_value),
+            SSTORE_SET_GAS,
+            SSTORE_RESET_GAS,
         ),
     )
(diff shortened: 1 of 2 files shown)


src/zkevm_specs/evm/execution/storage.py
```diff
@@ -70,36 +70,63 @@ def sstore(instruction: Instruction):
...
+        instruction.select(
+            instruction.is_equal(original_value, value_prev),
+            instruction.select(
+                instruction.is_zero(original_value),
+                SSTORE_SET_GAS,
+                SSTORE_RESET_GAS,
+            ),
+            SLOAD_GAS,
+        ),
+    )
```
HAOYUatHZ, line 128: in fact, this part can be simplified as 
```
warm_case_gas = instruction.select(
    instruction.is_equal(value_prev, value) | (~ instruction.is_equal(original_value, value_prev)),
    SLOAD_GAS,
    instruction.select(
        instruction.is_zero(original_value),
        SSTORE_SET_GAS,
        SSTORE_RESET_GAS,
    ),
)
```

but the simplified version doesn't reflect the logic in https://eips.ethereum.org/EIPS/eip-2200 directly, so I don't know which version I should keep.

Would like to apply it if we think the simplified version is better

ChihChengLiang: Does the ` | ` in the simplified version work in the circuit?

ChihChengLiang:
@HAOYUatHZ Can we do a rebase before the review?

src/zkevm_specs/evm/execution/storage.py
HAOYUatHZ: The logic is the same. It basically merges these two parts
1) L117~118 
2) L121~126 
together

HAOYUatHZ:
> @HAOYUatHZ Can we do a rebase before the review?

sure.

src/zkevm_specs/evm/execution/storage.py
ChihChengLiang: I see. The simplified version looks better to me. Will hold the review after you add it.

ChihChengLiang (approved):
LGTM

CPerezz (approved):
LGTM

PR #137, opened 2022-03-02
Don't allow key0 to be 0

z2trillion:
It being 0 will cause us to hit the unreachable code in line 363 because 0 isn't one of the allowed values for `Tag`.


HAOYUatHZ (approved):

ChihChengLiang (approved):
LGTM

PR #132, opened 2022-02-25
Add `mypy`

han0110:
This PR aims to add mypy for more strict type checking.

Following @ed255's suggestion in [this comment](https://github.com/appliedzkp/zkevm-specs/pull/116#issuecomment-1046723302), it introduces an interface `Expression` providing function `expr` which returns a `FQ`. In this way, we are always comparing number in field, and can avoid unsupported types comparing or integer underflow issues.

It's built upon #110 because #110 removes some unused things, then we don't bother to fix typing of them.

Co-authored-by: @ChihChengLiang
Co-authored-by: @ed255

PR #139, opened 2022-03-08
Specs for EXTCODEHASH

z2trillion:


src/zkevm_specs/evm/execution/extcodehash.py
```diff
@@ -0,0 +1,33 @@
...
+from ...util import keccak256
+
+
+def extcodehash(instruction: Instruction):
+    opcode = instruction.opcode_lookup(True)
+
+    address = instruction.stack_pop()
+
+    call_id = instruction.call_context_lookup(CallContextFieldTag.TxId)
+    is_warm = instruction.add_account_to_access_list(call_id, address)
```
ed255, line 14:
```suggestion
    tx_id = instruction.call_context_lookup(CallContextFieldTag.TxId)
    is_warm = instruction.add_account_to_access_list(tx_id, address)
```
Typo maybe?

ed255 (changes requested):
I've just added a minor comment for a typo.  But a part from that, everything looks great!

ed255 (approved):
LGTM! Great work!

src/zkevm_specs/evm/execution/extcodehash.py
```diff
@@ -0,0 +1,33 @@
...
+    address = instruction.stack_pop()
+
+    tx_id = instruction.call_context_lookup(CallContextFieldTag.TxId)
+    is_warm = instruction.add_account_to_access_list(tx_id, address)
+
+    nonce = instruction.account_read(address, AccountFieldTag.Nonce)
+    balance = instruction.account_read(address, AccountFieldTag.Balance)
+    code_hash = instruction.account_read(address, AccountFieldTag.CodeHash)
+
+    is_empty = nonce == 0 and balance == 0 and code_hash == int.from_bytes(keccak256(""), "big")
```
icemelon, line 20: you should use `instruction.is_zero` to check if a value is 0, and multiply all three conditions together, like
`instruction.is_zero(nonce) * instruction.is_zero(balance) * instruction.is_zero(code_hash - hash)`

src/zkevm_specs/evm/execution/extcodehash.py
```diff
@@ -0,0 +1,33 @@
...
+
+    instruction.constrain_equal(
+        0 if is_empty else code_hash,
```
icemelon, line 23: use `instruction.select` here.

src/zkevm_specs/util/param.py
```diff
@@ -71,6 +71,7 @@
 
 
 COLD_SLOAD_COST = 2100
+COLD_ACCOUNT_ACCESS_COST = 100
```
icemelon, line 74: merge this part with the gas cost section in line 18-63, also add docs for each constant

src/zkevm_specs/evm/execution/extcodehash.py
```diff
@@ -0,0 +1,38 @@
...
+    code_hash = instruction.account_read(address, AccountFieldTag.CodeHash)
+
+    is_empty = (
+        instruction.is_zero(nonce)
+        * instruction.is_zero(balance)
+        * instruction.is_zero(code_hash - EMPTY_CODE_HASH)
+    )
+
+    instruction.constrain_equal(
+        instruction.select(not instruction.is_zero(is_empty), 0, code_hash),
```
icemelon, line 28:
```suggestion
        instruction.select(is_empty, 0, code_hash),
```
We don't need to add `is_zero` for `is_empty` because the way `is_empty` is calculated makes it be only 0 or 1 value.

icemelon (commented):
also need to rebase to latest master again

src/zkevm_specs/evm/execution/extcodehash.py
```diff
@@ -0,0 +1,38 @@
...
+        instruction.stack_push(),
+    )
+
+    instruction.step_state_transition_in_same_context(
+        opcode,
+        rw_counter=Transition.delta(7),
+        program_counter=Transition.delta(1),
+        stack_pointer=Transition.delta(0),
+        dynamic_gas_cost=is_cold * EXTRA_GAS_COST_ACCOUNT_COLD_ACCESS,
```
han0110, line 37: Sorry that after looking into circuit, I realized `is_warm` is more straightforward in real circuit (witnessing the previous value directly), so I revert the change made in #132 by #145. Then this line might also need to be updated.

```suggestion
        dynamic_gas_cost=(1 - is_warm) * EXTRA_GAS_COST_ACCOUNT_COLD_ACCESS,
```

icemelon (approved):
LGTM

PR #145, opened 2022-03-11
Use `is_warm` instead of `is_cold` for access list value

han0110:
Just realized using `is_warm` as value of access list in real circuit seems more straightforward, so this PR aims to revert the change made in #132, and also update the usage of access list in `BeginTx`.

(to be nitpicking, it should be `is_already_warm` or `is_warm_prev` since it's the previous value, but omitting `already` or `prev` seems fine here)


ed255 (approved):
LGTM!

HAOYUatHZ (approved):

Issue #102, opened 2022-01-22
Retrieve code `length`


han0110:
Currently we can use lookup with input `(hash, index, byte, is_code)` to `bytecode_table` to retrieve the `byte` at specific `index`. However, we haven't taken the `length` of code into too much consideration, which is required for several places:

1. `CODESIZE`, `EXTCODESIZE`
2. `PUHS*`, `CODECOPY`, `EXTCODECOPY` right padding

    When copying out of code bound, implicit `STOP`s will be copied. It could be seen as right padding of the source. <br>See _yellowpaper[^yellowpaper] PUSH* and *CODECOPY_.

[^yellowpaper]: https://ethereum.github.io/yellowpaper/paper.pdf

3. Execution implicit `STOP`

    When execution has program counter out of code bound, an implicit `STOP` will be returned. See _yellowpaper (148)_.

Since the `length` already exists in Bytecode circuit, we can just also populate it into `bytecode_table` to be a 5-columns table <code>(hash, <b>length</b>, index, byte, is_code)</code>.

Then there is a caveat: If the code is empty, it's not allowed to be in current Bytecode circuit. So when we want to retrieve code `length` by `bytecode_table` lookup, we need to first check if `hash != EMPTY_HASH`.

Any feedback is appreciated.

ed255:
I think the right padding behavior also happens when accessing call data with `CALLDATALOAD` and `CALLDATACOPY`
- CALLDATALOAD: https://github.com/ethereum/go-ethereum/blob/015fde9a2c/core/vm/instructions.go#L286
- CALLDATACOPY: https://github.com/ethereum/go-ethereum/blob/015fde9a2c/core/vm/instructions.go#L312
- getData definition: https://github.com/ethereum/go-ethereum/blob/015fde9a2c/core/vm/common.go#L54

In that case we do have `CallContext` -> `CallDataLength` in rw_table, but what's the strategy going to be?  This is how I imagined it: 2 lookups: 1st for length, and 2nd for value.  If query_length > lookup_length, then the 2nd value "output" will be shortcut with 0, and "ignore" the second lookup (maybe by looking up a known constant row).

So what about doing the same approach with bytecode by adding a tag?
- (hash, TagLength, length, 0, 0)
- (hash, TagByte, index, byte)
And then we always insert a row with (EMPTY_HASH, TagLength, 0, 0, 0)

Does this make sense?

han0110:
[...]

Issue #73, opened 2021-12-04
Executed bytecode source for opcode lookup


han0110:
In EVM there are 3 different sources for executed bytecode:

1. When contract interaction, it's from **contract's bytecode**
2. When contract creation in root call, it's from **transaction's calldata**
3. When contract creation in internal call, it's from **caller's memory**

To avoid too much complexity on EVM circuit, we try to remove 3. by explicitly copying the caller's memory to `bytecode_table`, since for `CREATE2` we already need to hash the `initcode`, so we abuse this for also `CREATE`.

Then the sources now are either 1. or 2., but there is another thing EVM circuit desires for:

For executed bytecode, EVM circuit expect executed bytecode to be analyzed and annotated if it's `is_code` on every byte, for verifying `JUMP*` in a single-step way, otherwise every time when it comes to `JUMP*`, it needs to look back to check if it's `is_code` or not, and this could be multi-steps.

So there are at least 2 options here to support creation transaction:

1. EVM circuit lookup sources 1. and 2. by condition, and implement `is_code` annotation on both bytecode circuit and tx circuit.

    This seems reasonable but tx circuit has not been explored too much yet. Although it brings some implementation redundancy, the `is_code` annotation should not be too costly to do.

2. EVM circuit lookup sources 1. always, and explicitly copying transaction's calldata to `bytecode_table` when it's a creation one.

    This seems to bring more loading to `bytecode_table` at first glance, but if a DOS attacker wants to blow up the `bytecode_table`, it would rather use `EXTCODECOPY` to read different huge contracts, which only costs a constant `2600` and generates `0x6000/2600 ≈ 9.45` bytes per gas in worst case. So it seems also a reasonable approach to avoid implementation redundancy and make the executed bytecode single source for EVM circuit.

Any feedbacks are appreciated.

roynalnaruto:
I am currently addressing the review comments on https://github.com/appliedzkp/zkevm-specs/pull/191, and as @ed255 pointed out, it is important to finalise the definition of `StepState.code_source` in the specs.

@han0110 My opinion on the above issue is to use the bytecode table as the single source even for contract creation, specifically, this:
> EVM circuit lookup sources 1. always, and explicitly copying transaction's calldata to bytecode_table when it's a creation one.

This will also reduce implementation complexity and any chances of possibly missing the `is_create` cases for any `*CODE` related opcodes.

[...]

PR #148, opened 2022-03-16
CODECOPY opcode

roynalnaruto:
Circuits PR: https://github.com/appliedzkp/zkevm-circuits/pull/391

### DIFF
--- a/specs/opcode/39CODECOPY.md
+++ b/specs/opcode/39CODECOPY.md
@@ -20,7 +20,7 @@ The gadget then transits to the internal state of `CopyCodeToMemory`.
 
 1. opId = 0x39
 2. State Transitions:
-   - rw_counter -> rw_counter + 6 (3 stack reads, 1 call context read, 2 account reads)
+   - rw_counter -> rw_counter + 5 (3 stack reads, 1 call context read `CalleeAddress`, 1 account read `CodeHash`)
    - stack_pointer -> stack_pointer + 3
    - pc -> pc + 1
    - gas -> 3 + dynamic_cost (memory expansion and copier cost when `size > 0`)
@@ -32,8 +32,8 @@ The gadget then transits to the internal state of `CopyCodeToMemory`.
    - `code_offset` is at the second position of the stack
    - `size` is at the third position of the stack
    - `callee_address` is in the call context for the current call
-   - `code_size` is in the account context for the `callee_address`
    - `code_hash` is in the account context for the `callee_address`
+   - `code_size` from the bytecode table
 
 ## Exceptions
 
--- a/specs/opcode/39CODECOPY.md
+++ b/specs/opcode/39CODECOPY.md
@@ -20,7 +20,7 @@ The gadget then transits to the internal state of `CopyCodeToMemory`.
 
 1. opId = 0x39
 2. State Transitions:
-   - rw_counter -> rw_counter + 5 (3 stack reads, 1 call context read `CalleeAddress`, 1 account read `CodeHash`)
+   - rw_counter -> rw_counter + 3 (3 stack reads)
    - stack_pointer -> stack_pointer + 3
    - pc -> pc + 1
    - gas -> 3 + dynamic_cost (memory expansion and copier cost when `size > 0`)
@@ -31,8 +31,6 @@ The gadget then transits to the internal state of `CopyCodeToMemory`.
    - `memory_offset` is at the top of the stack
    - `code_offset` is at the second position of the stack
    - `size` is at the third position of the stack
-   - `callee_address` is in the call context for the current call
-   - `code_hash` is in the account context for the `callee_address`
    - `code_size` from the bytecode table
 
 ## Exceptions
@@ -42,4 +40,4 @@ The gadget then transits to the internal state of `CopyCodeToMemory`.
 
 ## Code
 
-Please refer to `src/zkevm_specs/evm/execution/codecode.py`
+Please refer to `src/zkevm_specs/evm/execution/codecopy.py`
--- a/specs/opcode/39CODECOPY.md
+++ b/specs/opcode/39CODECOPY.md
@@ -1,0 +1,45 @@
+# CODECOPY opcode
+
+## Procedure
+
+The `CODECOPY` opcode pops `memory_offset`, `code_offset` and `size` from the stack.
+It then copies `size` bytes of code running in the current environment from an offset `code_offset` to the memory at the address `memory_offset`. For out-of-bound scenarios where `size > len(code) - code_offset`, EVM pads 0 to the end of the copied bytes.
+
+The gas cost of `CODECOPY` opcode consists of two parts:
+
+1. A constant gas cost: `3 gas`
+2. A dynamic gas cost: cost of memory expansion and copying (variable depending on the `size` copied to memory)
+
+## Circuit Behaviour
+
+`CODECOPY` makes use of the internal execution step `CopyCodeToMemory` and loops over these steps iteratively until there are no more bytes to be copied. The `CODECOPY` circuit itself only constrains the values popped from stack and call context/account read lookups.
+
+The gadget then transits to the internal state of `CopyCodeToMemory`.
+
+## Constraints
+
+1. opId = 0x39
+2. State Transitions:
+   - rw_counter -> rw_counter + 6 (3 stack reads, 1 call context read, 2 account reads)
+   - stack_pointer -> stack_pointer + 3
+   - pc -> pc + 1
+   - gas -> 3 + dynamic_cost (memory expansion and copier cost when `size > 0`)
+   - memory_size
+     - `prev_memory_size` if `size = 0`
+     - `max(prev_memory_size, (memory_offset + size + 31) / 32)` if `size > 0`
+3. Lookups:
+   - `memory_offset` is at the top of the stack
+   - `code_offset` is at the second position of the stack
+   - `size` is at the third position of the stack
+   - `callee_address` is in the call context for the current call
+   - `code_size` is in the account context for the `callee_address`
+   - `code_hash` is in the account context for the `callee_address`
+
+## Exceptions
+
+1. Stack Underflow: `1021 <= stack_pointer <= 1024`
+2. Out-of-Gas: remaining gas is not enough
+
+## Code
+
+Please refer to `src/zkevm_specs/evm/execution/codecode.py`
--- a/src/zkevm_specs/bytecode.py
+++ b/src/zkevm_specs/bytecode.py
@@ -1,13 +1,13 @@
 from typing import Sequence, Union, Tuple, Set, NamedTuple
 from collections import namedtuple
-from .util import keccak256, FQ, RLC
-from .evm import get_push_size, BytecodeTableRow
+from .util import keccak256, EMPTY_HASH, FQ, RLC
+from .evm import get_push_size, BytecodeFieldTag, BytecodeTableRow
 from .encoding import U8, U256, is_circuit_code
 
 # Row in the circuit
 Row = namedtuple(
     "Row",
-    "q_first q_last hash index byte is_code push_data_left hash_rlc hash_length byte_push_size is_final padding",
+    "q_first q_last hash tag index is_code value push_data_left hash_rlc hash_length byte_push_size is_final padding",
 )
 # Unrolled bytecode
 class UnrolledBytecode(NamedTuple):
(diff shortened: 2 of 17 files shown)


CPerezz:
Hey @roynalnaruto! I saw the PR for the circuits already up and ready for review.
Could we set it in Draft and focus on merging this one first? As until we don't have a merged spec we can't approve a circuit PR directly.

roynalnaruto:
> Hey @roynalnaruto! I saw the PR for the circuits already up and ready for review. Could we set it in Draft and focus on merging this one first? As until we don't have a merged spec we can't approve a circuit PR directly.

@CPerezz I just had to push an unchanged change and rebase with upstream. I've now opened this PR.

han0110:
Hi @roynalnaruto, thanks for working on this!

Have you considered about retrieving the `CodeSize` information from `bytecode_table`? Bytecode circuit already has it, but just currently not designed to be lookup-able (see #102 for more discussion).

Also state trie doesn't contain the `CodeSize` of account, so we can't rely on MPT circuit to open its value, then we still need to check it with the one in Bytecode circuit every time when we access an account, so I think it doesn't makes too much sense to have such information maintained in State circuit.

src/zkevm_specs/evm/execution/codecopy.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+    instruction.step_state_transition_in_same_context(
+        opcode,
+        rw_counter=Transition.delta(instruction.rw_counter_offset),
+        program_counter=Transition.delta(1),
+        stack_pointer=Transition.delta(3),
+        memory_size=Transition.to(next_memory_size),
+        dynamic_gas_cost=gas_cost,
+    )
```
CPerezz, line 53: Add a new line at the end of the file please :)

src/zkevm_specs/evm/execution/codecopy.py
```diff
@@ -0,0 +1,53 @@
...
+    account = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
+    code_size = instruction.account_read(account, AccountFieldTag.CodeSize)
+    code_hash = instruction.account_read(account, AccountFieldTag.CodeHash)
+
+    next_memory_size, memory_expansion_gas_cost = instruction.memory_expansion_dynamic_length(
+        memory_offset, size
+    )
+    gas_cost = instruction.memory_copier_gas_cost(size, memory_expansion_gas_cost)
+
+    if not instruction.is_zero(size):
```
CPerezz, line 29: I'm missing a call to `CopyCodeToMemory` or `BufferedReader` gadget here. Otherwise, you're just constraining the offsets and lengths, but not actually copying anything no?

Maybe I'm missing something.

src/zkevm_specs/evm/execution/copy_code_to_memory.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ...util import FQ, N_BYTES_MEMORY_SIZE, RLC
+from ..execution_state import ExecutionState
+from ..instruction import Instruction, Transition
+from ..step import CopyCodeToMemoryAuxData
+from ..table import RW
+from ..util import BufferReaderGadget
+
+
+MAX_COPY_BYTES = 54
```
CPerezz, line 12: Please, if this constant is not already present in the repo, document it and import it from `util`

CPerezz (commented):
Some nits/questions.
Overall is looking good! :)

CPerezz:
@han0110 made an excellent point there!
Bytecode table should be a place that allows us to leverage info holded now by te State circuit.

Worth considering to refactor this in that regard.

src/zkevm_specs/evm/execution/codecopy.py
roynalnaruto: @CPerezz We check that if the "to be copied" `size` is non-zero, then the next execution state MUST BE the internal state `CopyCodeToMemory`. We add a few more constraints on the auxiliary data, but the rest is taken over by the `CopyCodeToMemory` gadget.

roynalnaruto:
@han0110 @CPerezz I will check more on #102 and refactor the `CodeSize` part. Thanks for the review :)

roynalnaruto:
@han0110 @CPerezz I've added a commit to extend the bytecode circuit (and related code/tests) that:
1. Adds tags `Length` and `Byte` as `BytecodeFieldTag`
2. Adds a row to the bytecode circuit (for every bytecode) for its length
3. `CODECOPY` makes use of `instruction.bytecode_length(code_hash)` to fetch the bytecode size

han0110 (commented):
Overall looks good! Although I thought you could just change the interface in `table.py` without needing to worry about updating `bytecode.py` itself.

Currently Bytecode circuit is modified to have access to previous row and next row, which seems hard to track all the branches (also brings some duplication I think), I have submitted another proposal #151 that I always want to try, if you are interesting perhaps you can have a look and implement it in another PR.

src/zkevm_specs/bytecode.py
```diff
@@ -33,44 +33,93 @@ def select(
...
-        # hash_rlc := hash_rlc_prev * r + byte
-        assert row.hash_rlc == prev_row.hash_rlc * r + row.byte
-
-        # padding needs to remain the same
-        assert row.padding == prev_row.padding
-        # hash needs to remain the same
-        assert row.hash == prev_row.hash
-        # hash_length needs to remain the same
-        assert row.hash_length == prev_row.hash_length
+        if row.tag == BytecodeFieldTag.Length:
```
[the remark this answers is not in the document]
roynalnaruto: Yes, I am refactoring the constraints to make them simpler and also removing some duplication cases occurred while comparing `row` to `prev_row` and `next_row`

src/zkevm_specs/bytecode.py
```diff
@@ -33,44 +33,93 @@ def select(
...
+                # if len(bytecode) > 0 then the following rows are the bytes
+                assert next_row.tag == BytecodeFieldTag.Byte
+                # the immediate following row is the byte at index == 0
+                assert next_row.index == 0
+                # equality for bytecode length
+                assert next_row.hash_length == row.value
+        else:
+            if prev_row.tag == BytecodeFieldTag.Length:
+                # index starts from 0
+                assert row.index == 0
```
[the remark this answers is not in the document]
roynalnaruto: Line 87 was adding a constraint for the `index` of row with tag Length, but I believe that's unnecessary. So I'll remove it.

roynalnaruto:
> I have submitted another proposal https://github.com/appliedzkp/zkevm-specs/issues/151 that I always want to try, if you are interesting perhaps you can have a look and implement it in another PR.

@han0110 Yes, I think the proposal makes a lot of sense. I would like to take that up and refactor the bytecode circuit. But I believe that should be done as a separate PR, once this one is approved and merged (just trying to avoid too many changes and potential conflicts for others working on this repo).

src/zkevm_specs/evm/execution/codecopy.py
```diff
@@ -0,0 +1,51 @@
...
+    account = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
+    code_hash = instruction.account_read(account, AccountFieldTag.CodeHash)
+    code_size = instruction.bytecode_length(code_hash)
+
+    next_memory_size, memory_expansion_gas_cost = instruction.memory_expansion_dynamic_length(
+        memory_offset, size
+    )
+    gas_cost = instruction.memory_copier_gas_cost(size, memory_expansion_gas_cost)
+
+    if not instruction.is_zero(size):
```
han0110, line 29: I missed this in `calldatacopy.py` in #132, please help me also fix that, thanks.
```suggestion
    if instruction.is_zero(size) == FQ(0):
```

src/zkevm_specs/evm/step.py
```diff
@@ -95,3 +96,25 @@ def __init__(
...
         self.from_tx = FQ(from_tx)
         self.tx_id = FQ(tx_id)
+
+
+class CopyCodeToMemoryAuxData:
+    src_addr: FQ
+    dst_addr: FQ
+    bytes_left: FQ
+    src_addr_end: FQ
+    code: Bytecode
```
han0110, line 106: `aux_data` is tracked as witness in circuit, so we should just have `code_hash` here.
```suggestion
    code_hash: RLC
```

src/zkevm_specs/evm/execution/codecopy.py
```diff
@@ -0,0 +1,51 @@
...
+            instruction.next.execution_state, ExecutionState.CopyCodeToMemory
+        )
+        next_aux = instruction.next.aux_data
+        assert isinstance(next_aux, CopyCodeToMemoryAuxData)
+        instruction.constrain_equal(next_aux.src_addr, code_offset)
+        instruction.constrain_equal(next_aux.dst_addr, memory_offset)
+        instruction.constrain_equal(next_aux.src_addr_end, code_size)
+        instruction.constrain_equal(next_aux.bytes_left, size)
+        instruction.constrain_equal(
+            FQ(next_aux.code.hash()), instruction.rlc_to_fq_exact(code_hash, n_bytes=32)
```
han0110, line 41:
```suggestion
            next_aux.code_hash, instruction.curr.code_source
```

src/zkevm_specs/evm/execution/copy_code_to_memory.py
```diff
@@ -0,0 +1,58 @@
...
+        assert next_aux is not None and isinstance(next_aux, CopyCodeToMemoryAuxData)
+        instruction.constrain_equal(next_aux.src_addr, aux.src_addr + copied_bytes)
+        instruction.constrain_equal(next_aux.dst_addr, aux.dst_addr + copied_bytes)
+        instruction.constrain_equal(next_aux.bytes_left + copied_bytes, aux.bytes_left)
+        instruction.constrain_equal(next_aux.src_addr_end, aux.src_addr_end)
+        instruction.constrain_equal(FQ(next_aux.code.hash()), FQ(aux.code.hash()))
+
+    instruction.constrain_step_state_transition(
+        rw_counter=Transition.delta(instruction.rw_counter_offset),
+    )
```
han0110, line 58: I think `memory_copy.py` also misses this, the `constrain_step_state_transition` is default to constraint nothing except explicitly specified. Will do a refactor to make `step_state_transition_in_same_context` for such opcode and gas unrelated case. (btw test also needs update, some step state is incorrect)

```suggestion
    instruction.constrain_step_state_transition(
        rw_counter=Transition.delta(instruction.rw_counter_offset),
        call_id=Transition.same(),
        is_root=Transition.same(),
        is_create=Transition.same(),
        code_source=Transition.same(),
        program_counter=Transition.same(),
        stack_pointer=Transition.same(),
        memory_size=Transition.same(),
        state_write_counter=Transition.same(),
    )
```

src/zkevm_specs/evm/table.py
```diff
@@ -165,6 +174,7 @@ class AccountFieldTag(IntEnum):
     Nonce = auto()
     Balance = auto()
     CodeHash = auto()
+    CodeSize = auto()
```
han0110, line 177:
```suggestion
```

han0110:
> Yes, I think the proposal makes a lot of sense. I would like to take that up and refactor the bytecode circuit. But I believe that should be done as a separate PR, once this one is approved and merged (just trying to avoid too many changes and potential conflicts for others working on this repo). Wdyt?

Sure, let's do it in another PR.

src/zkevm_specs/evm/execution/codecopy.py
roynalnaruto: You mean:
```suggestion
    if instruction.is_zero(size) != FQ(0):
```
right? :)

han0110: But we want this branch to run when `size != 0`, right?  When `size != 0`, we have `instruction.is_zero(size) == FQ(0)`. The original code `not instruction.is_zero(size)` will always be `False` because `FQ` doesn't implement `__bool__`.

src/zkevm_specs/evm/step.py
roynalnaruto: I used the entire bytecode in `aux_data` because I wanted a way to get the `is_code` field of the bytecode table row.

Is there another way I can achieve that? I am referring to this part:
https://github.com/scroll-tech/zkevm-specs/blob/feat/codecopy/src/zkevm_specs/evm/execution/copy_code_to_memory.py#L20-L30

han0110: Ah I see. You can make the `is_code` to be optional, then don't pass it should work. (note that `Instruciton` also needs to be updated to have `is_code: bool = None`)

https://github.com/appliedzkp/zkevm-specs/blob/96c10a2d188d913421cfcfe34c38f35d951d200b/src/zkevm_specs/evm/table.py#L364-L377

```diff
-         is_code: Expression,
+         is_code: Expression = None,
```

roynalnaruto: I initially thought that it isn't required to check `is_code`, since it is simply copying byte-by-byte to memory. So I did consider having `is_code` optional. Still wanted to check with you, so that's cool, I will update this part of the codebase :)

han0110: I see, currently in circuit we indeed need such extra auxiliary information to make lookup work. We can try to not include `is_code` column for such case in the future.

roynalnaruto: Yes, as I can see, the `opcode_lookup` in fact needs the `is_code` field, whereas something like `CODECOPY` that copies byte-by-byte does not really care about `is_code`. Correct me if I'm wrong.

src/zkevm_specs/evm/table.py
```diff
@@ -366,14 +365,15 @@ def bytecode_lookup(
...
+        is_code: Expression = None,
     ) -> BytecodeTableRow:
         query = {
             "bytecode_hash": bytecode_hash,
             "field_tag": field_tag,
             "index": index,
-            "is_code": is_code,
         }
+        if is_code is not None:
+            query["is_code"] = is_code
```
han0110, line 376: It should be okay to put optional key in query, line 422 will omit for us.

src/zkevm_specs/evm/execution/codecopy.py
```diff
@@ -0,0 +1,51 @@
...
+        instruction.stack_pop(),
+        instruction.stack_pop(),
+    )
+
+    memory_offset, size = instruction.memory_offset_and_length(memory_offset_word, size_word)
+    code_offset = instruction.rlc_to_fq_exact(code_offset_word, N_BYTES_MEMORY_ADDRESS)
+
+    account = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
+    code_hash = instruction.account_read(account, AccountFieldTag.CodeHash)
+    instruction.constrain_equal(code_hash, instruction.curr.code_source)
```
han0110, line 22: Just realized `CODECOPY` copies current executed code, including contract creation code. While contract creation, the account's code hash is not updated actually.

These lines should be removed, and we should use `code_source` as the `aux.code_hash`, it will always be the pointer to where the code is executed (as for why it's `code_source` instead of `code_hash`, see #73).
```suggestion
```

han0110 (approved):
All LGTM now. Great work!

roynalnaruto:
@CPerezz and @icemelon would be good to have your reviews on the PR :)

CPerezz (approved):
LGTM!! Nice job!!!

Only a small nit in the markdown file 🙂

specs/opcode/39CODECOPY.md
```diff
@@ -0,0 +1,45 @@
...
+   - `code_size` from the bytecode table
+
+## Exceptions
+
+1. Stack Underflow: `1021 <= stack_pointer <= 1024`
+2. Out-of-Gas: remaining gas is not enough
+
+## Code
+
+Please refer to `src/zkevm_specs/evm/execution/codecode.py`
```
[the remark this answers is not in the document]
roynalnaruto: I have been adding a new line, and I was kinda surprised those are being removed. Its `mdformat --number` that's removing the empty lines.

If I add an empty line at the end, `make lint` fails.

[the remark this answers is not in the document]
CPerezz: Oh wow...

Let's forget it then.😔

ChihChengLiang:
@roynalnaruto Can we rebase this and merge?

PR #152, opened 2022-03-21
Set MAX_KEY3 to MAX_MEMORY_ADDRESS

z2trillion:
2^32 is already a good upper bound. The maximum memory address that can be touched with the current gas limit of 30e6 is about 123,170, which sadly is just over 2^16.

### DIFF
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -7,12 +7,16 @@ from .encoding import U8, is_circuit_code
 from .evm import RW, AccountFieldTag, CallContextFieldTag
 
 MAX_KEY_DIFF = 2**32 - 1
+# The gas cost for memory operations is quadratic in the maximum memory address
+# touched. From equation 326 in the yellow paper, for C_mem(a), the maximum
+# memory address touched cannot exceed or equal 2^32 until the gas limit is
+# over 3.6e16.
 MAX_MEMORY_ADDRESS = 2**32 - 1
 MAX_STACK_PTR = 1023
 MAX_KEY0 = 10  # Number of Tag variants
 MAX_KEY1 = 2**16 - 1  # Maximum number of calls in a block
 MAX_KEY2 = 2**160 - 1  # Ethereum Address size
-MAX_KEY3 = 2**40 - 1  # Maximum value for Memory Address
+MAX_KEY3 = MAX_MEMORY_ADDRESS  # Maximum value for Memory Address
 KEY0_BITS = ceil(log(MAX_KEY0 + 1, 2))  # 4
 KEY1_BITS = ceil(log(MAX_KEY1 + 1, 2))  # 16
 KEY2_BITS = ceil(log(MAX_KEY2 + 1, 2))  # 160
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -20,7 +20,7 @@ MAX_KEY3 = MAX_MEMORY_ADDRESS  # Maximum value for Memory Address
 KEY0_BITS = ceil(log(MAX_KEY0 + 1, 2))  # 4
 KEY1_BITS = ceil(log(MAX_KEY1 + 1, 2))  # 16
 KEY2_BITS = ceil(log(MAX_KEY2 + 1, 2))  # 160
-KEY3_BITS = ceil(log(MAX_KEY3 + 1, 2))  # 40
+KEY3_BITS = ceil(log(MAX_KEY3 + 1, 2))  # 32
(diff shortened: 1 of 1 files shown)


z2trillion:
@ed255

src/zkevm_specs/state.py
```diff
@@ -7,12 +7,16 @@
...
+# touched. From equation 326 in the yellow paper, for C_mem(a), the maximum
+# memory address touched cannot exceed or equal 2^32 until the gas limit is
+# over 3.6e16.
 MAX_MEMORY_ADDRESS = 2**32 - 1
 MAX_STACK_PTR = 1023
 MAX_KEY0 = 10  # Number of Tag variants
 MAX_KEY1 = 2**16 - 1  # Maximum number of calls in a block
 MAX_KEY2 = 2**160 - 1  # Ethereum Address size
-MAX_KEY3 = 2**40 - 1  # Maximum value for Memory Address
+MAX_KEY3 = MAX_MEMORY_ADDRESS  # Maximum value for Memory Address
```
ed255, line 19: Could you please update line 23 so that it becomes this?
```python
KEY3_BITS = ceil(log(MAX_KEY3 + 1, 2)) # 32
```
Notice the updated commented value

ed255 (commented):
LGTM! I've added a minor note to update the commented value for `KEY3_BITS`

z2trillion:
Actually, is there a reason the memory address is in the key3 column instead of the key2 (account address) column? There's no account address that will be present when there's a memory address, so we can check that the memory address is fits into 140 bits for no marginal cost?

ed255 (approved):
LGTM!

ed255:
> Actually, is there a reason the memory address is in the key3 column instead of the key2 (account address) column? There's no account address that will be present when there's a memory address, so we can check that the memory address is fits into 140 bits for no marginal cost?

It's just to keep key2 exclusive for values of the Ethereum Address, as a matter of organization.  We could move the `memoryAddress` to key3 in Memory rows, but I don't see any particular advantage on doing that.  Do you think this would be an improvement?

han0110 (approved):
LGTM!

z2trillion:
> > s that will be present when there's a memory address, so we can check that the memory address is fits into 140 bits for no marginal cost?
> 
> It's just to keep key2 exclusive for values of the Ethereum Address, as a matter of organization. We could move the `memoryAddress` to key3 in Memory rows, but I don't see any particular advantage on doing that. Do you think this would be an improvement?

My understanding is that at the moment, key2 holds account addresses (or is 0 for rows with no account address). key3 is used for field tag values, stack addressses, and memory addresses. With u16 lookups, we can check that key2 fits into 160 bits using 10 advice columns, and that key3 fits into 32 bits using 2 advice columns.

If we move memory addresses to key2 (thus "allowing" memory addresses to be up to 160 bits), key3 will fit into 10 bits (max value of stack address), so we will no longer need any advice columns for the key3 range check. Incidentally, this also allows us to fit more key4 bytes into `k0` in `get_keys_compressed_in_order`, but since it won't eliminate the need for `k1`, I don't think this is a real improvement.

If we go furth put all addresses (account, memory, and stack) into the key2 column, then we can just think of it as the "address" column and key3 as the "field tag" column. What do you think @ed255?

ed255:
> My understanding is that at the moment, key2 holds account addresses (or is 0 for rows with no account address). key3 is used for field tag values, stack addressses, and memory addresses. With u16 lookups, we can check that key2 fits into 160 bits using 10 advice columns, and that key3 fits into 32 bits using 2 advice columns.
> 
> If we move memory addresses to key2 (thus "allowing" memory addresses to be up to 160 bits), key3 will fit into 10 bits (max value of stack address), so we will no longer need any advice columns for the key3 range check. Incidentally, this also allows us to fit more key4 bytes into `k0` in `get_keys_compressed_in_order`, but since it won't eliminate the need for `k1`, I don't think this is a real improvement.
> 
> If we go furth put all addresses (account, memory, and stack) into the key2 column, then we can just think of it as the "address" column and key3 as the "field tag" column. What do you think @ed255?

Thanks for the thorough explanation! It's now clear to me why it's interesting to move the `memoryAddress` (and also `stackAddress`) to key2 :)  The change sounds good to me!  Would you like to address it in this PR or in a future PR?

z2trillion:
I will do it in another PR.

ed255:
> I will do it in another PR.

I've opened an issue to track this :) https://github.com/appliedzkp/zkevm-specs/issues/160

I'm going to merge this PR now!

PR #153, opened 2022-03-22
remove mdformat

ChihChengLiang:
### Problem 

mdformat is not latex compatible and doesn't support ignore. We remove it to unblock #149


ed255 (approved):
LGTM!

PR #155, opened 2022-03-22
Add specs for blockCtx ops

bchyl:


src/zkevm_specs/evm/execution/block_ctx.py
```diff
@@ -0,0 +1,44 @@
...
+from ...util import FQ
+
+
+def blockctx(instruction: Instruction):
+    opcode = instruction.opcode_lookup(True)
+
+    res_op = instruction.curr.execution_state.responsible_opcode()[0]
+    if isinstance(res_op, int):
+        res_op_int = int(res_op)
+    instruction.constrain_equal(opcode, FQ(res_op_int))
```
ed255, line 14: Is this really needed? In `Instruction.step_state_transition_to_new_context` we're doing `self.responsible_opcode_lookup(opcode)` which I think checks the same thing.
The other opcodes are not doing this constraint check.

src/zkevm_specs/util/param.py
```diff
@@ -1,5 +1,5 @@
 # Maximun number of bytes with composition value that doesn't wrap around the field
-MAX_N_BYTES = 31
+MAX_N_BYTES = 32
```
ed255, line 2: There's no guarantee that 32 bytes don't wrap around the field, so this value should be kept at `31`.

src/zkevm_specs/evm/execution/block_ctx.py
```diff
@@ -0,0 +1,44 @@
+from ...util.param import N_BYTES_ACCOUNT_ADDRESS, N_BYTES_U64
+from ..instruction import Instruction, Transition
+from ..table import BlockContextFieldTag
+from ..opcode import Opcode
+from ...util import FQ
+
+
+def blockctx(instruction: Instruction):
```
ed255, line 8: I have one question about this function.  Currently in the circuits we have one gadget per execution state.  And some execution states handle multiple opcodes.
Will `blockctx` be a single gadget?
If it's a single gadget, I wonder if there's an advantage if we define a BlockContext execution state that handles all the block context opcodes in a single execution state :thinking: Maybe we would reduce the number of constraints a bit?

ed255 (changes requested):

ed255:
Thanks for submitting this PR!  I think merging the logic of all block context opcodes is a nice improvement :)

specs/opcode/41COINBASE_45GASLIMIT_48BASEFEE.md
```diff
@@ -0,0 +1,60 @@
+# block context op codes
```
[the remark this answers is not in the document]
bchyl: This is supposed to be on range from `41COINBASE` to `45GASLIMIT` and `48 BASEFEE` as trade-offs,  while the md file name as `41COINBASE_42TIMESTAMP_43NUMBER_44DIFFICULTY_45GASLIMIT_48BASEFEE.md` seems too long.
Maybe we can name it as `41COINBASE-45GASLIMIT_48BASEFEE.md`?

src/zkevm_specs/evm/execution/block_ctx.py
```diff
@@ -0,0 +1,44 @@
...
+    opcode = instruction.opcode_lookup(True)
+
+    res_op = instruction.curr.execution_state.responsible_opcode()[0]
+    if isinstance(res_op, int):
+        res_op_int = int(res_op)
```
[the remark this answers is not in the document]
bchyl: yeah!

src/zkevm_specs/evm/execution/block_ctx.py
bchyl: yeah, nice! when coding i also wonder for `responsible_opcode` and `responsible_opcode_lookup` have double checking for opcodeId.

src/zkevm_specs/util/param.py
bchyl: last version testing typo, sorry.

src/zkevm_specs/evm/execution/block_ctx.py
bchyl: yeah, in circuit code there is a  `blockctxGadget` for BlockContext ops.  `define a BlockContext execution state` is a good idea for reducing constraints. i will refactor it latter.

Thank you very much for your careful reviews.

specs/opcode/41COINBASE_45GASLIMIT_48BASEFEE.md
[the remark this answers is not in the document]
ed255: > Maybe we can name it as `41COINBASE-45GASLIMIT_48BASEFEE.md`?

I understand that `41COINBASE_42TIMESTAMP_43NUMBER_44DIFFICULTY_45GASLIMIT_48BASEFEE.md` is a very long filename, although it makes it explicit which opcodes are defined in the file.  I'm OK with `41COINBASE-45GASLIMIT_48BASEFEE.md` then!

Note that if we move on with this https://github.com/appliedzkp/zkevm-specs/issues/154 we'll be able to add an index of all the opcodes, so the filenames will be less important :)

ed255 (approved):
LGTM! Thanks for working on this :)

han0110 (approved):
LGTM! Perhaps we can just have `BlockContextField::{Coinbase,Timestamp,Number,GasLimit,Difficulty,BaseFee}` identical to `opcode`, then we can use `opcode` as `tag` of lookup input. Also if #135 is adopted, then `ctx_expr` could be identical for all opcodes.

barryWhiteHat:
@bchyl can you resolve conflicts here ? after that i think we can merge :)

PR #156, opened 2022-03-23
Fix `BeginTx` and `EndTx` step state transition

han0110:
This PR aims to fix the step state transition of `BeginTx` and `EndTx`, it includes:

- Enforce `BeginTx` to `EndTx` when callee with empty code
- Enforce `EndTx` to `EndBlock` has the same `call_id`

Also it removes the optional final state (by using a dummy step state), which is annoying for us to write `assert next_step is not None`


ed255 (approved):
LGTM!

CPerezz (approved):
LGTM! Can we update the PR and merge?

PR #157, opened 2022-03-23
Spec for opcode `ISZERO`

silathdiir:
circuit code: https://github.com/appliedzkp/zkevm-circuits/pull/399


src/zkevm_specs/evm/execution/iszero.py
```diff
@@ -0,0 +1,20 @@
...
+    instruction.constrain_equal(
+        instruction.is_zero(value),
+        instruction.stack_push(),
+    )
+
+    instruction.step_state_transition_in_same_context(
+        opcode,
+        rw_counter=Transition.delta(2),
+        program_counter=Transition.delta(1),
+        stack_pointer=Transition.delta(0),
```
DreamWuGit, line 19:
```suggestion
        stack_pointer=Transition.same(),
```

DreamWuGit (approved):
LGTM!

src/zkevm_specs/evm/execution/iszero.py
```diff
@@ -0,0 +1,20 @@
...
+from ..opcode import Opcode
+
+
+def iszero(instruction: Instruction):
+    opcode = instruction.opcode_lookup(True)
+
+    value = instruction.stack_pop()
+
+    instruction.constrain_equal(
+        instruction.is_zero(value),
```
icemelon, line 11: It's possible that an RLC encoded value for a non-zero value to be zero mod Fp, right? I think it's dangerous to directly use the RLC value to do this. It'd be safer to sum up all limbs and assert 0.

HAOYUatHZ: I think it makes sense, if encoded value is p, then this can be zero.

That means we need to fix
+ this PR
+ https://github.com/appliedzkp/zkevm-specs/issues/135 Observation 3
+ `sstore` specs and circuit
    + we use is_zero for RLC encoded value without summing up there
    + the reason why I did those was I was simply following what I used for `sload`: https://github.com/appliedzkp/zkevm-circuits/pull/334#discussion_r821535077. But I didn't realize that `sstore` case is different from `sload`'s.

HAOYUatHZ: > That means we need to fix
> 
> * specs PR: this PR
> * circuit PR: [opcode `ISZERO` zkevm-circuits#399](https://github.com/appliedzkp/zkevm-circuits/pull/399)
> * [Discussion about RLC requirements and alternatives #135](https://github.com/appliedzkp/zkevm-specs/issues/135) Observation 3
> * `sstore` specs and circuit
>   
>   * we use is_zero for RLC encoded value without summing up there
>   * the reason why I did those was I was simply following what I used for `sload`: see this comment in [feat: add `sload` op (non-bus-mapping version) zkevm-circuits#334 (comment)](https://github.com/appliedzkp/zkevm-circuits/pull/334#discussion_r821535077). But I didn't realize that `sstore` case is different from `sload`'s.

Would like to have confirmation from @ed255!

ed255: > It's possible that an RLC encoded value for a non-zero value to be zero mod Fp, right? I think it's dangerous to directly use the RLC value to do this. It'd be safer to sum up all limbs and assert 0.

It's possible, but the probability of this happening (assuming the randomness fulfills the properties we expect to use it for the Fiat–Shamir heuristic) should be negligible.

Finding a word !=0 that has an RLC encoding of 0 means a collision.  To use RLC, the probability of collision should be negligible.  Otherwise many things would break.  For instance, just with this example we could have a stack push of such word !=0, and a following stack pop where the prover could use as advice the 0 word, and all constraints would pass! (the constraints would pass because we track pushed and popped values of the stack via the rw_table, which contains the RLC encoded words).

So I don't think there's the need to sum all limbs and assert equality to 0.

@han0110 @barryWhiteHat what do you think about this?

icemelon: @ed255 This makes sense to me. I agree that we can just use RLC value to check if it's 0.

specs/opcode/15ISZERO.md
```diff
@@ -0,0 +1,26 @@
+# ISZERO opcodes
```
icemelon, line 1:
```suggestion
# ISZERO opcode
```

specs/opcode/15ISZERO.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Exceptions
+
+1. stack underflow:
+   - the stack is empty: `1024 == stack_pointer`
+2. out of gas: remaining gas is not enough
+
+## Code
+
+See `src/zkevm_specs/opcode/iszero.py`
```
icemelon, line 26:
```suggestion
See `src/zkevm_specs/evm/execution/iszero.py`
```

icemelon (approved):
Two minor comments. otherwise LGTM

PR #162, opened 2022-03-24
Move stack and memory addresses from key3 to key2

z2trillion:
Fixes https://github.com/appliedzkp/zkevm-specs/issues/160

### DIFF
--- a/specs/tables.md
+++ b/specs/tables.md
@@ -42,10 +42,10 @@ Type sizes:
 > - **address**: 160 bits
 > - **callID**: 32 bits
 > - **stackPointer**: 10 bits
-> - **memoryAddress**: 160 bits
+> - **memoryAddress**: 32 bits
 > - **Memory -> value, valuePrev**: 1 byte
 > - **storageKey**: field size, RLC encoded (Random Linear Combination)
-> - **value, valuePrev**: variable size, depending on Key 0 (Tag) and Key 3 (field tag) where appropiate.
+> - **value, valuePrev**: variable size, depending on Key 0 (Tag) and Key 3 (field tag) where appropriate.
 > - **Key2** is reserved for stack, memory, and account addresses.
 > - **Key4** is reserved for RLC encoded key type
 
--- a/specs/tables.md
+++ b/specs/tables.md
@@ -38,64 +38,64 @@ Proved by the state circuit.
 
 Type sizes:
 
-> - **txID**: 24 bytes
-> - **address**: 160 bytes
-> - **callID**: 25 bytes
-> - **stackPointer**: 10 bytes
-> - **memoryAddress**: 40 bytes
+> - **txID**: 32 bits
+> - **address**: 160 bits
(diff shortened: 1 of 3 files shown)


z2trillion:
@ed255

specs/tables.md
```diff
@@ -38,64 +38,64 @@ Proved by the state circuit.
...
-> - **txID**: 24 bytes
-> - **address**: 160 bytes
-> - **callID**: 25 bytes
-> - **stackPointer**: 10 bytes
-> - **memoryAddress**: 40 bytes
+> - **txID**: 32 bits
+> - **address**: 160 bits
+> - **callID**: 32 bits
+> - **stackPointer**: 10 bits
+> - **memoryAddress**: 160 bits
```
ed255, line 45: Even though the `memoryAddress` is assigned to a column which is 160 bits, I think it's worth specifying here that `memoryAddress` is 40 bits, or 32 bits actually, to follow the comment about maximum memory address size (so that we know the corresponding range constraint is guaranteed somewhere)

src/zkevm_specs/state.py
```diff
@@ -127,21 +121,17 @@ def check_memory(row: Row, row_prev: Row):
     if not all_keys_eq(row, row_prev) and row.is_write == 0:
         assert row.value == 0
 
-    # 2. mem_addr in range
-    mem_addr = get_mem_addr(row)
-    assert_in_range(mem_addr, 0, MAX_MEMORY_ADDRESS)
-
-    # 3. value is a byte
+    # 2. value is a byte
```
ed255, line 124: Following my comment about defining memory address as 32 bits.

I'm not sure if the EVM circuit will make sure that the memory address is range constrained already, but we could easily add a range constraint here by constraining that `key2_limbs[2:]` are 0.

ed255 (commented):
Overall looks good to me!  Please take a look at my suggestion to constraint the memory address to be 32 bits (instead of letting it be 160 bits).  Let me know what you think!

specs/tables.md
z2trillion: I was actually thinking that we would "allow" memoryAddress to have up to 160 bits, knowing that that's actually impossible because you wouldn't be able to pay the gas cost.

I agree it's a better idea to add explicit constraints in the state circuit that it must fit into 32 bits.

src/zkevm_specs/state.py
z2trillion: The only parts of the evm circuit that constraint the memory address size are related to the gas cost calculations. Now that I think about it, it's possible that the gas cost calculation could overflow for memory addresses > 2^127 because of the square in the formula.

z2trillion: we won't have to worry about this with the 32 bit constraint for memory addresses.

ed255 (approved):
LGTM! Thanks for the PR!

miha-stopar (approved):
LGTM

PR #167, opened 2022-03-30
Fix tx circuit RLC usage with 64 bytes input

ed255:
Today I merged https://github.com/appliedzkp/zkevm-specs/pull/150 and I broke the tests in master.  This is the fix


ed255:
> Sorry that I made the change of `RLC` in #82 without warning. Thanks for fixing!

No worries :smile:

PR #170, opened 2022-04-05
change SSTORE_CLEARS_SCHEDULE to 4800 according to EIP-3529, which wa…

lispc:
…s included in London fork

ref:  <https://github.com/ethereum/go-ethereum/blob/06540146524222f648d8b19c5bcb00a8fd3dd339/params/protocol_params.go#L65-L69>


CPerezz (approved):
LGTM!

Thanks for fixing this!

ed255:
Currently our rollup testnet is setup to disable some EIPs: https://github.com/appliedzkp/zkevm-chain/tree/master/docs#layer-2---go-ethereum , and one of them is EIP-3529, so I wonder how should we proceed here.

han0110:
I am wondering why EIP-3529 is disabled, since its main dependency EIP-2929 is supported, and we are ignoring EIP-2930 typed transaction, so it seems to be fine to have EIP-3529. Is there any specific reason to disable it?

ed255:
> I am wondering why EIP-3529 is disabled, since its main dependency EIP-2929 is supported, and we are ignoring EIP-2930 typed transaction, so it seems to be fine to have EIP-3529. Is there any specific reason to disable it?

Maybe @pinkiebell can explain the reason to disable that EIP?

pinkiebell:
EIP-3529 depends on [2930](https://eips.ethereum.org/EIPS/eip-2930) and the access list checks depends on each other in geth. It would be possible to branch around/decouple that in geth.

I'm open on suggestions how to proceed here. We could also enable the typed transaction envelope + 2930 on the testnet but I'm not up to date on how far the support  for this EIPs are in the circuit.

pinkiebell:
I think that this can be merged. I took a look again and saw no other hard dependencies for this EIP.
Sorry for the fluff, must have been a bad memory regarding another EIP :)

ed255:
> I think that this can be merged. I took a look again and saw no other hard dependencies for this EIP. Sorry for the fluff, must have been a bad memory regarding another EIP :)

Thanks for taking a look at this!

han0110:
Yeah, I think so, since we already disable EIP-2718, it automatically rejects EIP-2930 typed transactions ([ref](https://github.com/ethereum/go-ethereum/blob/65825cd134378df2eb7bd1b78f3bf3bf1ba613b4/core/tx_pool.go#L588-L590)). But we still can enable EIP-2929 and EIP-3529 to make the gas calculation up-to-date to mainnet.

ed255 (approved):
LGTM!

PR #178, opened 2022-04-09
CALLDATACOPY: Read from caller's memory for internal call

### LINKED ISSUE #134: `CALLDATACOPY` should read caller's memory
Currently `CALLDATACOPY` is implemented to read its own memory when `not is_root` (without `call_id` specified):

https://github.com/appliedzkp/zkevm-specs/blob/af758f49341fa11f2bc8d47a0e71d5b3eed1de13/src/zkevm_specs/evm/execution/memory_copy.py#L22-L33

We can fix this by extend the field `tx_id` of `CopyToMemoryAuxData` to `src_id`, to contain a `tx_id` when `from_tx`, and `caller_id` when `not from_tx`.

And with a slight change to the code above:

```diff
 for i in range(MAX_COPY_BYTES): 
     if not buffer_reader.read_flag(i): 
         byte = FQ.zero() 
     elif aux.from_tx == 1: 
         byte = instruction.tx_context_lookup( 
-            aux.tx_id, TxContextFieldTag.CallData, aux.src_addr + i
+            aux.src_id, TxContextFieldTag.CallData, aux.src_addr + i
         ) 
     else: 
-        byte = instruction.memory_lookup(RW.Read, aux.src_addr + i)
+        byte = instruction.memory_lookup(RW.Read, aux.src_addr + i, caller_id=aux.src_id)
     buffer_reader.constrain_byte(i, byte) 
     if buffer_reader.has_data(i): 
         instruction.constrain_equal(byte, instruction.memory_lookup(RW.Write, aux.dst_addr + i)) 
```

And `CALLDATACOPY` should also set the `src_id` correctly.

roynalnaruto:
Closes #134 

Circuits PR:
- [ ] https://github.com/appliedzkp/zkevm-circuits/pull/450

### DIFF
--- a/src/zkevm_specs/evm/execution/calldatacopy.py
+++ b/src/zkevm_specs/evm/execution/calldatacopy.py
@@ -15,11 +15,12 @@ def calldatacopy(instruction: Instruction):
     memory_offset, length = instruction.memory_offset_and_length(memory_offset_word, length_word)
     data_offset = instruction.rlc_to_fq(data_offset_word, N_BYTES_MEMORY_ADDRESS)
 
-    tx_id = instruction.call_context_lookup(CallContextFieldTag.TxId, RW.Read)
     if instruction.curr.is_root:
-        call_data_length = instruction.tx_context_lookup(tx_id, TxContextFieldTag.CallDataLength)
+        src_id = instruction.call_context_lookup(CallContextFieldTag.TxId, RW.Read)
+        call_data_length = instruction.tx_context_lookup(src_id, TxContextFieldTag.CallDataLength)
         call_data_offset: Expression = FQ.zero()
     else:
+        src_id = instruction.call_context_lookup(CallContextFieldTag.CallerId, RW.Read)
         call_data_length = instruction.call_context_lookup(
             CallContextFieldTag.CallDataLength, RW.Read
         )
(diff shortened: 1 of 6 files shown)


src/zkevm_specs/evm/execution/calldatacopy.py
```diff
@@ -15,11 +15,12 @@ def calldatacopy(instruction: Instruction):
     memory_offset, length = instruction.memory_offset_and_length(memory_offset_word, length_word)
     data_offset = instruction.rlc_to_fq(data_offset_word, N_BYTES_MEMORY_ADDRESS)
 
-    tx_id = instruction.call_context_lookup(CallContextFieldTag.TxId, RW.Read)
     if instruction.curr.is_root:
-        call_data_length = instruction.tx_context_lookup(tx_id, TxContextFieldTag.CallDataLength)
+        src_id = instruction.call_context_lookup(CallContextFieldTag.TxId, RW.Read)
+        call_data_length = instruction.tx_context_lookup(src_id, TxContextFieldTag.CallDataLength)
```
han0110, line 20: The `CallDataLength` in `tx_table` is actually the length of input of transaction, in creation transaction case, the `call_data` is actually empty, so the length of input should not be used. We could read `CallDataLength` from call context which is guaranteed to be length of `call_data`. Note that it adds an extra `rw_table` read, so both markdown and test need an update.

This seems also confusing (my apology)... Perhaps we should rename `CallData*` to `Input*` in `tx_table` to explicitly make them different, or add some more explanation on that.

```suggestion
        call_data_length = instruction.call_context_lookup(CallContextFieldTag.CallDataLength)
```

han0110 (commented):
Thanks a lot to fix this! There is a small issue of reading `CallDataLength` from `tx_table`, except that all look good to me!

src/zkevm_specs/evm/execution/calldatacopy.py
roynalnaruto: We can do the renaming in a separate PR.

Very good catch, I've fixed this part :)

han0110 (approved):
LGTM! Thanks!

ChihChengLiang (approved):
LGTM

PR #185, opened 2022-04-14
Find and replace state_write_counter with reversible_write_counter

z2trillion:
This is a followup to https://github.com/appliedzkp/zkevm-circuits/pull/409

### DIFF
--- a/src/zkevm_specs/evm/execution/call.py
+++ b/src/zkevm_specs/evm/execution/call.py
@@ -67,8 +67,8 @@ def call(instruction: Instruction):
     is_reverted_by_caller = is_success.expr() == FQ(1) and reversion_info.is_persistent == FQ(0)
     if is_reverted_by_caller:
         # Propagate rw_counter_end_of_reversion when callee succeeds but one of callers revert at some point.
-        # Note that we subtract it with current caller's state_write_counter as callee's endpoint, where caller's
-        # state_write_counter here is added by 1 due to adding callee to access list
+        # Note that we subtract it with current caller's reversible_write_counter as callee's endpoint, where caller's
+        # reversible_write_counter here is added by 1 due to adding callee to access list
         instruction.constrain_equal(
             callee_reversion_info.rw_counter_end_of_reversion,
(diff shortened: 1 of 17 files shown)


src/zkevm_specs/evm/execution/call.py
```diff
@@ -151,7 +151,7 @@ def call(instruction: Instruction):
             (CallContextFieldTag.StackPointer, instruction.curr.stack_pointer + 6),
             (CallContextFieldTag.GasLeft, instruction.curr.gas_left - gas_cost - callee_gas_left),
             (CallContextFieldTag.MemorySize, next_memory_size),
-            (CallContextFieldTag.StateWriteCounter, instruction.curr.state_write_counter + 1),
+            (CallContextFieldTag.StateWriteCounter, instruction.curr.reversible_write_counter + 1),
```
han0110, line 154: This should also be renamed.

```suggestion
            (CallContextFieldTag.ReversibleWriteCounter, instruction.curr.reversible_write_counter + 1),
```

han0110 (commented):
I also did the same thing in #100, but it's nice to do this in a separate PR. (btw for circuit, I updated the naming in appliedzkp/zkevm-circuits#419.)

src/zkevm_specs/evm/execution/call.py
z2trillion: thanks, i've updated those.

z2trillion:
> I also did the same thing in #100, but it's nice to do this in a separate PR. (btw for circuit, I updated the naming in [appliedzkp/zkevm-circuits#419](https://github.com/appliedzkp/zkevm-circuits/pull/419).)

Great minds think alike:

I am renaming state write -> reversible write in https://github.com/appliedzkp/zkevm-docs too.

han0110 (approved):
Ah I didn't see that you already planed to fix it in spec and in doc. All renaming look good to me now!

ed255 (approved):
LGTM! Thanks for the update!

PR #187, opened 2022-04-20
Update tx circuit according to implementation

ed255:
I have nearly finished the implementation of the tx circuit (See [here](https://github.com/appliedzkp/zkevm-circuits/blob/feature%2Ftx-circuit/zkevm-circuits/src/tx_circuit.rs) and [here](https://github.com/appliedzkp/zkevm-circuits/blob/feature%2Ftx-circuit/zkevm-circuits/src/tx_circuit/sign_verify.rs) for a preview), and during the implementation I diverged a bit from the spec.  In this PR I update the spec to match what I have implemented.


src/zkevm_specs/tx.py
```diff
@@ -137,30 +158,33 @@ def assign(cls, signature: KeyAPI.Signature, pub_key: KeyAPI.PublicKey, msg_hash
...
 
-    def verify(self, is_enabled: FQ, assert_msg: str):
-        msg_hash = bytes(reversed(self.msg_hash.limbs))
-        sig_r = int.from_bytes(self.signature[0].limbs, "little")
-        sig_s = int.from_bytes(self.signature[1].limbs, "little")
+    def verify(self, assert_msg: str):
+        msg_hash = bytes(reversed(self.msg_hash.to_le_bytes()))
+        sig_r = int.from_bytes(self.signature[0].to_le_bytes(), "little")
+        sig_s = int.from_bytes(self.signature[1].to_le_bytes(), "little")
         signature = KeyAPI.Signature(vrs=[0, sig_r, sig_s])
```
adria0, line 165: `v` always zero?

src/zkevm_specs/tx.py
```diff
@@ -186,43 +213,49 @@ def assign(
         return cls(self_pub_key_hash, self_address, self_msg_hash_rlc, self_ecdsa_chip)
 
     def verify(self, keccak_table: KeccakTable, randomness: FQ, assert_msg: str):
-        is_enabled = FQ(1 - (self.msg_hash_rlc == 0))  # 1 - is_zero(self.msg_hash_rlc)
+        is_not_padding = FQ(1 - (self.address == 0))  # 1 - is_zero(self.address)
```
adria0, line 216: Why the padding check is now using `address` instead `msg_hash_rlc`?

adria0 (commented):
Additional comment:

- In `tx_row_index = tx_index * Tag.TxSignHash`, is not `tx_row_index = tx_index * Tag.TxCallData` ?

src/zkevm_specs/tx.py
ed255: Ah, good question!  Originally I intended to use `msg_hash == 0` to disable a transaction (and make it a padding one).  But for that I was assuming I would have an ECDSA Verify chip that would output 1 or 0 depending on whether the signature was correct.  But the current ECDSA Verify chip only supports correct signatures, and fails the constraints on invalid ones; so I decided to find a "dummy" input + witness values for padding transactions to pass the ECDSA verification.  And it turns out that `msg_hash = 0` is an [invalid value for a signature](https://crypto.stackexchange.com/questions/9379/is-there-any-weak-message-for-an-ecdsa-signature)!
So I had to change it to use `address == 0` to signal the padding transaction.  Since the address is the hash of the public key (which we verify via the keccak table), it's easy to deactivate the lookup when `address == 0` and use dummy values for the signature verification.

src/zkevm_specs/tx.py
ed255: Here this signature object is used to verify a signature against the message hash and public key.  For an ECDSA signature verification, the `v` value is not needed (Actually an ECDSA signature is only [`(r, s)`](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm#Signature_generation_algorithm)).  That's why I set `v = 0`, because it's not used here.
The `v` value is used when recovering the public key from a signature and message.

ed255:
>     * In `tx_row_index = tx_index * Tag.TxSignHash`, is not `tx_row_index = tx_index * Tag.TxCallData` ?

If you pay close attention, the tx table is split into two sections: https://github.com/appliedzkp/zkevm-specs/blob/ba1f2c5875a5d16f5f7aef58964b404302b77acf/specs/transactions-proof.md#circuit-behaviour-shortcut-1
The first section contains all tags up to `TxSignHash` for each tx, and the second section contains all the `TxCallData` of all transactions.  The value `tx_row_index` is the transaction offset of the first section (which is the transaction index times the number of fields per transaction in the first section: `TxSignHash = 11`)

src/zkevm_specs/tx.py
adria0: Thanks for the detailed comment!

So, we need to remove this comment, isn't it?

https://github.com/appliedzkp/zkevm-specs/pull/187/files#diff-095d7a59798d973b0472d088a362302355f91c3acd241fc396758d705555c8d9R183

adria0:
> 



> > ```
> > * In `tx_row_index = tx_index * Tag.TxSignHash`, is not `tx_row_index = tx_index * Tag.TxCallData` ?
> > ```
> 
> If you pay close attention, the tx table is split into two sections: https://github.com/appliedzkp/zkevm-specs/blob/ba1f2c5875a5d16f5f7aef58964b404302b77acf/specs/transactions-proof.md#circuit-behaviour-shortcut-1 The first section contains all tags up to `TxSignHash` for each tx, and the second section contains all the `TxCallData` of all transactions. The value `tx_row_index` is the transaction offset of the first section (which is the transaction index times the number of fields per transaction in the first section: `TxSignHash = 11`)

ok! I misunderstood this shortcut! :)

src/zkevm_specs/tx.py
ed255: > So, we need to remove this comment, isn't it?
> 
> https://github.com/appliedzkp/zkevm-specs/pull/187/files#diff-095d7a59798d973b0472d088a362302355f91c3acd241fc396758d705555c8d9R183

Yes!  Thanks for noticing this!  I've updated this in https://github.com/appliedzkp/zkevm-specs/pull/187/commits/8c37a3749a5788ab394e919ff97eafe351337d6c

miha-stopar (commented):
Seems cool, but I need to dive a bit more. I have two questions which are not specific to this update:

- Is there RLC over all transaction bytes computed, so that it can be compared to the keccak input? Asking because I am interested how you can connect transactions to their corresponding call data at the end of the table?
- There is mentioned that MPT update can't be done with a single lookup because the call data is of variable size. I think this could be done with some tweaking inside MPT, or is there some other reason?

ed255:
>     * Is there RLC over all transaction bytes computed, so that it can be compared to the keccak input? Asking because I am interested how you can connect transactions to their corresponding call data at the end of the table?

The spec currently defines a final description of what the tx circuit should look like (but misses some details), and then a shortcut which moves some checks from the circuit to the verifier; which contains all the details to have an implementation.

In the final version, the tx table should contain all the tx fields (including the call data) such that the RLP encoding of the transaction is computed and stored in an RLC value which would then be used with the keccak lookup table to get the transaction hash.

In the current shortcut, we skip both the RLP encoding of the transaction and the keccak lookup to get the transaction hash.  In this shortcut we assume that the verifier prepares the tx table contents and computes the RLP + Hash of the transaction outside of the circuit, and passes this value as a row for each transaction.  So currently the circuit doesn't check anything in regards of the call data (which now is at the end of the table), because we assume the verifier calculates the transaction hash correctly.

The reason why the call data is at the end of the table in this shortcut is because we're not using an ECDSA verification table yet: we're using an ECDSA verification chip directly.  To use it we need copy constraints, and for that we need the inputs to the ECDSA verification at fixed offsets (and the call data is dynamic length), so we solve this by moving all the call data at the end.  In the final version we'll have an ECDSA verification table, removing the need for copy constraints and thus we will allow having the transaction fields at non-fixed offsets.

>     * There is mentioned that MPT update can't be done with a single lookup because the call data is of variable size. I think this could be done with some tweaking inside MPT, or is there some other reason?

The truth is that I made the assumption of requiring several lookups without much thought.  There was no particular reason.  Now that you say this, I think this may introduce confusion, so I'll just remove the sentence from the spec.  The main point was that to work on the integration with the MPT in a future iteration.

src/zkevm_specs/tx.py
```diff
@@ -125,6 +140,12 @@ def __init__(
         self.signature = signature
         self.pub_key = pub_key
         self.msg_hash = msg_hash
+        self.pub_key_x_bytes = pub_key[0].to_le_bytes()
+        self.pub_key_y_bytes = pub_key[1].to_le_bytes()
+        self.msg_hash_bytes = msg_hash.to_le_bytes()
+        # NOTE: The circuit must constraint that all elements in the `*_bytes`
```
miha-stopar, line 146: nitpick: constraint -> constrain

miha-stopar:
> > ```
> > * Is there RLC over all transaction bytes computed, so that it can be compared to the keccak input? Asking because I am interested how you can connect transactions to their corresponding call data at the end of the table?
> > ```
> 
> 
> The spec currently defines a final description of what the tx circuit should look like (but misses some details), and then a shortcut which moves some checks from the circuit to the verifier; which contains all the details to have an implementation.
> 
> 
> In the final version, the tx table should contain all the tx fields (including the call data) such that the RLP encoding of the transaction is computed and stored in an RLC value which would then be used with the keccak lookup table to get the transaction hash.
> 
> 
> In the current shortcut, we skip both the RLP encoding of the transaction and the keccak lookup to get the transaction hash.  In this shortcut we assume that the verifier prepares the tx table contents and computes the RLP + Hash of the transaction outside of the circuit, and passes this value as a row for each transaction.  So currently the circuit doesn't check anything in regards of the call data (which now is at the end of the table), because we assume the verifier calculates the transaction hash correctly.
> 
> 
> The reason why the call data is at the end of the table in this shortcut is because we're not using an ECDSA verification table yet: we're using an ECDSA verification chip directly.  To use it we need copy constraints, and for that we need the inputs to the ECDSA verification at fixed offsets (and the call data is dynamic length), so we solve this by moving all the call data at the end.  In the final version we'll have an ECDSA verification table, removing the need for copy constraints and thus we will allow having the transaction fields at non-fixed offsets.
> 
> 
> > ```
> > * There is mentioned that MPT update can't be done with a single lookup because the call data is of variable size. I think this could be done with some tweaking inside MPT, or is there some other reason?
> > ```
> 
> 
> The truth is that I made the assumption of requiring several lookups without much thought.  There was no particular reason.  Now that you say this, I think this may introduce confusion, so I'll just remove the sentence from the spec.  The main point was that to work on the integration with the MPT in a future iteration.

Cool, thanks for a detailed expla

> > ```
> > * Is there RLC over all transaction bytes computed, so that it can be compared to the keccak input? Asking because I am interested how you can connect transactions to their corresponding call data at the end of the table?
> > ```
> 
> 
> The spec currently defines a final description of what the tx circuit should look like (but misses some details), and then a shortcut which moves some checks from the circuit to the verifier; which contains all the details to have an implementation.
> 
> 
> In the final version, the tx table should contain all the tx fields (including the call data) such that the RLP encoding of the transaction is computed and stored in an RLC value which would then be used with the keccak lookup table to get the transaction hash.
> 
> 
> In the current shortcut, we skip both the RLP encoding of the transaction and the keccak lookup to get the transaction hash.  In this shortcut we assume that the verifier prepares the tx table contents and computes the RLP + Hash of the transaction outside of the circuit, and passes this value as a row for each transaction.  So currently the circuit doesn't check anything in regards of the call data (which now is at the end of the table), because we assume the verifier calculates the transaction hash correctly.
> 
> 
> The reason why the call data is at the end of the table in this shortcut is because we're not using an ECDSA verification table yet: we're using an ECDSA verification chip directly.  To use it we need copy constraints, and for that we need the inputs to the ECDSA verification at fixed offsets (and the call data is dynamic length), so we solve this by moving all the call data at the end.  In the final version we'll have an ECDSA verification table, removing the need for copy constraints and thus we will allow having the transaction fields at non-fixed offsets.
> 
> 
> > ```
> > * There is mentioned that MPT update can't be done with a single lookup because the call data is of variable size. I think this could be done with some tweaking inside MPT, or is there some other reason?
> > ```
> 
> 
> The truth is that I made the assumption of requiring several lookups without much thought.  There was no particular reason.  Now that you say this, I think this may introduce confusion, so I'll just remove the sentence from the spec.  The main point was that to work on the integration with the MPT in a future iteration.

Cool, thanks for a detailed explanation, Edu!

miha-stopar (approved):
LGTM

adria0 (approved):
LGTM

PR #192, opened 2022-05-02
[Keccak] Add Start and End state_tags and references

ChihChengLiang:
- Fix the state transition by replacing Null state_tag with explicit Start and End tags.
- Add references to the sparse representation idea and reference implementation.


CPerezz (approved):
LGTM!

Thanks for this!

PR #194, opened 2022-05-07
Copy Circuit

icemelon:
This PR brings the copy circuit that eliminates the multi-slot opcodes such as CALLDATACOPY, CODECOPY, and LOG.


roynalnaruto:
@icemelon You will need to edit this PR from "draft" to "open"

src/zkevm_specs/evm/typing.py
```diff
@@ -615,3 +648,141 @@ def _append(
...
         return self
+
+
+class CopyCircuit:
+    rows: List[CopyCircuitRow]
+    pad_rows: List[CopyCircuitRow]
+
+    def __init__(self) -> None:
+        self.rows = []
+        self.pad_rows = [CopyCircuitRow(*[FQ(0)] * 18)] * 2
```
ed255, line 659: This is an example of how padding rows work right?  Not padding in the sense that a 0 is read after the src end; but in the sense that the circuit has a fixed number of rows and maybe not all of them are used.

If that's the case, could you add a line in the markdown doc saying that padding rows are set to all 0s, and that these values always pass the constraints?

specs/copy-proof.md
```diff
@@ -0,0 +1,57 @@
...
+- Construct the LT gadget to compare `Address` and `AddressEnd` in the read operation. If `Address >= AddressEnd`, constrain `Padding == 1`
+- Constrain `Padding[1] == 0` as the write operation is never padded.
+
+Third, the circuit adds the lookup arguments to the corresponding tables.
+
+- When `Type` is `Memory` or `is_memory == 1` and `Padding == 0`, look up the `Value` to `rw_table` with `Memory` tag.
+- When `Type` is `TxCalldata` or `is_tx_calldata == 1` and `Padding == 0`, look up the `Value` to `tx_table`.
+- When `Type` is `Bytecode` or `is_bytecode == 1` and `Padding == 0`, look up the `Value` and `IsCode` to `bytecode_table`
+- When `Type` is `TxLog` or `is_tx_log == 1`, look up the `Value` to `rw_table` with `TxLog` tag.
+
```
ed255, line 54: I'm wondering about the scenario where the CopyCircuit contains a copy that is not requested by the EVM circuit.  If my reasoning is correct, that shouldn't be a problem because:
1. The copy circuit can only write to memory and bytecode.  Writing to bytecode doesn't affect the execution because bytecode is accessed by hash.  Writing to memory without a request from the EVM would be an issue.
2. All memory writes in the `rw_table` must match 1 to 1 with the EVM circuit (this is something we will guarantee by design).  This means that the copy circuit can't write to memory that the EVM hasn't accounted for.
3. So the only case in which the CopyCircuit contains extra non-padding rows, is when it has a duplicate copy, which doesn't affect anything as it will act as an idempotent function (the same `rw_counter` will be used).

Does this reasoning make sense?

ed255 (commented):
Overall looks very good!
I've added a few notes here and there but they are mainly topics for discussion and some small suggestions.

roynalnaruto:
> Overall looks very good!
> 
> I've added a few notes here and there but they are mainly topics for discussion and some small suggestions.

Thanks for the suggestions. They all make sense and I'll update the PR soon :)

specs/copy-proof.md
```diff
@@ -0,0 +1,57 @@
+# Copy Proof
+
+The copy proof checks the values in the copy table and applies the lookup arguments to the corresponding tables to check if the value read from and write to data source is correct.
+It also checks the padding behavior that the value read from an out-of-boundary address is 0.
+
+## Circuit Layout
+
+First, copy circuit contains 14 columns from the [copy table](./tables.md#copytable) with the same witness assignment.
+Every two rows in the copy circuit represent a copy step where the first row is a read operation and the second is a write operation.
```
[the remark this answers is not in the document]
icemelon: If you put a copy step in a single row, you will almost double the number of columns and double the number of lookups. Right now, you only need to have 4 lookups (rw memory, rw txlog, tx table, bytecode table). If you put both src and dst in a single row, you need to have 8 lookups for both src and dst.

src/zkevm_specs/copy_circuit.py
```diff
@@ -0,0 +1,117 @@
...
+        cs.constrain_equal(rows[0].rwc_inc_left - rw_diff, rows[1].rwc_inc_left)
+    with cs.condition(rows[0].q_last) as cs:
+        # rwc_inc_left == 1 for last row in the copy slot
+        cs.constrain_equal(rows[0].rwc_inc_left, rw_diff)
+
+
+def verify_step(cs: ConstraintSystem, rows: Sequence[CopyCircuitRow], tables: Tables):
+    with cs.condition(rows[0].q_step):
+        # lookup to copy pairs
+        tables.fixed_lookup(FQ(FixedTableTag.CopyPairs), rows[0].tag, rows[1].tag)
```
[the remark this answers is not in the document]
icemelon: yes, thought about this too. I think we can remove it.

src/zkevm_specs/evm/table.py
```diff
@@ -71,6 +72,24 @@ def table_assignments(self) -> List[FixedTableRow]:
...
+        elif self == FixedTableTag.CopyPairs:
+            return [
+                FixedTableRow(FQ(self), FQ(src_type), FQ(dst_type), FQ(0))
+                for src_type, dst_type in [
+                    # calldatacopy (internal)
+                    (CopyDataTypeTag.Memory, CopyDataTypeTag.Memory),
+                    # calldatacopy (root)
+                    (CopyDataTypeTag.TxCalldata, CopyDataTypeTag.Memory),
+                    # create/create2 (root)
+                    (CopyDataTypeTag.TxCalldata, CopyDataTypeTag.Bytecode),
```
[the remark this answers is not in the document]
icemelon: @roynalnaruto added this. Could you address this question?

src/zkevm_specs/copy_circuit.py
```diff
@@ -0,0 +1,117 @@
...
+        # bytes_left == 1 for last step
+        cs.constrain_zero(rows[1].q_last * (1 - rows[0].bytes_left))
+        # bytes_left == bytes_left_next + 1 for non-last step
+        cs.constrain_zero((1 - rows[1].q_last) * (rows[0].bytes_left - rows[2].bytes_left - 1))
+        # write value == read value
+        cs.constrain_equal(rows[0].value, rows[1].value)
+        # value == 0 when is_pad == 1 for read
+        cs.constrain_zero(rows[0].is_pad * rows[0].value)
+        # is_pad == 1 - (src_addr < src_addr_end) for read row
```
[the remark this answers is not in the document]
icemelon: `src_len` is still not accurate tho as it could be confused for the number of bytes to be copied. But indeed this column holds the value of `src_addr + num_bytes`. How about calling it `src_boundary`?

specs/copy-proof.md
```diff
@@ -0,0 +1,57 @@
...
+- The transition constraints for `RwCounter` and `RwcIncreaseLeft` column
+    - define `rw_diff` to be 1 if the `Type` is `Memory` or `TxLog` and `Padding` is 0 in the current row; otherwise 0.
+    - when it's not the last row in a copy event (`is_last == 0`), `RwCounter` increases by `rw_diff` and `RwcIncreaseLeft` decrases by `rw_diff`.
+    - when it's the last row in a copy event (`is_last == 1`), `RwcIncreaseLeft` is equal to `rw_diff`.
+
+Second, the circuit adds the constraints for every copy step in the circuit, when `q_step` is 1.
+
+- Look up the copy type pair `(Type, Type[1])` in a fixed table to make sure it's a valid copy step.
+- Constrain the transition for `BytesLeft`
+    - when it's not the last step (`is_last[1] != 1`), increase by 1 in the next step
```
DreamWuGit, line 41: decrease by 1 in the next step not `increase`?

src/zkevm_specs/copy_circuit.py
```diff
@@ -0,0 +1,114 @@
...
+        # rwc_inc_left == 1 for last row in the copy slot
+        cs.constrain_equal(rows[0].rwc_inc_left, rw_diff)
+
+
+def verify_step(cs: ConstraintSystem, rows: Sequence[CopyCircuitRow]):
+    with cs.condition(rows[0].q_step):
+        # bytes_left == 1 for last step
+        cs.constrain_zero(rows[1].is_last * (1 - rows[0].bytes_left))
+        # bytes_left == bytes_left_next + 1 for non-last step
+        cs.constrain_zero((1 - rows[1].is_last) * (rows[0].bytes_left - rows[2].bytes_left - 1))
```
DreamWuGit, line 63: not sure why use `rows[2].bytes_left` , refer to last step second row?

DreamWuGit:
just have a quick look ,one point to discuss:  seems to me maybe there is possible duplicated constraint between copy circuit and state circuit.   for example,  read value constrain to be equal to writing value at the same address.  in the state circuit should constrain this (even though not implemented yet) by design,   but it is not big problem from my side.

src/zkevm_specs/copy_circuit.py
```diff
@@ -23,54 +23,51 @@ def lt(lhs: Expression, rhs: Expression, n_bytes: int) -> FQ:
...
+    with cs.condition(rows[0].is_last) as cs:
         # rwc_inc_left == 1 for last row in the copy slot
         cs.constrain_equal(rows[0].rwc_inc_left, rw_diff)
 
 
-def verify_step(cs: ConstraintSystem, rows: Sequence[CopyCircuitRow], tables: Tables):
+def verify_step(cs: ConstraintSystem, rows: Sequence[CopyCircuitRow]):
     with cs.condition(rows[0].q_step):
-        # lookup to copy pairs
-        tables.fixed_lookup(FQ(FixedTableTag.CopyPairs), rows[0].tag, rows[1].tag)
```
roynalnaruto: @icemelon Why did you remove the lookup to fixed table for the `(src, dst)` copy-pair?

roynalnaruto (commented):
Small question WRT some removed checks

src/zkevm_specs/copy_circuit.py
roynalnaruto: If `rows[0]` is the read step (given that condition `rows[0].q_step`) then we want to constrain:
`rows[0].bytes_left == rows[2].bytes_left + 1` as `rows[2]` is the next read step if `rows[1]` was not the last step.

specs/copy-proof.md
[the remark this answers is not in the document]
ed255: Thanks for the explanation!  The 2 row approach seems much better :)

specs/tables.md
```diff
@@ -271,3 +271,52 @@ Columns expressions in circuit:
...
+Unlike other lookup tables, the copy table is a virtual table. The lookup entry is not a single row in the table, and not every row corresponds to a lookup entry.
+Instead, a lookup entry is constructed from the first two rows in each copy event as
+`(q_first, ID, Type, ID[1], Type[1], LogId[1], Address, AddressEnd, Address[1], BytesLeft, RwCounter, RwcIncreaseLeft)`, where `q_first` is 1 and `Column[1]` indicates the next row in the corresponding column.
+
+The table below lists all of copy pairs supported in the copy table:
+- Copy from Tx call data to memory (`CALLDATACOPY`).
+- Copy from caller/callee memory to callee/caller memory (`CALLDATACOPY`, `RETURN` (not create), `RETURNDATACOPY`).
+- Copy from bytecode to memory (`CODECOPY`, `EXTCODECOPY`).
+- Copy from memory to bytecode (`CREATE`, `CREATE2`, `RETURN` (create))
+
```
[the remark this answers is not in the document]
ed255: Cool.  I've added an issue to track this https://github.com/privacy-scaling-explorations/zkevm-specs/issues/228
Feel free to assign it to yourself in case you'd like or plan to work on it!

ed255:
> just have a quick look ,one point to discuss: seems to me maybe there is possible duplicated constraint between copy circuit and state circuit. for example, read value constrain to be equal to writing value at the same address. in the state circuit should constrain this (even though not implemented yet sort of) by design, but it is not big problem from my side.

Are you referring to the constraint that the value in the write row is the same as the value in the read row (of the same copy step)?  If that's the case, I don't see how it's redundant as the CopyCircuit is copying between two different places.  Could you elaborate on this duplicate constraint between Copy Circuit and State Circuit; and point to the exact constraint in the python spec?

Thanks!

src/zkevm_specs/copy_circuit.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # contrain the transition for `rw_counter` and `rwc_inc_left`
+    rw_diff = (1 - rows[0].is_pad) * (rows[0].is_memory + rows[0].is_tx_log)
+    with cs.condition(1 - rows[0].is_last) as cs:
+        # not last row
+        cs.constrain_equal(rows[0].rw_counter + rw_diff, rows[1].rw_counter)
+        cs.constrain_equal(rows[0].rwc_inc_left - rw_diff, rows[1].rwc_inc_left)
+    with cs.condition(rows[0].is_last) as cs:
+        # rwc_inc_left == 1 for last row in the copy slot
+        cs.constrain_equal(rows[0].rwc_inc_left, rw_diff)
```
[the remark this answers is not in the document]
icemelon: Ah, the comment is wrong here. It should be `rw_diff` because the last row may not be a lookup to the rw table.

src/zkevm_specs/evm/typing.py
```diff
@@ -615,3 +648,141 @@ def _append(
...
+        dst_addr: IntOrFQ,
+        copy_length: IntOrFQ,
+        src_data: Mapping[IntOrFQ, Union[IntOrFQ, Tuple[IntOrFQ, IntOrFQ]]],
+        log_id: IntOrFQ = FQ(0),
+    ):
+        new_rows: List[CopyCircuitRow] = []
+        for i in range(int(copy_length)):
+            if int(src_addr + i) < int(src_addr_boundary):
+                is_pad = False
+                assert src_addr in src_data, f"Cannot find data at the offset {src_addr}"
```
ed255, line 682:
```suggestion
                assert src_addr+i in src_data, f"Cannot find data at the offset {src_addr+i}"
```
otherwise the real offset is not used in the check

icemelon: fixed in a000bf3

src/zkevm_specs/evm/table.py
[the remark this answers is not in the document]
ed255: ping @roynalnaruto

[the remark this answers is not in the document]
roynalnaruto: Sorry for the late response

> when this init code returns, it's the returned memory that defines the bytecode

Yes, this is true, and I hadn't looked into this in depth before adding that pair. Although after reviewing that the fixed pair lookup was redundant, I have removed the entire `CopyPairs` variant. So this should be resolved :)

DreamWuGit:
> Are you referring to the constraint that the value in the write row is the same as the value in the read row (of the same copy step)? If that's the case, I don't see how it's redundant as the CopyCircuit is copying between two different places. Could you elaborate on this duplicate constraint between Copy Circuit and State Circuit; and point to the exact constraint in the python spec?

I check current state circuit don't constrain value read/write consistency,  so it not problem !

specs/tables.md
```diff
@@ -271,3 +271,52 @@ Columns expressions in circuit:
...
+Proved by the copy circuit.
+
+The copy table consists of 13 columns, described as follows:
+
+- **q_step**: a fixed column for boolean value to indicate a copy step, always alternating between 1 and 0, where 1 indicates a read op and 0 indicates a write op.
+- **is_first**: a boolean value to indicate the first row in a copy event.
+- **is_last**: a boolean value to indicate the last row in a copy event.
+- **ID**: could be `$txID`, `$callID`, `$codeHash` (RLC encoded).
+- **Type**: indicates the type of data source, including `Memory`, `Bytecode`, `TxCalldata`, `TxLog`.
+- **Address**: indicates the address in the source data, could be memory address, byte index in the bytecode, tx call data, and tx log data. When the data type is `TxLog`, the address is the combination of byte index, `TxLogFieldTag.Data` tag, and `LogID`.
```
DreamWuGit, line 286: after #214 merged, `Address` for `TxLog` change to `log_id` + `field_tag` + `index` now.  maybe need update accordingly . thanks !

ed255:
> I check current state circuit don't constrain value read/write consistency for now, so it not problem !

But that's just because it's a TODO rigth?  The spec has de read consistency constraint: https://github.com/privacy-scaling-explorations/zkevm-specs/blob/34eafad5f5c1031834d48adc2499a0f94bc3ae3e/src/zkevm_specs/state.py#L490
And the circuit should have it too.

Once the read consistency constraint is added to the state circuit, do you think there will be redundancy of constraints with the copy circuit?

specs/tables.md
icemelon: Yes, I already updated the txlog lookup in the latest commit. could you help take a look?

ed255:
@lispc in case you haven't reviewed this yet; could you take a look please?  So that we reach 2 approvals :)

lispc:
@ed255 @roynalnaruto OK! I missed it! i will have a review now

lispc (approved):

DreamWuGit:
> And the circuit should have it too.
> Once the read consistency constraint is added to the state circuit, do you think there will be redundancy of constraints with the copy circuit?

@ed255  sorry for the delay, yes,  I think so , current circuit don't implement yet.    we can keep an eye on when it starts implementation .

DreamWuGit:
> > I check current state circuit don't constrain value read/write consistency for now, so it not problem !
> 
> But that's just because it's a TODO rigth? The spec has de read consistency constraint:
> 
> https://github.com/privacy-scaling-explorations/zkevm-specs/blob/34eafad5f5c1031834d48adc2499a0f94bc3ae3e/src/zkevm_specs/state.py#L490-L496
> 
> 
> And the circuit should have it too.
> Once the read consistency constraint is added to the state circuit, do you think there will be redundancy of constraints with the copy circuit?


@ed255  sorry for the delay, yes,  I think so , current circuit don't implement yet.    we can keep an eye on when it starts implementation .

PR #197, opened 2022-05-12
Specify the tag usage in bytecode spec

ed255:
A small update to have specs description consistent with the python implementation and circuits.


miha-stopar (approved):
LGTM

PR #198, opened 2022-05-12
[Keccak] misc fixes

ChihChengLiang:
Make these changes to reflect the implementation in https://github.com/appliedzkp/zkevm-circuits/pull/445

1. Fixed the full-pad case. When the next input is a full-pad, the next state_tag must be `Finalize`
2. Fixed an off-by-one error in the is_pad_zone. 
3. Instead of enabling/disabling the whole region with`f_enable`, we enabled the padding regions for all permutation rounds. The benefit is that we can pass input through this region without worrying about branching the input into the padding or non-padding path and merging them.


CPerezz (approved):
LGTM!!

Thanks for these fixes!!

PR #200, opened 2022-05-17
Add MPT table lookup to state circuit

ed255:
Introduce MPT table lookups from the state circuit for Storage and Account operations.

### DIFF
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -351,7 +351,6 @@ class RWTableRow(TableRow):
     value: Expression = field(default=FQ(0))
     value_prev: Expression = field(default=FQ(0))
     aux0: Expression = field(default=FQ(0))
-    aux1: Expression = field(default=FQ(0))
 
 
 @dataclass(frozen=True)
@@ -451,7 +450,6 @@ class Tables:
         value: Expression = None,
         value_prev: Expression = None,
         aux0: Expression = None,
-        aux1: Expression = None,
     ) -> RWTableRow:
         query = {
             "rw_counter": rw_counter,
@@ -464,7 +462,6 @@ class Tables:
             "value": value,
             "value_prev": value_prev,
             "aux0": aux0,
-            "aux1": aux1,
         }
         return lookup(RWTableRow, self.rw_table, query)
 
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -190,9 +190,9 @@ class RWTableTag(IntEnum):
 
 
 class AccountFieldTag(IntEnum):
-    Nonce = 1
-    Balance = 2
-    CodeHash = 4
+    Nonce = auto()
+    Balance = auto()
+    CodeHash = auto()
 
 
 class CallContextFieldTag(IntEnum):
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -144,6 +144,17 @@ class RW(IntEnum):
     Write = 1
 
 
+class MPTTableTag(IntEnum):
+    """
+    Tag for MPTTable lookup
+    """
+
+    Nonce = 1
+    Balance = 2
+    CodeHash = 4
+    Storage = 8
+
+
 class RWTableTag(IntEnum):
     """
     Tag for RWTable lookup, where the RWTable an advice-column table built by
@@ -179,9 +190,9 @@ class RWTableTag(IntEnum):
 
 
 class AccountFieldTag(IntEnum):
-    Nonce = auto()
-    Balance = auto()
-    CodeHash = auto()
+    Nonce = 1
+    Balance = 2
+    CodeHash = 4
 
 
 class CallContextFieldTag(IntEnum):
@@ -343,6 +354,16 @@ class RWTableRow(TableRow):
     aux1: Expression = field(default=FQ(0))
 
 
+@dataclass(frozen=True)
+class MPTTableRow(TableRow):
+    counter: Expression
+    target: Expression  # MPTTableTag
+    address: Expression
+    key: Expression
+    value: Expression
+    value_prev: Expression
+
+
 class Tables:
     """
     A collection of lookup tables used in EVM circuit.
@@ -391,7 +412,7 @@ class Tables:
         self, field_tag: Expression, block_number: Expression = FQ(0)
     ) -> BlockTableRow:
         query = {"field_tag": field_tag, "block_number_or_zero": block_number}
-        return _lookup(BlockTableRow, self.block_table, query)
+        return lookup(BlockTableRow, self.block_table, query)
 
     def tx_lookup(
         self, tx_id: Expression, field_tag: Expression, call_data_index: Expression = FQ(0)
@@ -401,7 +422,7 @@ class Tables:
             "field_tag": field_tag,
             "call_data_index_or_zero": call_data_index,
         }
-        return _lookup(TxTableRow, self.tx_table, query)
+        return lookup(TxTableRow, self.tx_table, query)
 
     def bytecode_lookup(
         self,
@@ -416,7 +437,7 @@ class Tables:
             "index": index,
             "is_code": is_code,
         }
-        return _lookup(BytecodeTableRow, self.bytecode_table, query)
+        return lookup(BytecodeTableRow, self.bytecode_table, query)
 
     def rw_lookup(
         self,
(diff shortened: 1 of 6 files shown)


src/zkevm_specs/state.py
```diff
@@ -400,6 +469,12 @@ def get_keys_compressed_in_order(row: Row) -> List[FQ]:
     if row.is_write == 0 and all_keys_eq(row, row_prev):
         assert row.value == row_prev.value
 
+    # 7. Increment mpt_counter
+    #
+    # When previous row is Storage or Account, increment the mpt_counter by one
+    if row_prev.tag() == Tag.Storage or row_prev.tag() == Tag.Account:
+        assert row.mpt_counter == row_prev.mpt_counter + 1
+
```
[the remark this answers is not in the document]
ed255: I've taken the approach to make `mpt_counter` persistent because unfortunately `Tag.Storage` and `Tag.Account` are not contiguous (there's `Tag.CallContext` in the middle).  So now `mpt_counter` is forced to start at 0, and is forced to increase by 1 on Tag.Storage or Tag.Account, and be the same on other Tags (except for Start).

src/zkevm_specs/state.py
```diff
@@ -182,6 +239,12 @@ def check_storage(row: Row, row_prev: Row):
     if not all_keys_eq(row, row_prev):
         assert row.is_write == 1 and row.rw_counter == 0
 
+    # 2. MPT storage lookup with incremental counter
+    value_prev = row_prev.value if all_keys_eq(row, row_prev) else FQ(0)
```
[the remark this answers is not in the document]
ed255: After thinking about this and following our chat discussion I've made some changes:
1. I noticed that we had a single column for the tx_id in the storage, but we can just use the id column (key1), so I moved it there
2. We already have a `committed_value` column, which must match the previous value before the tx.  If we add tx to the keys, this means that on the first entry for a set of keys, the value_prev that must be used in the MPT lookup is the `committed_value`, so that's the value used in the lookup now in such chase.  This allows us to remove the need to add artificial rows for the first access, because we can already construct such lookup no matter if it's READ or WRITE by using the `commited_value` as `value_prev`.  At the same time, this makes the rules for `rw_counter` simpler!  Previously we were using `rw_counter = 0` for such artificial rows, but there was no constraint forbidding the usage of `rw_counter = 0` elsewhere!  So a malicious prover could insert operations freely.  Now `rw_counter` must be != 0 except for `Tag.Start`.

src/zkevm_specs/state.py
```diff
@@ -222,25 +225,25 @@ def check_stack(row: Row, row_prev: Row):
...
-    #
-    # We add an extra write to set the value of the state in previous block, with rwc=0.
-    #
-    # When the set of all keys changes (first access of storage (address, key))
-    # - It must be a WRITE
-    if not all_keys_eq(row, row_prev):
-        assert row.is_write == 1 and row.rw_counter == 0
+    # 1. When keys don't change, committed_value must be kept equal
+    if all_keys_eq(row, row_prev):
+        assert get_committed_value(row) == get_committed_value(row_prev)
```
han0110, line 235: Since now `tx_id` is also part of keys, when `tx_id` changes, we need to `assert get_committed_value(row) == row_prev.value` to make sure the `committed_value` is updated.

han0110 (approved):
LGTM! Nice work!

ed255:
@z2trillion could you help as a second reviewer for this PR?

src/zkevm_specs/state.py
```diff
@@ -400,6 +480,20 @@ def get_keys_compressed_in_order(row: Row) -> List[FQ]:
...
+    # one, otherwise maintain the same value
+    if row.tag() != Tag.Start:
+        if row.tag() == Tag.Storage or row.tag() == Tag.Account:
+            assert row.mpt_counter == row_prev.mpt_counter + 1
+        else:
+            assert row.mpt_counter == row_prev.mpt_counter
+
+    # 8. RWC !=0 except for Tag.Start
+    if row.tag() != Tag.Start:
+        assert row.rw_counter != 0
```
z2trillion, line 495: the comment in https://github.com/appliedzkp/zkevm-specs/pull/200/files#diff-8c8275ac52d72de465920627c9e900c9ae5d4fc1aba1e75bd44e5d6c346a4d55L209 suggests that originally we were going to allow rw_counter to be 0 for other tags as well. Do we no longer need that because we're doing MPT lookups for every Storage and Account row?

Will we need to get rid of this constraint when we do the optimization mentioned here: https://github.com/appliedzkp/zkevm-specs/pull/200/files#diff-8c8275ac52d72de465920627c9e900c9ae5d4fc1aba1e75bd44e5d6c346a4d55R279?

src/zkevm_specs/evm/table.py
```diff
@@ -179,9 +190,9 @@ def write_with_reversion(self) -> bool:
 
 
 class AccountFieldTag(IntEnum):
-    Nonce = auto()
-    Balance = auto()
-    CodeHash = auto()
+    Nonce = 1
```
z2trillion, line 193: Did you change these to powers of 2 so that it will be easier to select between them once we switch to using binary columns for the tag and field tag?

src/zkevm_specs/state.py
ed255: > the comment in https://github.com/appliedzkp/zkevm-specs/pull/200/files#diff-8c8275ac52d72de465920627c9e900c9ae5d4fc1aba1e75bd44e5d6c346a4d55L209 suggests that originally we were going to allow rw_counter to be 0 for other tags as well. Do we no longer need that because we're doing MPT lookups for every Storage and Account row?

I think originally the only rows (different than Start) where we were thinking of using `rwc = 0` where for Account and Storage to initialize the value (so that we can always get a valid `value_prev`).  A row with `rwc = 0` like that is not tracked by the EVM circuit, and as such an attacker could insert malicious rows anywhere in the State circuit; so we'd need to check that `rwc = 0` can only be enabled in specific scenarios, and otherwise it must be `rwc != 0`.  But then I realized that we already have a `commited_value` column which allows us to always have a valid `value_prev`, so we can remove the need of these Account/Storage rows with `rwc = 0`, and then the rules for `rwc` become simpler.

So removing the Storage/Account rows with `rwc = 0` is unrelated to doing one/many MPT lookup for every Storage/Account.

> Will we need to get rid of this constraint when we do the optimization mentioned here: https://github.com/appliedzkp/zkevm-specs/pull/200/files#diff-8c8275ac52d72de465920627c9e900c9ae5d4fc1aba1e75bd44e5d6c346a4d55R279?

Even with the mentioned optimization we can still keep the same `rwc` rules simplification as in this PR.  The optimization could be implemented by doing the MPT lookup with `(value = value, prev_value = committed_value)` only when the next row keys change (In `committed_value` we have the first read value, and in `value` of the last row for a set of equal keys we have the last write value).

Does this make sense?

src/zkevm_specs/evm/table.py
ed255: Ah, I didn't think about that.  The reason why I changed these to powers of 2 is because currently the MTP circuit is using boolean columns for each one of the possible account fields.  For the lookup expression it looks simpler if those 4 binary columns are merged into a single value, and I thought using powers of 2 would never mix the columns (it's just a binary encoding of the 4 columns).  Nevertheless the MPT circuit probably constraints that at most one of these 4 binary columns is 1, so using `1,2,3,4` would work as well I guess.

z2trillion: Makes sense.

My guess is that we will want these to be 1, 2, 3, 4, ... in a single column the state circuit, but I'm not positive. I'm fine with this change, since it'll be very easy to go back if needed when we add the MPT constraints in the state circuit.

ed255: I've decided to revert this change because after thinking about it, I don't find a compelling reason to change these values, so I'll leave them untouched.

lispc:
@[noel2004](https://github.com/scroll-tech/mpt-circuit/commits?author=noel2004)  have a question on the mpt_counter. In the current code, the mpt_counter increases with keys lexicographic ordering. 
But in the mpt side, the counter should be in wall clock time order?

So we have two option
(1) adjust mpt_counter in state circuit using wallet clock time order(or similar with rw_counter) too. EVM circuit can have constraints like ` if storage_related_rw() { rwc++; mpt_counter++; } else { rwc++}`  We may have to add mpt_counter into rw table? 
(2) let mpt circuit update state root in the same order with current state circuit mpt counter, lexicographic ordering.  It can get the same final root, but it is very counter intuitive.. And not easy to debug.

@ed255

lispc:
ok i understand the design. "we will encounter intermediate state roots that never happened in geth" is what i think as "counter intuitive".   For the "uneasy to debug", i think of bug case where the replay final root is not same with real final root, yeah, it is only bug after it is fixed it will be ok later. No problem~ Thanks for the explanation

PR #201, opened 2022-05-19
Update state circuit spec from impl

### LINKED ISSUE #199: Update state circuit specs based on circuit impl
Based on these notes (Annotated with FIXME are the differences between spec and implementation):
https://github.com/appliedzkp/zkevm-circuits/commit/adcab5ce9d026dbef81c1af3bbd71a3d0a20fb51

ed255:
The main change is that the lexicographic ordering now includes the RWC
as well (so we don't need to check the RWC is monotonically increasing
individually); and that the lexicographic ordering uses a LowerThan
comparison collecting all keys into 16 bit limbs.

Update the spec notes to use two numbers to be more clear.

Use column key names instead of key indexes to make the constraints more
clear (like in the implementation)

Resolve https://github.com/appliedzkp/zkevm-specs/issues/199

### DIFF
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -15,17 +15,19 @@ from .evm import (
     lookup,
 )
 
+MAX_RW_COUNTER = 2**32 - 1
 MAX_MEMORY_ADDRESS = 2**32 - 1
 MAX_KEY_DIFF = 2**32 - 1
 MAX_STACK_PTR = 1023
-MAX_KEY0 = 12  # Number of Tag variants
-MAX_KEY1 = 2**16 - 1  # Maximum number of calls in a block
-MAX_KEY2 = 2**160 - 1  # Ethereum Address size
-MAX_KEY3 = 24  # Max(# of CallContextFieldTag, # of AccountFieldTag) - 1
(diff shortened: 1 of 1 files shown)


ed255:
@lispc could you help review this PR?

lispc:
only one thing to discuss:  in the spec "little endian" is used in limbs sorting. while in the circuit, i am not sure whether it is also "little endian"? from the imple codes, it seems big endian.

lispc:
only one thing to discuss:  in the spec "little endian" is used in limbs sorting. while in the circuit, i am not sure whether it is also "little endian"? from the imple codes, it seems big endian.

lispc (approved):

ed255:
> only one thing to discuss: in the spec "little endian" is used in limbs sorting. while in the circuit, i am not sure whether it is also "little endian"? from the imple codes, it seems big endian.

You're right, the limbs array is implemented in big endian in the circuit and in little endian in this spec; but as Miha says, both are equivalent because it's a matter of how the `lower than` gadget works.  I've added a comment with the list of tags in order used in the lexicographic ordering to make it more explicit which is the order defined in this spec.

miha-stopar (approved):

PR #204, opened 2022-05-20
`MULMOD` opcode

davidnevadoc:
Implementation in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/562


src/zkevm_specs/evm/execution/mulmod.py
```diff
@@ -0,0 +1,66 @@
...
+        r = pushed_r
+
+    # Safety check
+    assert (a.int_value * b.int_value) == d * n.int_value + r.int_value
+
+    # Check (a * b) =  d * n + r
+    a_times_b = RLC((a.int_value * b.int_value) % MOD)
+    left_carry = instruction.mul_add_words(a, b, RLC(0), a_times_b)
+    right_carry = instruction.mul_add_words(RLC(d), n, r, a_times_b)
+    instruction.constrain_equal(left_carry, right_carry)
```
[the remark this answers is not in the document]
davidnevadoc: You're right. I've added a function similar to `mul_add_words` that also ensures the overflow of both sides match. I think this sould solve it.

tests/evm/test_mulmod.py
```diff
@@ -0,0 +1,86 @@
...
+    Block,
+    Bytecode,
+    RWDictionary,
+)
+from zkevm_specs.util import rand_fq, RLC
+
+MAXU256 = (2**256) - 1
+
+
+TESTING_DATA = [
```
adria0, line 18: Try adding `(MAXU256, MAXU256, 1)`...

davidnevadoc: Well spotted!
I think rather than increasing the  range of `k` up to `MAXU256^2` we could do a previous step, reducing one of the factors mod `n`. This way `k` won't overflow:

$$
k = \frac{(a \text{ mod } n) \cdot b }{n } \leq \frac{ (n-1) \cdot b}{n} < b \leq MAXU256 
$$

specs/opcode/09MULMOD.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Check the equality ` k1 * N + a_reduced  == a ` (1)
+- Check the equality ` a_reduced * b = k2 * N + r` (2)
+
+
+```
+TODO: Circuit gadgets description
+```
+
+- Check `r<n` (3)
```
ed255, line 35:
```suggestion
- Check `r<N` (3)
```
To be consistent with the upper case `N` used in this document.

src/zkevm_specs/evm/execution/mulmod.py
```diff
@@ -0,0 +1,75 @@
...
+
+    a_reduced_times_b = a_reduced * b.int_value
+    e = RLC(a_reduced_times_b % MOD)
+    d = RLC(a_reduced_times_b // MOD)
+
+    # Safety check
+    assert (a_reduced_times_b) == k2 * n.int_value + r.int_value
+
+    # Reduction of first factor
+    instruction.mul_add_words(RLC(k1), n, RLC(a_reduced), a)
```
ed255, line 57: Here what we want is:
`a_reduced = a mod n`
and we do it via
`k1 * n + a_reduced = a mod 8**256`

The first thing that comes to my mind is that we should constrain
`a_reduced < n`.  Otherwise there can be many pairs of `(k1, a_reduced)` which are valid, right?

The second thing I'm wondering is: can this multiplicacion (or the addition) overflow in the 256 bits?

davidnevadoc: Yes, `a_reduced < n` constraint is missing. I'll add that!
Regarding the second point, we don't need to worry about overflow in this equation. `a <= MAXU256` so neither side of the equation will overflow.

davidnevadoc: You're right! I'll add this too

adria0:
I'm also fixing the `ADDMOD` with the same strategy in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/212

specs/opcode/09MULMOD.md
```diff
@@ -0,0 +1,64 @@
...
+- Check the equality ` k1 * N + a_reduced  == a ` (1)
+- Check the equality ` a_reduced * b = k2 * N + r` (2)
+
+
+```
+TODO: Circuit gadgets description
+```
+
+- Check `r<N` (3)
+
```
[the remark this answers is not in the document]
adria0: Yep! @davidnevadoc I'll aprove the merge after you updating this 👍

src/zkevm_specs/evm/execution/mulmod.py
```diff
@@ -0,0 +1,73 @@
...
+    assert (a_reduced_times_b) == k * n.int_value + r.int_value
+
+    # Reduction of first factor
+    mod(instruction, a, n, RLC(a_reduced))
+
+    # Reduction of the product
+    instruction.mul_add_words_512(RLC(a_reduced), b, RLC(0), d, e)
+    instruction.mul_add_words_512(RLC(k), n, r, d, e)
+
+    # Check that r<n  and a_reduced<n if n!=0
```
ed255, line 63: I think that in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/204/commits/9949b9d4f71f1768b400c229122908b1b98a5183 you removed the constraint `a_reduced<n if n!=0`

ed255 (changes requested):

src/zkevm_specs/evm/execution/mulmod.py
ed255: Ah sorry, I see that this is now done in the `mod` gadget.  Could you update the comment to remove the part that mentions `a_reduced<n if n!=0` which is not done here now?

ed255 (approved):

src/zkevm_specs/evm/execution/mulmod.py
davidnevadoc: Yep, that shouldn't be there anymore. Done!

tests/evm/test_mulmod.py
davidnevadoc: Included!

PR #205, opened 2022-05-23
code_source rename to code_hash

### LINKED ISSUE #73: Executed bytecode source for opcode lookup

roynalnaruto:
This PR closes #73 . After this is merged, we can proceed with #191


ChihChengLiang (approved):
LGTM

src/zkevm_specs/evm/execution/begin_tx.py
```diff
@@ -58,7 +58,9 @@ def begin_tx(instruction: Instruction):
 
     if tx_is_create == 1:
         # TODO: Verify created address
-        # TODO: Decide what code_source should be (tx_id or hash of creation code)
+        # code_hash represents the contract creation code
+        # 1. In the case of root call, code_hash is the hash of the tx calldata.
+        # 2. In the case of internal call, code_hash is the hash of the caller memory.
```
ed255, line 63:
```suggestion
        # 2. In the case of internal call, code_hash is the hash of the init code within the caller memory.
```
This is a proposal to avoid confusion: It's not the hash of the entire caller memory, just the chunk corresponding to the init code (set by the arguments in the call step)

ed255 (commented):
Nice!  Thanks for addressing this :)  Please take a look at my suggestion; otherwise LGTM!

ed255 (approved):
LGTM!

PR #214, opened 2022-06-02
revise log id for look up

DreamWuGit:
two minor changes:
1. use next step log_id ( current step 's log_id + 1) to look up logs fields.  
  the reason is only in next copy step, log_id plus 1,  while current step log_id is zero,  however, for the semantic matter, they are belong to the same real step of evm/geth,  so both use next step log_id is reasonable for lookup, which also matched pre defined zero log id meaning no log data executed. 
3. add log static gas , which is done in circuit side https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/507


DreamWuGit:
@ed255  could you pls. have a look ? thx !

ed255 (approved):
LGTM! Thanks for the update!

PR #220, opened 2022-06-13
revise log state constrain according to circuit implementation 

DreamWuGit:
in circuit implementation there are some minor changes to raw spec, do update in this PR .  
circuit implementation:  https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/552

### DIFF
--- a/src/zkevm_specs/evm/instruction.py
+++ b/src/zkevm_specs/evm/instruction.py
@@ -500,9 +500,9 @@ class Instruction:
             RW.Write,
             RWTableTag.TxLog,
             key1=tx_id,
-            key2=log_id,
+            key2=FQ(index + int(field_tag) << 32 + log_id.expr().n << 48),
             key3=FQ(field_tag),
-            key4=FQ(index),
+            key4=FQ(0),
         ).value
         return value
 
--- a/src/zkevm_specs/evm/instruction.py
+++ b/src/zkevm_specs/evm/instruction.py
@@ -501,7 +501,7 @@ class Instruction:
             RWTableTag.TxLog,
             key1=tx_id,
             key2=FQ(index + int(field_tag) << 32 + log_id.expr().n << 48),
-            key3=FQ(field_tag),
+            key3=FQ(0),
             key4=FQ(0),
         ).value
         return value
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -376,14 +376,6 @@ def check_tx_log(row: Row, row_prev: Row):
     # tx_id | log_id | field_tag | index | value
     tx_id = row.id()
     prev_tx_id = row_prev.id()
-    log_id = row.address()
-    prev_log_id = row_prev.address()
-    prev_tag = row_prev.tag()
-    tag = row.tag()
-    prev_field_tag = row_prev.field_tag()
-    field_tag = row.field_tag()
-    index = row.storage_key()
-    prev_index = row_prev.storage_key()
 
     # 12.0 is_write is always true
     assert row.is_write == 1
@@ -391,32 +383,8 @@ def check_tx_log(row: Row, row_prev: Row):
     if row.tag() == row_prev.tag():
         if tx_id != prev_tx_id:
             assert tx_id == prev_tx_id + 1
-            # in RW table, log_id starts with 1 in tx, zero log_id means no log step executed,
-            # hence no log entry inserted into rw table
-            assert log_id == 1
-            # first field_tag is Address when tx changes
-            assert row.field_tag() == TxLogFieldTag.Address
-        else:
-            if log_id == prev_log_id and prev_field_tag == U256(TxLogFieldTag.Address):
-                assert (field_tag - prev_field_tag).n > 0
-
-            # increase log_id when field tag changes to Address within same tx
-            if field_tag == U256(TxLogFieldTag.Address):
-                assert log_id == prev_log_id + 1
-
-            # within same tx, log_id will not change if field_tag != Address
-            if field_tag != U256(TxLogFieldTag.Address):
-                assert log_id == prev_log_id
-            # index can only increase by one when field tag stays same.
-            if prev_field_tag == field_tag:
-                assert index == prev_index + 1
-
-    # if tag Topic appear, topic_index needs to be in range [0,4)
-    if field_tag == U256(TxLogFieldTag.Topic):
-        assert_in_range(index, 0, 3)
-    # index is zero for Address
-    if field_tag == U256(TxLogFieldTag.Address):
-        assert index == 0
+            # commented  field_tag-specific constraints as issue
+            # https://github.com/privacy-scaling-explorations/zkevm-specs/issues/221
 
 
 @is_circuit_code
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -399,9 +399,6 @@ def check_tx_log(row: Row, row_prev: Row):
         else:
             if log_id == prev_log_id and prev_field_tag == U256(TxLogFieldTag.Address):
                 assert (field_tag - prev_field_tag).n > 0
-            # index is zero for address
-            if field_tag == U256(TxLogFieldTag.Address):
-                assert index == 0
 
             # increase log_id when field tag changes to Address within same tx
             if field_tag == U256(TxLogFieldTag.Address):
@@ -410,14 +407,16 @@ def check_tx_log(row: Row, row_prev: Row):
             # within same tx, log_id will not change if field_tag != Address
             if field_tag != U256(TxLogFieldTag.Address):
                 assert log_id == prev_log_id
-            # if tag Data appear, data_index can only increase by one when tag stays same.
-            # if tag Topic appear, topic_index needs to be in range [0,4) and it can only increase by one
-            # when tag stays same.
-            if field_tag == U256(TxLogFieldTag.Topic):
-                assert_in_range(index, 0, 3)
-
+            # index can only increase by one when field tag stays same.
             if prev_field_tag == field_tag:
                 assert index == prev_index + 1
+
+    # if tag Topic appear, topic_index needs to be in range [0,4)
+    if field_tag == U256(TxLogFieldTag.Topic):
+        assert_in_range(index, 0, 3)
+    # index is zero for Address
+    if field_tag == U256(TxLogFieldTag.Address):
+        assert index == 0
 
 
 @is_circuit_code
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -379,12 +379,9 @@ def check_tx_log(row: Row, row_prev: Row):
 
     # 12.0 is_write is always true
     assert row.is_write == 1
-    # 12.1 tx_id change must be increasing
-    if row.tag() == row_prev.tag():
-        if tx_id != prev_tx_id:
-            assert tx_id == prev_tx_id + 1
-            # removed field_tag-specific constraints as issue
-            # https://github.com/privacy-scaling-explorations/zkevm-specs/issues/221
+
+    # removed field_tag-specific constraints as issue
+    # https://github.com/privacy-scaling-explorations/zkevm-specs/issues/221
 
 
 @is_circuit_code
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -379,11 +379,11 @@ def check_tx_log(row: Row, row_prev: Row):
 
     # 12.0 is_write is always true
     assert row.is_write == 1
-    # 12.1 reset log_id when tx_id increases
+    # 12.1 tx_id change must be increasing
     if row.tag() == row_prev.tag():
         if tx_id != prev_tx_id:
             assert tx_id == prev_tx_id + 1
-            # commented  field_tag-specific constraints as issue
+            # removed field_tag-specific constraints as issue
             # https://github.com/privacy-scaling-explorations/zkevm-specs/issues/221
 
 
--- a/src/zkevm_specs/state.py
+++ b/src/zkevm_specs/state.py
@@ -391,7 +391,8 @@ def check_tx_log(row: Row, row_prev: Row):
     if row.tag() == row_prev.tag():
         if tx_id != pre_tx_id:
             assert tx_id == pre_tx_id + 1
-            # in RW table, log_id starts with 1
+            # in RW table, log_id starts with 1 per tx, zero log_id means no any log step executed,
+            # hence no any log entry inserts into rw table.
             assert log_id == 1
             # first field_tag is Address when tx changes
             assert row.field_tag() == TxLogFieldTag.Address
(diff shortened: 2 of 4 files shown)


DreamWuGit:
@ed255  Could you pls. have a look ? Thx !

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,33 @@ def check_tx_log(row: Row, row_prev: Row):
     if row.tag() == row_prev.tag():
         if tx_id != pre_tx_id:
             assert tx_id == pre_tx_id + 1
-            assert log_id == 0
+            # in RW table, log_id starts with 1
+            assert log_id == 1
+            # first field_tag is Address when tx changes
```
miha-stopar, line 396: Could you add some comment why this need to be so?

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,33 @@ def check_tx_log(row: Row, row_prev: Row):
...
+            assert row.field_tag() == TxLogFieldTag.Address
         else:
             # increase log_id when tag changes to Address, make sure tag can only increase(Non-Decreasing),
```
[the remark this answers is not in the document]
DreamWuGit: these lines handle tx changes cases, when enter new tx, log_id resets to 1 and field tag to `Address`, line 409 handles within tx, log_id increases when field tag change to `Address`, which happens multi log step in one tx .

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,33 @@ def check_tx_log(row: Row, row_prev: Row):
...
+
+            # increase log_id when field tag changes to Address within same tx
+            if tag == prev_tag and field_tag == U256(TxLogFieldTag.Address):
+                assert log_id == prev_log_id + 1
+
+            # within same tx, log_id will not change if field_tag != Address
+            if tag == prev_tag and field_tag != U256(TxLogFieldTag.Address):
+                assert log_id == prev_log_id
+            # if tag Data appear, data_index can only increase by one when tag stays same.
+            # if tag Topic appear, topic_index in range [0,4),can only increase by one when tag stays same.
```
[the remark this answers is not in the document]
DreamWuGit: updated in [75fb5f7](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/75fb5f730d5659ddfb7c40b8c3cc8256857efac9) thanks !

src/zkevm_specs/state.py
DreamWuGit: added in [75fb5f7](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/75fb5f730d5659ddfb7c40b8c3cc8256857efac9) thanks !!

src/zkevm_specs/state.py
```diff
@@ -377,6 +377,9 @@ def check_tx_log(row: Row, row_prev: Row):
     tx_id = row.id()
     pre_tx_id = row_prev.id()
     log_id = row.address()
+    prev_log_id = row_prev.address()
+    prev_tag = row_prev.tag()
+    tag = row.tag()
     pre_field_tag = row_prev.field_tag()
```
ed255, line 383:
```suggestion
    prev_tx_id = row_prev.id()
    log_id = row.address()
    prev_log_id = row_prev.address()
    prev_tag = row_prev.tag()
    tag = row.tag()
    prev_field_tag = row_prev.field_tag()
```

Use `prev_` everywhere for consistency.

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,35 @@ def check_tx_log(row: Row, row_prev: Row):
     if row.tag() == row_prev.tag():
         if tx_id != pre_tx_id:
             assert tx_id == pre_tx_id + 1
-            assert log_id == 0
+            # in RW table, log_id starts with 1 per tx, zero log_id means no any log step executed,
```
ed255, line 394: I have the feeling that we may have redundant constraints in the state circuit; which are checking things that are already enforced by the EVM circuit.
For example, the EVM circuit is constrained to generate lookups for TxLog with correct `log_id`.  So if the EVM circuit passes the constraints, then there's no way a TxLog can exist that doesn't start with `1`.

I think for now it's OK to have some redundant constraints, but in the future I think we need to figure out a way to analyze this situation and remove redundant constraints (with confidence that we're not allowing invalid witnesses to pass)

src/zkevm_specs/state.py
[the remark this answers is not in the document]
DreamWuGit: I prefer to add condition when log_id stays the same,  there are can be without topic (log0 with no topic) case.

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,35 @@ def check_tx_log(row: Row, row_prev: Row):
...
+                assert index == 0
+
+            # increase log_id when field tag changes to Address within same tx
+            if tag == prev_tag and field_tag == U256(TxLogFieldTag.Address):
+                assert log_id == prev_log_id + 1
+
+            # within same tx, log_id will not change if field_tag != Address
+            if tag == prev_tag and field_tag != U256(TxLogFieldTag.Address):
+                assert log_id == prev_log_id
+            # if tag Data appear, data_index can only increase by one when tag stays same.
```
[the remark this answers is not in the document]
DreamWuGit: as we talked possible redundant constraint above, I am not willing to add this in state circuit side as evm circuit has been constrained because they use look up from zero index .

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,35 @@ def check_tx_log(row: Row, row_prev: Row):
...
+            # index is zero for address
+            if field_tag == U256(TxLogFieldTag.Address):
+                assert index == 0
+
+            # increase log_id when field tag changes to Address within same tx
+            if tag == prev_tag and field_tag == U256(TxLogFieldTag.Address):
+                assert log_id == prev_log_id + 1
+
+            # within same tx, log_id will not change if field_tag != Address
+            if tag == prev_tag and field_tag != U256(TxLogFieldTag.Address):
```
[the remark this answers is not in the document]
DreamWuGit: updated in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/36e299bc2531d1c540391cef04420c9eeda4ea28 thanks .

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,35 @@ def check_tx_log(row: Row, row_prev: Row):
...
             if pre_field_tag == U256(TxLogFieldTag.Address):
                 assert (field_tag - pre_field_tag).n > 0
-            # make sure if tag Data appear, data_index can only increase by one when tag stays same.
-            # make sure if tag Topic appear, topic_index in range [0,4),can only increase by one when tag stays same.
+            # index is zero for address
+            if field_tag == U256(TxLogFieldTag.Address):
+                assert index == 0
+
+            # increase log_id when field tag changes to Address within same tx
+            if tag == prev_tag and field_tag == U256(TxLogFieldTag.Address):
```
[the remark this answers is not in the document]
DreamWuGit: removed in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/36e299bc2531d1c540391cef04420c9eeda4ea28 thanks .

src/zkevm_specs/state.py
```diff
@@ -388,23 +391,35 @@ def check_tx_log(row: Row, row_prev: Row):
     if row.tag() == row_prev.tag():
         if tx_id != pre_tx_id:
             assert tx_id == pre_tx_id + 1
-            assert log_id == 0
+            # in RW table, log_id starts with 1 per tx, zero log_id means no any log step executed,
+            # hence no any log entry inserts into rw table.
```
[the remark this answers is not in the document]
DreamWuGit: updated in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/36e299bc2531d1c540391cef04420c9eeda4ea28 thanks .

src/zkevm_specs/state.py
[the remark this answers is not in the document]
DreamWuGit: add  log_id stays the same condition in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/36e299bc2531d1c540391cef04420c9eeda4ea28 thanks .

src/zkevm_specs/state.py
```diff
@@ -375,36 +375,49 @@ def check_account_destructed(row: Row, row_prev: Row):
 def check_tx_log(row: Row, row_prev: Row):
     # tx_id | log_id | field_tag | index | value
     tx_id = row.id()
-    pre_tx_id = row_prev.id()
+    prev_tx_id = row_prev.id()
     log_id = row.address()
-    pre_field_tag = row_prev.field_tag()
+    prev_log_id = row_prev.address()
```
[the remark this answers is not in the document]
DreamWuGit: yes, the spec don't update to match circuit side for log id lookup related, because  current circuit side implementation have shortage regarding (log_id + index ) as address.   I have created an issue #224 to describe it,  have to update it in future anyway.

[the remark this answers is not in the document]
ed255: I would suggest resolving https://github.com/privacy-scaling-explorations/zkevm-specs/pull/227 , and once it's merged, update this PR (if necessary), and then from my side this PR will look good :)
Let me know if you'd like to proceed in a different way.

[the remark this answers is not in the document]
DreamWuGit: I would rebase https://github.com/privacy-scaling-explorations/zkevm-specs/pull/227 into this PR as it is simple .

ed255:
Before approving this I would like to discuss the topic of redundant constraints; in particular the field_tag-specific constraints. This is also related to https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/552

Maybe it's worth removing the redundant constraints now instead of doing it later, mainly so that the state circuit implementation doesn't grow in columns and/or degree.  I'm talking specifically about the field_tag-specific constraints, which require using a selector expression.  I discussed this topic with @z2trillion via chat and would like to continue it here.

DreamWuGit:
> Before approving this I would like to discuss the topic of redundant constraints; in particular the field_tag-specific constraints. This is also related to [privacy-scaling-explorations/zkevm-circuits#552](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/552)
> 
> Maybe it's worth removing the redundant constraints now instead of doing it later, mainly so that the state circuit implementation doesn't grow in columns and/or degree. I'm talking specifically about the field_tag-specific constraints, which require using a selector expression. I discussed this topic with @z2trillion via chat and would like to continue it here.

I am OK with it because it indeed increases much complexity,  I will comment them with that issue #221

DreamWuGit:
@ed255  removed  field_tag-specific constraints and test  in  [c487d5c](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/c487d5c41025b97df5a3987509a77986e16a167a) & [72c84a0](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/72c84a06dee6af5ee4749b6908b6ec3d4d889694)

src/zkevm_specs/state.py
```diff
@@ -375,36 +375,16 @@ def check_account_destructed(row: Row, row_prev: Row):
...
-    pre_field_tag = row_prev.field_tag()
-    field_tag = row.field_tag()
-    index = row.storage_key()
-    pre_index = row_prev.storage_key()
+    prev_tx_id = row_prev.id()
 
     # 12.0 is_write is always true
     assert row.is_write == 1
-    # 12.1 reset log_id when tx_id increases
+    # 12.1 tx_id change must be increasing
```
z2trillion, line 382: I think "tx_id change is either 0 or 1" would be more accurate.

src/zkevm_specs/evm/instruction.py
```diff
@@ -500,9 +500,9 @@ def tx_log_lookup(
             RW.Write,
             RWTableTag.TxLog,
             key1=tx_id,
-            key2=log_id,
+            key2=FQ(index + int(field_tag) << 32 + log_id.expr().n << 48),
             key3=FQ(field_tag),
```
z2trillion, line 504: Is it possible to remove the field_tag from this key, since it already appears in key2?

src/zkevm_specs/state.py
DreamWuGit: since state circuit sorts by `id` key , I think it is should be increasing by default, otherwise lexicographic ordering will fail,  right ?   
if so , this check can also be removed here, generally we can skip all key increasing check in all type Ops state constraints!

src/zkevm_specs/evm/instruction.py
DreamWuGit: I don't see no bad effect for remaining it so far :)

z2trillion: It doesn't make sense for it to appear twice. Does anything use the fact that `key3=FQ(field_tag)` rather than `key3=FQ(0)`?

src/zkevm_specs/state.py
z2trillion: It's true that the lexicographic ordering chip forces the id to be increasing here, but actually it's the evm circuit that imposes the stronger constraint that the change must be either 0 or 1.

src/zkevm_specs/evm/instruction.py
DreamWuGit: @z2trillion  sound good to me ,  Will try to remove it.

src/zkevm_specs/state.py
DreamWuGit: evm circuit has no constrain for specific log operation.   I just realize that tx id change can be any value , because some tx may don't contain log* step, so the tx_id is not continuous  in the rw table. such as tx_1, tx_2...tx_5, only tx_1 and tx_5 contain log step.  we have table like this: 

| rw_counter  | tx_id      | log_id |  ..    |
| -----     | ---- | ----- | --- |
| 1   |  1      |   1       | ..  |
| 2  |   5     |     1        | ..  |

so I think we can remove the tx_id constrain since lexicographic ordering force it to be increasing .  
@ed255  WDYT?

src/zkevm_specs/evm/instruction.py
DreamWuGit: udpated in [9187833](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/91878339cd1b0bca2632c7af3a2d4deb2b2eb87b) thanks !

src/zkevm_specs/state.py
DreamWuGit: removed tx id constraint in [b7d16ff](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/220/commits/b7d16fff6027d8a70ca203c92c3afe5974fe07eb)

ed255: > evm circuit has no constrain for specific log operation. I just realize that tx id change can be any value , because some tx may don't contain log* step, so the tx_id is not continuous in the rw table. such as tx_1, tx_2...tx_5, only tx_1 and tx_5 contain log step. we have table like this:
> rw_counter 	tx_id 	log_id 	..
> 1 	1 	1 	..
> 2 	5 	1 	..
> 
> so I think we can remove the tx_id constraint since lexicographic ordering force it to be increasing . @ed255 WDYT?

I think you're right!  The constraint for `tx_id` increasing by `+1` will not always be satisfied, so it should be removed.

icemelon (approved):
LGTM

Issue #224, opened 2022-06-16
state circuit : fix( log_id +index) represents tx log address 


DreamWuGit:
current state circuit  make `address`  by (`index` + `log_id`) for `TxLog` Op , but somewhat different meaning from other types (`Memory`,  `Stack`, `Account`),  for those types, `address`  can identify each item,  i.e. for stack , `stack_ponter` as address can represent certain stack item .    but only by `index` + `log_id` is not adequate to identify each log item, because `index` can be the same with different field tag(`Topic`, `Data`), I think we either  use ( index + field_tag + log_id)  or follow raw layout, `log_id`  as the address, regard `index`  as the fourth key, like `storage_key`.    

@z2trillion  talked with you before, WDYT now?

ed255:
I've reviewed https://github.com/privacy-scaling-explorations/zkevm-specs/pull/227 and I like the approach of setting `address = index || field_tag || log_id` where each element is 16 bits.

DreamWuGit:
closing #214 merged !!

PR #227, opened 2022-06-21
fix issue #224 : add  field_tag into address representation

DreamWuGit:
Try to resolve issue #224

### DIFF
--- a/src/zkevm_specs/evm/instruction.py
+++ b/src/zkevm_specs/evm/instruction.py
@@ -500,7 +500,7 @@ class Instruction:
             RW.Write,
             RWTableTag.TxLog,
             key1=tx_id,
-            key2=FQ(index + int(field_tag) << 16 + log_id.expr().n << 32),
+            key2=FQ(index + int(field_tag) << 32 + log_id.expr().n << 48),
             key3=FQ(field_tag),
             key4=FQ(0),
         ).value
--- a/src/zkevm_specs/evm/instruction.py
+++ b/src/zkevm_specs/evm/instruction.py
@@ -500,9 +500,9 @@ class Instruction:
             RW.Write,
             RWTableTag.TxLog,
             key1=tx_id,
-            key2=log_id,
+            key2=FQ(index + int(field_tag) << 16 + log_id.expr().n << 32),
             key3=FQ(field_tag),
-            key4=FQ(index),
+            key4=FQ(0),
         ).value
         return value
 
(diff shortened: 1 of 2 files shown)


ed255 (approved):
LGTM!

I think this is a nice solution for the issue :).  Now the order of fields of a TxLog is set so that it can follow the  lexicographic ordering without scrambling rows, because following this approach `field_tag` will have higher precedence than `index` (which was not the case if `address = log_id || index`).  And since the address is also represented as 16 bit limbs columns in the state circuit, we can still do checks on `log_id`, `field_tag` and `index` individually without any penalty.

z2trillion:
Isn't the ordering unchanged with this packing? Previously it was: 
```
(tx_id, log_id, field_tag, index), 
```
and with this change it will be 
```
(tx_id, log_id, field_tag, index, field_tag), 
```
which is equivalent?

(This is all in a world where https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/566 gets merged in an restores the sort order of the keys to be key1, key2, key3, key4).

ed255:
> Isn't the ordering unchanged with this packing? Previously it was:
> 
> ```
> (tx_id, log_id, field_tag, index), 
> ```
> 
> and with this change it will be
> 
> ```
> (tx_id, log_id, field_tag, index, field_tag), 
> ```
> 
> which is equivalent?

This is correct, from the spec point of view.  But in between we currently have this in the circuits:
```
(tx_id, log_id, index, field_tag)
```
See https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2aa9875d9bcf5dd8937af7d88f121dad1a8877ed/zkevm-circuits/src/evm_circuit/witness.rs#L766

AFAIK the aim here was to remove the `index` from the key4 (storageKey) which is RLC encoded (and also expressed as 8 bit limbs) and put it somewhere else that allows simpler raw access: the key2 (address) which is also expressed as 16 bit limbs (which is convenient because `log_id`, `field_tag` `and `index` fit in 16 bits)

> (This is all in a world where [privacy-scaling-explorations/zkevm-circuits#566](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/566) gets merged in and restores the sort order of the keys to be key1, key2, key3, key4).

NOTE: I haven't taken a look yet at that PR.

DreamWuGit:
I merged this PR change into #220 , closing this,  any concern can discuss there if you have.

PR #235, opened 2022-07-22
SHA3 opcode | Keccak support in copy circuit

### LINKED ISSUE #223: SHA3 opcode
Specify the execution state for the SHA3 opcode.

This opcode uses dynamic length memory copy so it may depend on the Copy Circuit.

### LINKED ISSUE #228: Add keccak support in the copy circuit
Currently the Copy Circuit spec doesn't support Keccak from an EVM dynamic memory chunk.

See https://github.com/privacy-scaling-explorations/zkevm-specs/pull/194#discussion_r900128743

roynalnaruto:
This PR adds a `RlcAcc` type in the `CopyDataTypeTag` enum.

This tag is reserved for copy destination where instead of copying byte to byte (read to write), we're accumulating an RLC value in the destination. That is:
```
write_val::cur == (write_val::prev * r) + read_val::cur
```

This variant of the copy circuit will be used for `SHA3` opcode execution gadget, where the goal is to find a final accumulator value over each byte from the input. The final RLC accumulator will be used to lookup to the Keccak table and find the `SHA3` of the input bytes.

The PR also implements the `SHA3` opcode.

Closes #228 
Closes #223

### DIFF
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -419,7 +419,7 @@ class CopyTableRow(TableRow):
     src_addr_end: FQ
     dst_addr: FQ
     length: FQ
-    value_rlc: FQ
+    rlc_acc: FQ
     rw_counter: FQ
     rwc_inc: FQ
 
@@ -485,7 +485,7 @@ class Tables:
                         src_addr_end=first_row.src_addr_end,
                         dst_addr=next_row.addr,
                         length=first_row.bytes_left,
-                        value_rlc=row.rlc_acc,
+                        rlc_acc=row.rlc_acc,
                         rw_counter=first_row.rw_counter,
                         rwc_inc=first_row.rwc_inc_left,
                     )
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -397,6 +397,7 @@ class CopyCircuitRow(TableRow):
     src_addr_end: FQ
     bytes_left: FQ
     value: FQ
+    rlc_acc: FQ
     is_code: FQ
     is_pad: FQ
     rw_counter: FQ
@@ -474,10 +475,6 @@ class Tables:
                 assert i + 1 < len(copy_circuit), "Not enough rows in copy circuit"
                 next_row = copy_circuit[i + 1]
                 assert next_row.q_step == 0, "Invalid copy circuit"
-
-            # update `value_rlc` when we encounter the copy event's last row.
-            if row.is_last == 1:
-                value_rlc = row.value
                 rows.append(
                     CopyTableRow(
                         src_id=first_row.id,
@@ -488,12 +485,11 @@ class Tables:
                         src_addr_end=first_row.src_addr_end,
                         dst_addr=next_row.addr,
                         length=first_row.bytes_left,
-                        value_rlc=value_rlc,
+                        value_rlc=row.rlc_acc,
                         rw_counter=first_row.rw_counter,
                         rwc_inc=first_row.rwc_inc_left,
                     )
                 )
-
         return set(rows)
 
     def fixed_lookup(
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -418,8 +418,16 @@ class CopyTableRow(TableRow):
     src_addr_end: FQ
     dst_addr: FQ
     length: FQ
+    value_rlc: FQ
     rw_counter: FQ
     rwc_inc: FQ
+
+
+@dataclass(frozen=True)
+class KeccakTableRow(TableRow):
+    idx: FQ
+    hash_rlc: FQ
+    value_rlc: FQ
 
 
 class Tables:
@@ -433,6 +441,7 @@ class Tables:
     bytecode_table: Set[BytecodeTableRow]
     rw_table: Set[RWTableRow]
     copy_table: Set[CopyTableRow]
+    keccak_table: Set[KeccakTableRow]
 
     def __init__(
         self,
@@ -441,6 +450,7 @@ class Tables:
         bytecode_table: Set[BytecodeTableRow],
         rw_table: Union[Set[Sequence[Expression]], Set[RWTableRow]],
         copy_circuit: Sequence[CopyCircuitRow] = None,
+        keccak_table: Sequence[KeccakTableRow] = None,
     ) -> None:
         self.block_table = block_table
         self.tx_table = tx_table
@@ -451,29 +461,38 @@ class Tables:
         )
         if copy_circuit is not None:
             self.copy_table = self._convert_copy_circuit_to_table(copy_circuit)
+        if keccak_table is not None:
+            self.keccak_table = set(keccak_table)
 
     def _convert_copy_circuit_to_table(self, copy_circuit: Sequence[CopyCircuitRow]):
-        rows = []
+        rows: List[CopyTableRow] = []
         for i, row in enumerate(copy_circuit):
-            if row.is_first != 1:
-                continue
-            assert i + 1 < len(copy_circuit), "Not enough rows in copy circuit"
-            next_row = copy_circuit[i + 1]
-            assert next_row.q_step == 0, "Invalid copy circuit"
-            rows.append(
-                CopyTableRow(
-                    src_id=row.id,
-                    src_type=row.tag,
-                    dst_id=next_row.id,
-                    dst_type=next_row.tag,
-                    src_addr=row.addr,
-                    src_addr_end=row.src_addr_end,
-                    dst_addr=next_row.addr,
-                    length=row.bytes_left,
-                    rw_counter=row.rw_counter,
-                    rwc_inc=row.rwc_inc_left,
+            # the first row and the row next to it will be used for its fields.
+            if row.is_first == 1:
+                first_row = row
+                assert i + 1 < len(copy_circuit), "Not enough rows in copy circuit"
+                next_row = copy_circuit[i + 1]
+                assert next_row.q_step == 0, "Invalid copy circuit"
+
+            # update `value_rlc` when we encounter the copy event's last row.
+            if row.is_last == 1:
+                value_rlc = row.value
+                rows.append(
+                    CopyTableRow(
+                        src_id=first_row.id,
+                        src_type=first_row.tag,
+                        dst_id=next_row.id,
+                        dst_type=next_row.tag,
+                        src_addr=first_row.addr,
+                        src_addr_end=first_row.src_addr_end,
+                        dst_addr=next_row.addr,
+                        length=first_row.bytes_left,
+                        value_rlc=value_rlc,
+                        rw_counter=first_row.rw_counter,
+                        rwc_inc=first_row.rwc_inc_left,
+                    )
                 )
-            )
+
         return set(rows)
 
     def fixed_lookup(
@@ -580,6 +599,13 @@ class Tables:
             "rw_counter": rw_counter,
         }
         return lookup(CopyTableRow, self.copy_table, query)
+
+    def keccak_lookup(self, length: Expression, value_rlc: Expression):
+        query = {
+            "idx": length - FQ(1),
+            "value_rlc": value_rlc,
+        }
+        return lookup(KeccakTableRow, self.keccak_table, query)
 
 
 T = TypeVar("T", bound=TableRow)
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -425,9 +425,10 @@ class CopyTableRow(TableRow):
 
 @dataclass(frozen=True)
 class KeccakTableRow(TableRow):
-    idx: FQ
-    hash_rlc: FQ
-    value_rlc: FQ
+    state_tag: FQ
+    input_len: FQ
+    acc_input: FQ
+    output: FQ
 
 
 class Tables:
@@ -602,8 +603,9 @@ class Tables:
 
     def keccak_lookup(self, length: Expression, value_rlc: Expression):
         query = {
-            "idx": length - FQ(1),
-            "value_rlc": value_rlc,
+            "state_tag": FQ(2),  # Finalize
+            "input_len": length,
+            "acc_input": value_rlc,
         }
         return lookup(KeccakTableRow, self.keccak_table, query)
 
--- a/src/zkevm_specs/evm/table.py
+++ b/src/zkevm_specs/evm/table.py
@@ -289,6 +289,15 @@ class CopyDataTypeTag(IntEnum):
     TxCalldata = auto()
     TxLog = auto()
 
+    # RLC Accumulator tag can be used whenever we wish to
+    # accumulates `value` iteratively over all the steps in
+    # copy circuit. This is specifically used in the SHA3
+    # opcode execution where the copy table's last row has
+    # an accumulated value that is the RLC representation of
+    # all input bytes. Using this value, we can then lookup
+    # the Keccak table for the SHA3 of the input bytes.
+    RlcAcc = auto()
+
 
 class WrongQueryKey(Exception):
     def __init__(self, table_name: str, diff: Set[str]) -> None:
@@ -396,6 +405,7 @@ class CopyCircuitRow(TableRow):
     is_bytecode: FQ
     is_tx_calldata: FQ
     is_tx_log: FQ
+    is_rlc_acc: FQ
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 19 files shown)


ed255:
> This tag is reserved for copy destination where instead of copying byte to byte (read to write), we're accumulating an RLC value in the destination. That is:
> 
> ```
> write_val::cur == write_val::prev + read_val::cur
> ```

This is interesting.  While working on the Super Circuit I noticed that for RLC with dynamic-length inputs, it's easier (or I should say, more convenient) to calculate the `RLC(reversed(input))`.
This is assuming this definition of RLC:
`RLC(v) = v[0] + r * v[1] + r^2 * v[2] + r^3 * v[3] + ... + r^n * v[n]`
And this is because it's easy to lay out the inputs in order in a circuit, and then calculate an RLC forward (meaning that you get the RLC at the bottom row) like this:
`acc_rlc[i] = acc_rlc[i-1] * r + v[i]`
which makes `acc_rlc[n-1] = RLC(reversed(v))`

The bytecode circuit already does the lookups to keccak like this (using the reverse of the input).

Maybe we can agree to always use `RLC(reversed(input))` when the input length is dynamic.

roynalnaruto:
> This is assuming this definition of RLC:
> `RLC(v) = v[0] + r * v[1] + r^2 * v[2] + r^3 * v[3] + ... + r^n * v[n]`

I did know about this definition from the bytecode circuit. I made a mistake by only adding up every `value` while ignoring the `r^i` multipliers.

Based on what I've pushed in the recent fix, it turns out to be this, which is in fact the `RLC(reverse(input))`
```
rlc_acc(i) = v[i] + v[i-1]*r + v[i-2]*r^2 + ... + v[0]*r^i
```

ed255:
Before this feature, a copy_table lookup looked like this: 
```
                is_first: 1,
                src_id,
                src_tag,
                dst_id,
                dst_tag,
                src_addr,
                src_addr_end,
                dst_addr,
                length,
                rw_counter,
                rwc_inc,
```

1. How will the lookup table look like after this feature?
2. Which values will be used for an AccRlc lookup?  The accumulated rlc is found in the row with `is_last = 1`, but the input arguments (src_addr, rw_counter) are found in the row with `is_first = 1`.

roynalnaruto:
@ed255 I've handled the above points in this PR: https://github.com/scroll-tech/zkevm-specs/pull/46

I haven't opened it yet against upstream because I was waiting for this PR to be merged first. Do you prefer if I opened a single PR for this change as well as implementing the SHA3 opcode?

ed255:
> @ed255 I've handled the above points in this PR: [scroll-tech#46](https://github.com/scroll-tech/zkevm-specs/pull/46)
> 
> I haven't opened it yet against upstream because I was waiting for this PR to be merged first. Do you prefer if I opened a single PR for this change as well as implementing the SHA3 opcode?

Ah I see.  Considering that this PR is quite small I would say that I'd prefer if the SHA3 opcode spec was added to this PR as well :)  That is, merging https://github.com/scroll-tech/zkevm-specs/pull/46 into this PR.  I think this way the review will be more clear, because the same PR will contain the new copy circuit functionality, and the SHA3 opcode using such functionality, with the corresponding tests.

> > How will the lookup table look like after this feature?
> 
> Defined here: https://github.com/scroll-tech/zkevm-specs/blob/feat/sha3/src/zkevm_specs/evm/table.py#L411-L423
> 
> > Which values will be used for an AccRlc lookup? ...
> 
> Handled here: https://github.com/scroll-tech/zkevm-specs/blob/feat/sha3/src/zkevm_specs/evm/table.py#L467-L496
> 
> In the PR: [scroll-tech#46](https://github.com/scroll-tech/zkevm-specs/pull/46) I already have a basic test case for `SHA3` opcode execution, for `is_root=True` transaction, with the `SHA3` implemented here: https://github.com/scroll-tech/zkevm-specs/blob/feat/sha3/src/zkevm_specs/evm/execution/sha3.py

Thanks for the pointers!

roynalnaruto:
@ed255 Merged! So this PR now is a full implementation of the SHA3 opcode.

src/zkevm_specs/evm/table.py
```diff
@@ -408,10 +418,18 @@ class CopyTableRow(TableRow):
...
+    value_rlc: FQ
     rw_counter: FQ
     rwc_inc: FQ
 
 
+@dataclass(frozen=True)
+class KeccakTableRow(TableRow):
+    idx: FQ
+    hash_rlc: FQ
+    value_rlc: FQ
```
ed255, line 430: From my understanding, the keccak lookup has the following columns:
- is_enabled (should be 1 for a lookup)
- input_rlc
- length
- hash_rlc

Instead of `length` you defined `idx`.  Why is that?  I think this type should only use the column meanings used for lookup (without paying attention to the column meaning in the keccak table intermediate rows not used for lookup)

ed255 (commented):
I've left comments to discuss some stuff.

src/zkevm_specs/evm/typing.py
```diff
@@ -666,19 +667,43 @@ def _append(
...
+
+    def add(self, data: bytes, r: FQ) -> KeccakCircuit:
+        rows: List[KeccakTableRow] = []
+        hash_rlc = RLC(keccak256(data), r, n_bytes=32)
+        value_rlc = FQ.zero()
+        for i in range(len(data)):
+            value_rlc = value_rlc * r + data[i]
+            rows.append(
+                KeccakTableRow(
+                    idx=FQ(i),
```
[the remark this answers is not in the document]
roynalnaruto: My bad, I hadn't checked https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/keccak256/variable-length-input.md#lookup before.

I will make appropriate changes to the `KeccakTableRow`

src/zkevm_specs/evm/table.py
roynalnaruto: I hadn't seen these docs before. But yes, that helps. And I have fixed at least the Keccak lookup in the PR.

src/zkevm_specs/evm/table.py
```diff
@@ -441,29 +461,38 @@ def __init__(
...
+                    CopyTableRow(
+                        src_id=first_row.id,
+                        src_type=first_row.tag,
+                        dst_id=next_row.id,
+                        dst_type=next_row.tag,
+                        src_addr=first_row.addr,
+                        src_addr_end=first_row.src_addr_end,
+                        dst_addr=next_row.addr,
+                        length=first_row.bytes_left,
+                        value_rlc=value_rlc,
```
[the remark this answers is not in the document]
roynalnaruto: Yes. This part was a bit tricky since the RLC accumulated value is only in the last row, but all other fields are available in the first row.

So for `row.is_first` I store `first_row` and `next_row` (which is the second row). And wait until we encounter `row.is_last`, where I get the `row.value` and populate the rest of the fields from `first_row` and `next_row`. What do you think about this approach? And in your opinion, is it possible to avoid this?

In the circuits repository, I can add the RLC accumulated value to [`CopyEvent`](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/execution.rs#L206) itself, so we won't need to do this dynamic rotation. Wdyt?

src/zkevm_specs/evm/typing.py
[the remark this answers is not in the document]
ChihChengLiang: Just checked, I think the new commit makes sense now.

src/zkevm_specs/evm/table.py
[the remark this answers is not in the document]
ed255: > In the circuits repository, I can add the RLC accumulated value to [`CopyEvent`](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/execution.rs#L206) itself, so we won't need to do this dynamic rotation. Wdyt?

I think that still requires a dynamic rotation.

This is the current lookup expression:
```rust
impl<F: Field> LookupTable<F> for CopyCircuit<F> {
    fn table_exprs(&self, meta: &mut VirtualCells<F>) -> Vec<Expression<F>> {
        vec![
            meta.query_advice(self.is_first, Rotation::cur()),
            meta.query_advice(self.id, Rotation::cur()), // src_id
            self.tag.value(Rotation::cur())(meta),       // src_tag
            meta.query_advice(self.id, Rotation::next()), // dst_id
            self.tag.value(Rotation::next())(meta),      // dst_tag
            meta.query_advice(self.addr, Rotation::cur()), // src_addr
            meta.query_advice(self.src_addr_end, Rotation::cur()), // src_addr_end
            meta.query_advice(self.addr, Rotation::next()), // dst_addr
            meta.query_advice(self.bytes_left, Rotation::cur()), // length
            meta.query_advice(self.rw_counter, Rotation::cur()), // rw_counter
            meta.query_advice(self.rwc_inc_left, Rotation::cur()), // rwc_inc_left
        ]
    }
}
```
Which query would you do to get the `value_rlc`?  It would be `meta.query_advice(self.value, Rotation(n))` but `n` is not a fixed known value.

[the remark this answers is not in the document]
roynalnaruto: Okay, I understand the problem now. In that case may be I can add an additional field `acc_rlc` which is `0` for all other copy data types except for `RlcAcc`. And in this case it will be the same over all rows, so we get an additional constraint of checking whether the `value == acc_rlc` for the last row.

Once this is in place, we only need to lookup the first/second row (which is 0 or 1 rotation) even for the `acc_rlc`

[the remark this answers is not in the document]
ed255: > Okay, I understand the problem now. In that case may be I can add an additional field `acc_rlc` which is `0` for all other copy data types except for `RlcAcc`. And in this case it will be the same over all rows, so we get an additional constraint of checking whether the `value == acc_rlc` for the last row.
> 
> Once this is in place, we only need to lookup the first/second row (which is 0 or 1 rotation) even for the `acc_rlc`

This solution sounds correct!

Another idea that came to my mind which doesn't need a new column but complicates things is:
Instead of calculating the `acc_rlc` from top to bottom, do it from bottom to top.  So `acc_rlc[i] = acc_rlc[i+1] * r + v`, whichi leaves the `RLC(v)` in `acc_rlc[0]`.  The complication with this is that then we no longer have `RLC(reversed(input))` but `RLC(input)`.  And currently the bytecode circuit uses `RLC(reversed(input))`.  We could get back to this idea in the future as an optimization maybe.

So I think adding this new column which contains a copy of the last write value in all the steps in an event is a good way to go!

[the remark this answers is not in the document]
roynalnaruto: I just pushed here: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/235/commits/68ead11f2f40c706f3ecf11db21a02b3b6fbf591

Would be nice to get a review on this one :) @ed255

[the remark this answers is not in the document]
roynalnaruto: I think I personally would prefer keeping the practice of `RLC(reversed(input))` uniform throughout the codebase, if it is not too much of an issue adding an additional column. This column is `0` over all copy data types, except only for the `RlcAcc` type. So additionally there are 2 constraints:
1. `acc_rlc` is the same over all rows
2. `acc_rlc == value` for the last row (in case `dst_type` is `RlcAcc`.

tests/evm/test_sha3.py
```diff
@@ -0,0 +1,139 @@
...
+        [
+            (i, memory_snapshot[i] if i < len(memory_snapshot) else 0)
+            for i in range(offset, offset + length)
+        ]
+    )
+
+    bytecode = Bytecode()
+    for i, chunk in enumerate(memory_chunks):
+        bytecode.push(32 * i, n_bytes=32).push(chunk, n_bytes=32).mstore()
+    bytecode.push(offset_rlc, n_bytes=32).push(length_rlc, n_bytes=32).sha3().stop()
```
ed255, line 57: Although these opcodes are not tested in this test, I think it should be `offset` and `length` instead of their RLC versions.

ed255 (approved):
LGTM!  Great work!

tests/evm/test_sha3.py
roynalnaruto: oops, that's a silly error. Fixing it now.

src/zkevm_specs/util/param.py
```diff
@@ -72,6 +74,8 @@
 MEMORY_EXPANSION_QUAD_DENOMINATOR = 512
 # Coefficient of linear part of memory expansion gas cost
 MEMORY_EXPANSION_LINEAR_COEFF = 3
+# Coefficient of linear part of memory expansion gas cost, specifically in the case of SHA3
+MEMORY_EXPANSION_LINEAR_COEFF_SHA3 = 6
```
ChihChengLiang, line 78: Do you have the source where this parameter comes from? I don't think SHA3 has a special memory expansion cost.

src/zkevm_specs/evm/typing.py
```diff
@@ -666,19 +667,42 @@ def _append(
...
+    rows: List[KeccakTableRow]
+
+    def __init__(self) -> None:
+        self.rows = []
+
+    def add(self, data: bytes, r: FQ) -> KeccakCircuit:
+        output = RLC(keccak256(data), r, n_bytes=32)
+        acc_input = FQ.zero()
+        for i in range(len(data)):
+            acc_input = acc_input * r + data[i]
```
ChihChengLiang, line 680: Maybe use standard API so that we don't worry about the order of bytes?
```suggestion
        acc_input = RLC(data, r, n_bytes=len(data))
```

ChihChengLiang (commented):
I added some questions

src/zkevm_specs/util/param.py
roynalnaruto: So I initially checked here: https://evm.codes

Then verified it here: https://github.com/ethereum/go-ethereum/blob/master/core/vm/gas_table.go#L259

which uses: https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L40

ChihChengLiang: Right, but we have two variables in `param.py`, `GAS_COST_COPY_SHA3` and `MEMORY_EXPANSION_LINEAR_COEFF_SHA3`. They both have the value 6.
It looks like the `Keccak256WordGas=6` from `protocol_params.go` works like `GAS_COST_COPY_SHA3`.

But I don't get the `MEMORY_EXPANSION_LINEAR_COEFF_SHA3` one. You use it in the test with `memory_expansion`.
I'm also wondering why the test didn't fail if I change the `memory_expansion` back using the 3 gas expansion cost.

roynalnaruto: I declared the `MEMORY_EXPANSION_LINEAR_COEFF_SHA3` only because there was a `MEMORY_EXPANSION_LINEAR_COEFF` that was being used. So I assumed I was supposed to follow the same pattern.

What change did you make specifically before testing?

ChihChengLiang: I made this change
```diff
     next_memory_size, memory_expansion_cost = memory_expansion(
-        offset + length, offset + length, MEMORY_EXPANSION_LINEAR_COEFF_SHA3
+        offset + length, offset + length
     )
```

roynalnaruto: Oh nice catch. So the `MEMORY_EXPANSION_LINEAR_COEFF` should stay at `3`. It's only the multiplier for the minimum word size that changes from `3` to `6` (in the `SHA3` case).

The reason why the tests still passed after you made the above change is that, since the memory was already populated until `offset + length`, there is no further expansion and hence the the `memory_expansion_cost` is `0`.

roynalnaruto: Fixed it in the latest commit :)

ChihChengLiang (approved):
LGTM now. Great PR. Thank you for the contribution!

src/zkevm_specs/evm/table.py
```diff
@@ -289,6 +289,15 @@ class CopyDataTypeTag(IntEnum):
...
     TxLog = auto()
 
+    # RLC Accumulator tag can be used whenever we wish to
+    # accumulates `value` iteratively over all the steps in
+    # copy circuit. This is specifically used in the SHA3
+    # opcode execution where the copy table's last row has
+    # an accumulated value that is the RLC representation of
+    # all input bytes. Using this value, we can then lookup
+    # the Keccak table for the SHA3 of the input bytes.
+    RlcAcc = auto()
```
DreamWuGit, line 299: would you like to update copy-table in  `zkevm-specs/specs/tables.md `with new added field ?  maybe `zkevm-specs/specs/copy-proof.md` as well ?

DreamWuGit (approved):

src/zkevm_specs/evm/table.py
roynalnaruto: Added :)

PR #236, opened 2022-07-26
Consolidate MPT update lookups in state circuit

z2trillion:
Only make one mpt update lookup per access group in the state circuit, which allows us to remove the mpt counter


ed255:
Could you update `specs/tables.md` with:
- In `rw_table` rename column `Aux0` to `CommittedValue`
- In `mpt_table` remove all the field-specific tables, and leave only a single table that reflects the changes in this pr (mainly adding `root` and `prev_root`, removing `counter`.

Thanks!

ed255:
@z2trillion please ping me once this PR is ready for review (no longer draft).

z2trillion:
@ed255, sorry about that. I realized that I had forgotten to change the state roots in the assignment function.

ed255:
> @ed255, sorry about that. I realized that I had forgotten to change the state roots in the assignment function.

ah no worries :)

z2trillion:
@ed255, when will you be able to take a look at this?

specs/tables.md
```diff
@@ -194,83 +195,22 @@ Provided by the MPT (Merkle Patricia Trie) circuit.
...
+
+There are four types of MPT updates which the circuit proves are correct: account nonce,
+balance, or code hash updates, or account storage updates.
+
+| Address | FieldTag | Key | ValuePrev | Value | RootPrev | Root |
+| - | - | - | - | - | - | - |
+| $addr | Nonce | 0 | $noncePrev | $nonceCur | $rootPrev | $root |
+| $addr | Balance | 0 | $balancePrev | $balanceCur | $rootPrev | $root |
+| $addr | CodeHash | 0 |$codeHashPrev | $codeHash | $rootPrev | $root |
+| $addr | 0 | $key | $valuePrev | $value | $rootPrev | $root |
```
ed255, line 213:
```suggestion
| $addr | Storage | $key | $valuePrev | $value | $rootPrev | $root |
```

src/zkevm_specs/state.py
```diff
@@ -281,32 +254,25 @@ def check_call_context(row: Row, row_prev: Row):
...
 
-    # 6.1. When keys don't change, committed_value must be kept equal
-    if all_keys_eq(row, row_prev):
-        assert get_committed_value(row) == get_committed_value(row_prev)
-
-    # TODO: The current spec does an MPT lookup for every storage update.  The
-    # next optimization consists on doing a single lookup merging all updates
-    # for a given key, using the first and last access values.
-
     # 6.2. MPT storage lookup with incremental counter
```
ed255, line 265:
```suggestion
    # 6.2. MPT storage lookup for last access to (address, field_tag)
```

src/zkevm_specs/state.py
```diff
@@ -510,11 +469,11 @@ def keys_rwc_to_limbs_in_order(row: Row) -> List[FQ]:
     elif row.tag() == Tag.Stack:
         check_stack(row, row_prev)
     elif row.tag() == Tag.Storage:
-        check_storage(row, row_prev, tables)
+        check_storage(row, row_prev, row_next, tables)
```
ed255, line 472: I think a set of constraints is missing to forbid the prover from changing the root in unexpected rows.
Now we're only doing MPT lookup on the last entry for a unique key when field is Account or Storage.  So that's the only row where a root value change can happen.  I think we should constraint the rest of the rows to have a root which is the same as the previous row.

Otherwise, this example would pass the constraint but should be invalid:

| tag | address | root |
| ---  | ---           | ---   |
| TxRefund | A | X |
| Account | B | Y |
| Account | B | Z | 
| Memory | 1 | Z |

Here the root goes from X to Z, but only Y-Z transitions are MPT-proved, so in the transition X-Y the MPT could have an arbitrary change.

So I think the following constraint should be added:
```python
if not(not all_keys_eq(row, row_next) and (tag == Account or tag == Storage)):
    row.root == row_prev.root

**Review by ed255 — COMMENTED** [review]
In general look good but I believe a constraint is missing to limit when the root value can be changed.

File: specs/tables.md [hunk 4]
```diff
@@ -194,83 +195,22 @@ Provided by the MPT (Merkle Patricia Trie) circuit.
...
-- Address: `address_rlc`
-- Key: `key_rlc_mult`
-- ValuePrev: `is_nonce_mod * sel1 + is_balance_mod * sel2 + is_codehash_mod * sel2 + is_storage_mod * mult_diff`
-- ValueCur: `is_nonce_mod * s_mod_node_hash_rlc + is_balance_mod * c_mod_node_hash_rlc + is_codehash_mod * c_mod_node_hash_rlc + is_storage_mod * acc_c`
+- Value
+- RootPrev
+- Root
+
+There are four types of MPT updates which the circuit proves are correct: account nonce,
+balance, or code hash updates, or account storage updates.
```
miha-stopar, line 206: It appears code hash cannot change (without deleting the account and recreating it), so I removed code hash update from the MPT circuit. The proofs can be for: nonce change, balance change, storage change, delete account, account does not exist. Note that create account can be seen for example as nonce change.

miha-stopar (commented):
LGTM, but agree with Edu's comment.

specs/tables.md
ed255: Related: https://github.com/privacy-scaling-explorations/zkevm-specs/issues/249
The table specs are pending an update to reflect this.

src/zkevm_specs/state.py
ChihChengLiang: Haven't checked all the context yet. It looks like we can simplify
```
not(not all_keys_eq(row, row_next) and (tag == Account or tag == Storage))
```
to 

```
not(not all_keys_eq(row, row_next) and (tag in (Account, Storage))
```

to

```
all_keys_eq(row, row_next) or (tag not in (Account, Storage))
```

specs/tables.md
z2trillion: Even if code hashes cannot change, I believe we still need them in the MPT circuit to prove that the initial values for the code hashes in the rw table are correct?

src/zkevm_specs/state.py
z2trillion: I ended up adding `assert row.root == row_prev.root` for each of the tags, because the degree of the expression for `all_keys_eq(row, row_next) or (tag not in (Account, Storage))` would raise the total degree of the constraint system.

specs/tables.md
miha-stopar: Yes, code hash is still part of the circuit and it needs to be checked, it's just that the proof for code hash change is not supported.

z2trillion: got it. I've updated this description to match that.

src/zkevm_specs/state.py
ed255: Makes sense, addressing this :)

ed255 (approved):
LGTM!

lispc (approved):

PR #237, opened 2022-07-27
Update Public Input circuit 

davidnevadoc:
- Add 0 row in `tx_table` and `block_table`
 - Update selectors `q_end`, `q_not_end`
 - Update field names to match MD spec

### DIFF
--- a/src/zkevm_specs/public_inputs.py
+++ b/src/zkevm_specs/public_inputs.py
@@ -128,25 +128,26 @@ def verify_circuit(
     assert rows[0].rpi_rlc_acc == witness.public_inputs.rpi_rlc
 
     # 1.2 chain_id copy constraint from public input to raw_public_inputs
-    assert rows[BlockTag.ChainId - 1].raw_public_inputs == witness.public_inputs.chain_id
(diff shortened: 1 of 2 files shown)


han0110 (approved):
LGTM.

But just realized that in other circuits to use `{block,tx}_table`, we'd still multiply the `q_{block,tx}_table` to make sure prover doesn't put random values in rows with `q_{block,tx}_table = 0`. So this all-zero row seems not necessary (but having it might make things more straightforward).

But this increase degree of lookup argument, so another approach is to also constrain values to be `0` when `q_{block,tx}_table = 0 && not blinding_rows` (need another selector).

ed255 (approved):
LGTM!  I saw Han's comment and I see his point; I think we could move that discussion to an issue maybe?

han0110:
Builtin lookup argument has implicit degree 4 `{q_usable,input,table,z}`, normally input will have a selector, so it makes the degree to `5`, which is the maximum for `extended_k = 2`.

If table also needs a selector, it makes the `extended_k = 3`, it should be fine for now, but in the future if we want `extended_k = 2` for prover efficiency, then we might need to refactor.

PR #238, opened 2022-07-27
Add new lines for readability

z2trillion:


specs/opcode/F3RETURN.md
```diff
@@ -43,6 +43,7 @@ the following parameters:
     - `dst_addr: 0`
     - `rw_counter: callee.rw_counter`
     - `rwc_inc: length`
+
 D. Otherwise, if it's not a root call, it copies the source memory chunk to the
```
ChihChengLiang, line 47: Can we also fix the order of the A, B, C, and D here? We can remove them all if you like.

ed255: I wrote them in this order to follow an `if else` logic.  It's something like: (if X then A else D) and (if Y then B else C).
I used letters instead of numbers because I thought having unsorted numbers would be more confusing.

ed255 (approved):
LGTM!

specs/opcode/F3RETURN.md
ChihChengLiang: Oh, I see. Then LGTM now.

ChihChengLiang (approved):
LGTM

PR #240, opened 2022-07-28
Unify RLC functions

ChihChengLiang:
We defined duplicated `linear_combine` functions throughout the codebase.

We define a common `linear_combine` to make sure all circuits behave in the same way.

### DIFF
--- a/src/zkevm_specs/util/arithmetic.py
+++ b/src/zkevm_specs/util/arithmetic.py
@@ -1,8 +1,27 @@
 from __future__ import annotations
-from typing import Protocol, Sequence, Type, TypeVar, Union
-from functools import reduce
+from typing import Sequence, Protocol, Type, TypeVar, Union
 from py_ecc import bn128
 from py_ecc.utils import prime_field_inv
+
+
+def linear_combine(le_bytes: Sequence[Union[int, FQ]], base: FQ, range_check: bool = True) -> FQ:
+    """
+    Aggregate bytes into a single field element.
+    If we intend to use it as a commitment, the base must be a secured random number.
+    >>> r = 10
+    >>> assert linear_combine([1, 2, 3], r) == 1 * r**2 + 2 * r + 3
+    If the input represents a sequence of data, apply the function directly.
+    If the input represents a number, it must be in little-endian order.
+    Do not use linear_combine(le_limbs, 256) to evaluate the integer value.
+    """
+    result = FQ.zero()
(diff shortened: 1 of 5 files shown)


src/zkevm_specs/util/arithmetic.py
```diff
@@ -1,10 +1,29 @@
 from __future__ import annotations
-from typing import Protocol, Sequence, Type, TypeVar, Union
-from functools import reduce
+from typing import Sequence, Protocol, Type, TypeVar, Union
 from py_ecc import bn128
 from py_ecc.utils import prime_field_inv
 
 
+def linear_combine(le_bytes: Sequence[Union[int, FQ]], base: FQ, range_check: bool = True) -> FQ:
```
ed255, line 7: Maybe don't use the naming `bytes` for the variable, because input values higher than 255 when `range_check = False` are also valid?

src/zkevm_specs/util/arithmetic.py
```diff
@@ -1,10 +1,29 @@
...
+    """
+    Aggregate bytes into a single field element.
+    If we intend to use it as a commitment, the base must be a secured random number.
+    >>> r = 10
+    >>> assert linear_combine([1, 2, 3], r) == 1 * r**2 + 2 * r + 3
```
ed255, line 12: Shouldn't this be `== 1 + r * 2 + r**2 + 3`?

ChihChengLiang: 😱 nice catch

src/zkevm_specs/util/arithmetic.py
ChihChengLiang: sounds good. I'll go with `seq`

ChihChengLiang: Addressed in [a375590](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/240/commits/a375590df25dd9318f4b34d676096adbde15e235)

src/zkevm_specs/util/arithmetic.py
ChihChengLiang: Addressed in [a375590](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/240/commits/a375590df25dd9318f4b34d676096adbde15e235)

ed255 (approved):
LGTM!  It's always nice to have these removals of redundant code :D

Issue #231, opened 2022-07-04
Missing CallDataGasCost in TxTable (tables.md, tx circuit, publicInputs circuit)


ed255:
CallDataGasCost is defined in the EVM Circuit but is missing in other places.

Maybe we can begin unifying the TxFieldTag enums between evm tables and tx circuit.

PR #243, opened 2022-07-29
Update EndBlock spec; update tx_table

ed255:
The main change is using 2 strategic lookups to the padding rows with sequential numbers to prove a lower bound on padding rows which allows proving the number of rw operations in the rw_table and txs in the tx table; as devised by @z2trillion in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/558

NOTE: To implement this spec, EndBlock needs to be height = 1, which can be achieved after reducing the number of cells used in the state selector in the EVM circuit (by increasing the selector expression degree by 1).

Resolve https://github.com/privacy-scaling-explorations/zkevm-specs/issues/231

### DIFF
--- a/src/zkevm_specs/evm/execution/end_block.py
+++ b/src/zkevm_specs/evm/execution/end_block.py
@@ -2,9 +2,15 @@ from ...util import FQ
 from ..instruction import Instruction, Transition
 from ..table import CallContextFieldTag
 
+# EndBlock is an execution state that constraints the following:
+# 1. Once the EndBlock state is reached, there's no other execution states appearing until the end of the EVM Circuit.  In particular, after the first EndBlock, there will be no more lookups to the rw_table.
+# 2. The number of meaningful entries (non-padding) in the rw_table match the rw_counter after the EndBlock state is processed.
+#   - instruction.curr.rw_counter
+# 3. The number of txs processed by the EVM Circuit match the number of txs in the TxTable
+#    total_tx = instruction.call_context_lookup(CallContextFieldTag.TxId)
 
-# TODO: Introduce constrain_instance to constrain the equality between witness
-#       and public input, for total_tx and total_rw
+# Extra
+# 4. We need to prove that the EndBlock state exists
 
 
 def end_block(instruction: Instruction):
@@ -23,9 +29,21 @@ def end_block(instruction: Instruction):
             total_rw,
             FQ(len(instruction.tables.rw_table)),
         )
+
+        # TODO: lookup to rw_table to verify rw_counter
     else:
         # Propagate rw_counter and call_id all the way down
         instruction.constrain_step_state_transition(
             rw_counter=Transition.same(),
             call_id=Transition.same(),
         )
+
+    rw_table_size = len(instruction.tables.rw_table)
+    total_rw = instruction.curr.rw_counter + 1  # extra 1 from the tx_id lookup
+    # Verify that there are at most total_rw meaningful entries in the rw_table
+    instruction.rw_table_start_lookup(1)
+    instruction.rw_table_start_lookup(rw_table_size - total_rw)
+    # Since every lookup done in the EVM circuit must succeed and uses a unique
+    # rw_counter, we know that at least there are total_rw meaningful entries
+    # in the rw_table.
+    # We conclude that the number of meaningful entries in the rw_table is total_rw.
--- a/src/zkevm_specs/evm/execution/end_block.py
+++ b/src/zkevm_specs/evm/execution/end_block.py
@@ -33,10 +33,13 @@ def get_tx_table_max_txs(table: Set[TxTableRow]) -> int:
     fixed_field_count = 0
     for row in table:
         print(vars(row))
-        if (row.field_tag != TxContextFieldTag.CallData) or (row.field_tag == TxContextFieldTag.Pad and row.tx_id != 0):
+        if (row.field_tag != TxContextFieldTag.CallData) or (
+            row.field_tag == TxContextFieldTag.Pad and row.tx_id != 0
+        ):
             fixed_field_count += 1
-    print(f'fixed_field_count = {fixed_field_count}')
+    print(f"fixed_field_count = {fixed_field_count}")
     return fixed_field_count // TxContextFieldTag.TxSignHash
+
 
 def end_block(instruction: Instruction):
     if instruction.is_last_step:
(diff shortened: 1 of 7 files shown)


src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -1,28 +1,74 @@
...
+# EndBlock.  This will require the EndBlock to have height = 1 in the circuit,
+# which can be achieved after reducing the number of cells used in the state
+# selector.
+
+# Count the max number of txs that the TxTable can hold by counting rows of
+# fixed fields + padding rows in the fixed fields section.
+def get_tx_table_max_txs(table: Set[TxTableRow]) -> int:
+    fixed_field_count = 0
+    for row in table:
+        print(vars(row))
```
ChihChengLiang, line 35: Forgot to remove the print?
```suggestion
```

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -1,28 +1,74 @@
...
+        if (row.field_tag != TxContextFieldTag.CallData) or (
+            row.field_tag == TxContextFieldTag.Pad and row.tx_id != 0
+        ):
+            fixed_field_count += 1
+    print(f"fixed_field_count = {fixed_field_count}")
```
ChihChengLiang, line 40:
```suggestion
```

src/zkevm_specs/tx.py
```diff
@@ -270,6 +269,8 @@ def verify_circuit(
     """
 
     rows = witness.rows
+    for row in rows:
+        print(vars(row))
```
ChihChengLiang, line 273:
```suggestion
```

src/zkevm_specs/tx.py
```diff
@@ -337,18 +338,19 @@ def tx2witness(
 
     sign_verification = SignVerifyChip.assign(sig, pk, tx_sign_hash, randomness)
 
+    call_data_gas_cost = sum([(4 if byte == 0 else 16) for byte in tx.data])
```
ChihChengLiang, line 341: We have `GAS_COST_TX_CALL_DATA_PER_ZERO_BYTE` and `GAS_COST_TX_CALL_DATA_PER_NON_ZERO_BYTE` in `param.py`

ChihChengLiang (commented):
Added some nitpick comments

src/zkevm_specs/evm/table.py
```diff
@@ -130,7 +130,9 @@ class TxContextFieldTag(IntEnum):
     Value = auto()
     CallDataLength = auto()
     CallDataGasCost = auto()
+    TxSignHash = auto()
```
ChihChengLiang, line 133: should we sync tables.md for these changes?

src/zkevm_specs/tx.py
```diff
@@ -16,16 +16,15 @@ class Tag(IntEnum):
...
-    CallDataLength = 10
-    TxSignHash = 11
-    CallData = 12
-    Pad = 13
+    CallerAddress = 4
+    CalleeAddress = 5
+    IsCreate = 6
+    Value = 7
+    CallDataLength = 8
+    CallDataGasCost = 9
```
ChihChengLiang, line 24: sync `tables.md`? What's the difference between `Tag` and `TxContextFieldTag`?

ChihChengLiang (commented):
Added feedback on updating the `tables.md`.
We should also rebase to apply the lint check.

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -1,28 +1,74 @@
...
+        # Verify that there are at most total_txs meaningful entries in the tx_table
+        instruction.tx_context_lookup(FQ(1), TxContextFieldTag.Pad)
+        instruction.tx_context_lookup(
+            FQ((tx_table_max_txs - total_tx.expr().n) * TxContextFieldTag.TxSignHash),
+            TxContextFieldTag.Pad,
         )
+        # Since every tx lookup done in the EVM circuit must succeed and uses a unique
+        # tx_id, we know that at least there are total_tx meaningful txs
+        # in the tx_table.
+        # We conclude that the number of meaningful txs in the tx_table is total_tx.
```
han0110, line 71: With padding it seems to add more constraints to construct the `tx_table`, so rather than enforcing rows with tag `Pad` in `tx_table` to have sequential ids, is that possible to have a `NumTx` tag somewhere (perhaps `tx_table` with `tx_id = 0` row?) for us to check directly?

han0110 (approved):
LGTM! Nice work, only got one question.

src/zkevm_specs/evm/table.py
ed255: Thanks for catching this! I've updated the `tables.md` in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/243/commits/27b969d0c735242fc726134aedd4c6441584c560!

src/zkevm_specs/tx.py
ed255: Ah good point... They are the same :sweat_smile:
I originally created a new type to account for small differences between the Tx Table seen by the EVM and the same defined by the Tx Circuit, but this PR unifies that, so I'll remove the `Tag` in favor of `TxContextFieldTag`

ed255: Addressed in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/243/commits/bdc1189fa4b923d78e0057bc05d4b6525f3c03a3

ChihChengLiang (approved):
LGTM

src/zkevm_specs/evm/execution/end_block.py
ed255: Currently the `tx_table` is constructed by the verifier (out of publicly available data), so we don't have any constraint on it.  So having the sequential ids in the `Pad` rows doesn't need any constraint, it's just a logic that the verifier must follow when preparing the table contents.
A part from this, I think I don't understand your proposal.  Do you mean to have a tag that appears only in one row which contains the total number of Txs?  If we consider that the `tx_table` is constructed by the verifier, that seems good to me.  But if this proposal is to reduce the constraint complexity, I don't see why this constraint (which requires counting txs, which I imagine would need a new column) is simpler.

han0110: If verifier always calculate same length RLC then it makes sense to me now, I thought we'd need constraint for that inside circuit, sorry for confusing.

What I imagined is that verifier doesn't need always to calculate so same length RLC when there are fewer txs than the maximum, the rest values should be auto padded with `0` because we don't add any value into RLC (or we should pad with sequential ids? otherwise prover could pad `0` ahead in advice column and still get the same RLC value, will think about this again). So instead we add another tag `NumTx` and put the value within such row, then we can lookup it directly.

ed255: Ah I see, so you're thinking about having trailing 0s in the RLC'd array so that you can skip the operations for the tail of the array.  The idea makes sense!  Nevertheless the current arrangement of the tx table into the RLC'd value is column after column, so you would get the trailing 0s in the middle.

I think this is an interesting discussion so I'll move it to an issue so that we can move on with this PR :)

han0110: Sure! Let's merge this PR first :)

ed255: Moved to https://github.com/privacy-scaling-explorations/zkevm-specs/issues/253

PR #245, opened 2022-08-08
opcode EXP

roynalnaruto:
### Addition

The copy circuit has been modified to add a `Exp` row type.

This type is constrained with:
```
write_value::next == write_value::cur * read_value::cur
```

For all `read` values being equal (`base`) and the number of `write` rows being equal to `exponent`, we have the final value equals `base ** exponent` which is the exponentiation result.

### Modification

Since the extra column (previously called `rlc_acc`) is used for multiple purposes now, it has been renamed to `aux_value` to denote an auxiliary value. The auxiliary value can be an RLC accumulator (for `RlcAcc`) or can be the exponentiation (for `Exp`). It's set to `FQ.zero()` for all other cases.

### DIFF
--- a/specs/copy-proof.md
+++ b/specs/copy-proof.md
@@ -16,6 +16,7 @@ In addition to the columns in the copy table, copy circuit adds a few auxiliary 
 - `is_tx_calldata`: indicates if `Type` is `TxCalldata` using `IsZero` gadget.
 - `is_tx_log`: indicates if `Type` is `TxLog` using `IsZero` gadget.
 - `is_rlc_acc`: indicates if `Type` is `RlcAcc` using `IsZero` gadget.
+- `is_exp`: indicates if `Type` is `Exp` using `IsZero` gadget.
 
 *Note*: We use `IsZero` gadget in the specs for simplicity. In the actual circuit implementation, we make use of `BinaryGadget`.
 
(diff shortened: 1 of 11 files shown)


roynalnaruto:
> The RLC calculation should be mod F, but EXP is supposed to be mod 2^256

Yes, I am still thinking about how to do this. In the copy circuit, each row has a `write_value` which will also be `F` field element. But it should instead be `mod 2**256`. What do you think we can do here?

z2trillion:
I don't see an easy way to do this in in the existing copy circuit without adding a lot of advice columns and EXP-only constraints. It may be necessary to have another circuit that only computes EXP?

roynalnaruto:
@z2trillion How about I add `aux_value1` and `aux_value2` to the copy circuit, and in this specific case (`Exp`), `aux_value1` represents the `lo` bytes of `a ** exponent (mod 2**256)` and `aux_value2` represents the `hi` bytes of `a ** exponent (mod 2**256)` ?

The constraints for `Exp` will also change:
- `rows[0].aux_value1 -> lo bytes of a ** k (mod 2**256)`
- `rows[0].aux_value2 -> hi bytes of a ** k (mod 2**256)`
- `rows[2].value      -> current value`
- `rows[2].aux_value1 -> lo bytes of rows[0].value * exp::prev`
- `rows[2].aux_value2 -> hi bytes of rows[0].value * exp::prev`

where `exp::prev = Word::from(lo_bytes, hi_bytes)`

The trick is to be able to compute `rows[0].aux_value` from `rows[0].aux_value1` (lo bytes) and `rows[0].aux_value2` (hi bytes)

roynalnaruto:
> How about I add aux_value1 and aux_value2 to the copy circuit, and in this specific case (Exp), aux_value1 represents the lo bytes of a ** exponent (mod 2**256) and aux_value2 represents the hi bytes of a ** exponent (mod 2**256)

This will also not work because we still need to at each read-write pair constrain the product which should be `mod 2**256`, but the constraints are only allowed over `F`

roynalnaruto:
Closing as the current approach does not work. Will re-open with a new approach

PR #254, opened 2022-08-22
Add spec for `BLOCKHASH` opcode

jeong0982:
Not sure for handling valid range. If there is any appropriate way for checking valid range, please let me know.
Thank you.

### DIFF
--- a/specs/opcode/40BLOCKHASH.md
+++ b/specs/opcode/40BLOCKHASH.md
@@ -1,0 +1,27 @@
+# BLOCKHASH opcode
+
+## Procedure
+
+The `BLOCKHASH` opcode pops `block_number` off the stack and pushes the hash from chosen block. If the chosen block number is not in valid range (last 256 blocks, not including current block), it will push 0 onto the stack.
+
+## Constraints
+
+1. opId == 0x40
+2. State transition:
+    - gc + 2 (1 stack read + 1 stack write)
+    - stack_pointer + 0 (one pop and one push)
+    - pc + 1
+    - gas + 20
+3. Lookups: 3
+    - `block_number` is popped from the stack.
+    - `current_block_number` is in the block context table.
+    - `block_hash` is in the block context table.
+
+## Exceptions
+
+1. stack underflow: if the stack starts empty
+2. out of gas: remaining gas is not enough
+
+## Code
+
+Please refer to 'src/zkevm-specs/evm/execution/blockhash.py'
--- a/specs/opcode/40BLOCKHASH.md
+++ b/specs/opcode/40BLOCKHASH.md
@@ -16,6 +16,8 @@ The `BLOCKHASH` opcode pops `block_number` off the stack and pushes the hash fro
     - `block_number` is popped from the stack.
     - `current_block_number` is in the block context table.
     - `block_hash` is in the block context table.
+4. Additional Constraint
+    - `block_number` should be in valid range.
 
 ## Exceptions
 
@@ -24,4 +26,4 @@ The `BLOCKHASH` opcode pops `block_number` off the stack and pushes the hash fro
 
 ## Code
 
-Please refer to 'src/zkevm-specs/evm/execution/blockhash.py'
+Please refer to [blockhash.py](src/zkevm-specs/evm/execution/blockhash.py).
--- a/specs/opcode/40BLOCKHASH.md
+++ b/specs/opcode/40BLOCKHASH.md
@@ -16,6 +16,8 @@ The `BLOCKHASH` opcode pops `block_number` off the stack and pushes the hash fro
     - `block_number` is popped from the stack.
     - `current_block_number` is in the block context table.
     - `block_hash` is in the block context table.
+4. Additional Constraint
+    - `block_number` should be in valid range.
 
 ## Exceptions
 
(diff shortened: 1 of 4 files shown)


src/zkevm_specs/evm/execution/blockhash.py
```diff
@@ -0,0 +1,31 @@
...
+    current_block_number = instruction.block_context_lookup(BlockContextFieldTag.Number)
+    is_current_bigger = instruction.compare(block_number, current_block_number.expr(), N_BYTES_U64)[0]
+
+    diff = current_block_number.expr() - block_number if is_current_bigger == 1 else block_number - current_block_number.expr()
+    is_invalid_range = 1 - (
+        is_current_bigger
+        * instruction.compare(diff, FQ(257), N_BYTES_U64)[0]
+    )
+    op = BlockContextFieldTag.HistoryHash
+    block_hash = FQ(0) if is_invalid_range == 1 else instruction.block_context_lookup(op, block_number)
```
roynalnaruto, line 19: You probably can't use the `Range256` variant in `FixedTable` to check whether or not the difference between `current_block_number` and `block_number` is `< 256`. Mainly because you want to handle the case where the lookup fails, i.e. simply push `0` to the stack.

So I think you will need a comparison gadget to check that `block_number < current_block_number`.

So you would have something like:
```python
# get value that was pushed to stack (RLC-encoded)
block_hash = instruction.stack_push()

# comparing block_number and current_block_number
block_lt, _ = instruction.compare(block_number, current_block_number, 8)
diff_lt, _ = instruction.compare(current_block_number - block_number, FQ(256), 2)

# get the expected block hash depending on the above conditions (RLC-encoded)
expected_block_hash = instruction.select(
    block_lt * diff_lt,
    instruction.block_context_lookup(
        BlockContextFieldTag.HistoryHash,
        block_number,
    ),
    FQ.zero(),
)

# block hash is as expected
instruction.constrain_equal(block_hash, expected_block_hash)
```

roynalnaruto (changes requested):
Mostly looks good. I have posted some comments/suggestions :)

src/zkevm_specs/evm/execution/blockhash.py
jeong0982: That's exactly what I want to know. Thank you for your comment!

jeong0982: @roynalnaruto There is an error from `block_conext_lookup()` in `select` instruction. Because `instruction.select()` function is called after calculating argument, so `block_context_lookup` with invalid `block_number` raise `LookupUnsatFailure`. Do you have any idea for handling this situation? Thank you.

roynalnaruto: Ah, of course. Yes that's expected.

In that case, you could scrap the `instruction.select(..)` and simply have:
```python
if block_lt * diff_lt == FQ.one():
    expected_block_hash = instruction.block_context_lookup(
        BlockContextFieldTag.HistoryHash,
        block_number,
    ),
else:
    expected_block_hash = FQ.zero()
```

jeong0982: Thanks a lot 👍🏼

roynalnaruto: @jeong0982 you will need to fix the formatting:

* To check which file needs formatting
```
$ make lint
```
* Formatting
```
$ black src/
$ black tests/
```

jeong0982: Ok. Thank you!

specs/opcode/40BLOCKHASH.md
```diff
@@ -0,0 +1,29 @@
...
+    - gc + 2 (1 stack read + 1 stack write)
+    - stack_pointer + 0 (one pop and one push)
+    - pc + 1
+    - gas + 20
+3. Lookups: 3
+    - `block_number` is popped from the stack.
+    - `current_block_number` is in the block context table.
+    - `block_hash` is in the block context table.
+4. Additional Constraint
+    - `block_number` should be in valid range.
```
roynalnaruto, line 20: This needs to be removed now

roynalnaruto (approved):
LGTM! But we will wait for other reviewers to review as well!

specs/opcode/40BLOCKHASH.md
jeong0982: Fixed. Thanks!

CPerezz (approved):
LGTM modulo the comment I added.

Thanks for the work!

specs/opcode/40BLOCKHASH.md
roynalnaruto: Oops! yes of course it _should_ be included. When I saw the markdown file on a small screen, due to some indentation  issues I thought the `block_number` was included within `lookups` (hence I mentioned that this is not a lookup).

@jeong0982 since this is a constraint, yes it should be present in the spec doc. Thanks for noticing @CPerezz :)

jeong0982: I was also confused. I will put it back. Thank you👍🏼

specs/opcode/40BLOCKHASH.md
```diff
@@ -0,0 +1,27 @@
...
+    - `block_hash` is in the block context table.
+
+## Exceptions
+
+1. stack underflow: if the stack starts empty
+2. out of gas: remaining gas is not enough
+
+## Code
+
+Please refer to 'src/zkevm-specs/evm/execution/blockhash.py'
```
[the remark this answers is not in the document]
jeong0982: Fixed. Thanks!

roynalnaruto:
Thank you @jeong0982 for the contribution! I will merge this since it has 2 approvals and all comments/discussions have been resolved!

PR #255, opened 2022-08-23
Exponentiation circuit and EXP opcode

roynalnaruto:
* Design Document: https://hackmd.io/@rohitnarurkar/BJhpYGiCc
* Circuits Implementation: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/700

This PR implements the following:
- [x] EXP Gadget
- [x] Exponentiation Circuit
- [x] Exponentiation Table (verified within Exponentiation Circuit)

The `EXP` opcode computes the exponentiation result of an integer base to an integer exponent:
```
base ^ exponent == exponentiation (mod 2^256)
```

The  exponentiation circuit verifies the exponentiation trace, i.e. intermediate multiplication steps generated by the [Exponentiation by Squaring](https://hackmd.io/@rohitnarurkar/BJhpYGiCc#Background) algorithm.

The `EXP` opcode gadget does lookups (conditional based on the `exponent` value) to the exponentiation table.

### DIFF
--- a/specs/exp-proof.md
+++ b/specs/exp-proof.md
@@ -56,7 +56,7 @@ The exponentiation circuit consists of the following columns:
         - `exponent` MUST reduce to `exponent // 2`, which means:
 	    - Low 128 bits of `exponent::cur` MUST equal low 128 bits of `parity_check`'s multiplication result `d`.
 	        - That is, `exp_table.exponent_lo == parity_check.d_lo`.
-	    - High 128 bits of `exponent::cur` MUST equal high 128 bits of `parity_check's multiplication result `d`.
+	    - High 128 bits of `exponent::cur` MUST equal high 128 bits of `parity_check`'s multiplication result `d`.
 	        - That is, `exp_table.exponent_hi == parity_check.d_hi`.
 	    - Low 128 bits of `exponent::next` MUST equal low 128 bits of `parity_check`'s multiplicand `q`.
 	        - That is, `exp_table.exponent_lo::next == parity_check.q_lo`. We compute `q_lo` using the 64-bit limbs of `q`.
--- a/specs/exp-proof.md
+++ b/specs/exp-proof.md
@@ -21,46 +21,56 @@ For a detailed algorithm, please refer the [design document](https://hackmd.io/@
 
 ## Circuit Layout
 
-The exponentiation circuit contains columns from the [exponentiation table](./tables.md#exponentiation-table) in addition to the following columns:
-1. `q_step`: A selector to indicate whether or not the current row represents a step in the exponentiation trace.
-2. `is_pad`: A boolean-valued advice column to indicate whether or not the step is reserved for padding or not.
-3. `is_odd`: a boolean-valued advice column to indicate whether or not the exponent at this step is odd or even.
-4. `quotient`: An RLC field representing the quotient in `2 * q + is_odd == exponent`.
-
-We even use a few columns to perform multiplication (modulo `2^256`) of the intermediate steps in the exponentiation trace.
+The exponentiation circuit consists of the following columns:
+1. `q_usable`: A selector to indicate whether or not the current row will be used in the circuit's layout.
+2. `exp_table`: The columns from [exponentiation table](./tables.md#exponentiation-table)
+3. `mul_gadget`: The columns from a multiplication gadget, responsible for validating that each step within the exponentiation trace was multiplied correctly. For instance, in the above example we would want to verify that `729 * 729 == 531441 (mod 2^256)`.
+4. `parity_check`: The columns from a multiplication gadget, responsible for checking the parity (odd/even) of the `exponent` at the specific step of exponentiation trace. Depending on whether the `exponent` is odd or even, we calculate the `exponent` at the next step.
 
 ## Circuit Constraints
 
-- For every step except the last step, validate that:
+- For every row where `is_step == true`, except the last step, validate that:
     - `base` MUST be the same across subsequent steps.
     - Multiplication result `d` from the next row MUST be equal to the first multiplicand `a` in the current row. For instance, if we consider `row_0` (`531441 * 3 = 1594323`) and `row_1` (`729 * 729 = 531441`) from the above example, `d_1` is `531441` and `a_0` is also `531441`.
-    - `identifier` MUST be the same across subsequent steps.
+    - `identifier` MUST be the same across subsequent steps, i.e. `identifier::cur_step == identifier::next_step`.
 
-- For every step, validate that:
-    - `is_last` MUST be boolean.
-    - `is_pad` MUST be boolean.
-    - `is_odd` MUST be boolean.
-    - `is_last == 1` MUST be followed by a padding row, i.e. `is_pad::next == 1`.
-    - `is_last == 0` MUST be followed by a non-padding row, i.e. `is_pad::next == 0`.
-    - The multiplication `a * b + c == d` MUST be assigned correctly, where `c == 0`.
-    - At every intermediate step in the exponentiation trace, the `exponentiation` MUST be equal to `d`, which is the result of the multiplication.
-    - The multiplication `2 * q + is_odd == exponent` MUST be correct.
+- For every row, validate that:
+    - `exp_table.is_step` MUST be boolean.
+    - `exp_table.is_last` MUST be boolean.
 
-- For every step except the last step where `is_odd == 1`:
-    - The second multiplicand in the multiplication `b` MUST be equal to `base`.
-    - `exponent::next == exponent::cur - 1`, which implies:
-        - `exponent_lo::next == exponent_lo::cur - 1`.
-	- `exponent_hi::next == exponent_hi::cur`.
-
-- For every step except the last step where `is_odd == 0`:
-    - Both the multiplicands in the multiplication `a` and `b` MUST be equal, i.e. `a == b`.
-    - `exponent::next == exponent::cur // 2`, which implies:
-        - `exponent_lo::next == quotient_lo`.
-	- `exponent_hi::next == quotient_hi`.
-
-- For the last step in the exponentiation trace, i.e. `is_last == 1`:
-    - The exponent has reduced to `2`, i.e. `exponent_lo == 2` and `exponent_hi == 0`.
-    - Both multiplicands in the multiplication are equal to the `base` of the exponentiation operation, i.e. `a == b == base`.
+- For every row where `is_step == true`, validate that:
+    - `exponentiation_lo` MUST equal `mul_gadget`'s multiplication result `d_lo`.
+    - `exponentiation_hi` MUST equal `mul_gadget`'s multiplication result `d_hi`.
+    - Since we are only performing multiplication with the equation `a * b + c == d`, `c` in the `mul_gadget` MUST equal `0`.
+    - Since we are performing `2 * q + r == exponent` in the `parity_check` multiplication gadget, `r` is boolean, that is:
+        - `r_hi` MUST equal `0`.
+	- `r_lo` MUST be boolean.
+	- `overflow` in the `parity_check` MUST equal `0`.
+    - If parity check is `odd`, i.e. `parity_check.r_lo == 1`:
+        - `exponent` MUST reduce by 1, which means:
+	    - Low 128 bits of `exponent::next` MUST equal low 128 bits of `exponent::cur - 1`.
+	    - High 128 bits of `exponent::next` MUST equal high 128 bits of `exponent::cur`.
+	- `exponent` is odd also means it was a multiplication by `base` operation, that is:
+	    - For each limb, `exp_table.base_limb == mul_gadget.b`.
+    - If parity check if `even`, i.e. `parity_check.r_lo == 0`:
+        - `exponent` MUST reduce to `exponent // 2`, which means:
+	    - Low 128 bits of `exponent::cur` MUST equal low 128 bits of `parity_check`'s multiplication result `d`.
+	        - That is, `exp_table.exponent_lo == parity_check.d_lo`.
+	    - High 128 bits of `exponent::cur` MUST equal high 128 bits of `parity_check's multiplication result `d`.
+	        - That is, `exp_table.exponent_hi == parity_check.d_hi`.
+	    - Low 128 bits of `exponent::next` MUST equal low 128 bits of `parity_check`'s multiplicand `q`.
+	        - That is, `exp_table.exponent_lo::next == parity_check.q_lo`. We compute `q_lo` using the 64-bit limbs of `q`.
+	    - High 128 bits of `exponent::next` MUST equal high 128 bits of `parity_check`'s multiplicand `q`.
+	        - That is, `exp_table.exponent_hi::next == parity_check.q_hi`. We compute `q_hi` using the 64-bit limbs of `q`.
+	- `exponent` is even also means it was a squaring operation, that is:
+	    - `mul_gadget`'s multiplicands MUST be equal, or, `mul_gadget.a == mul_gadget.b`.
+    - For the last step, i.e. `is_last == true`, validate that `a * a == a^2 (mod 2^256)`:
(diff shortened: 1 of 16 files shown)


specs/exp-proof.md
```diff
@@ -0,0 +1,67 @@
...
+    - `is_last == 0` MUST be followed by a non-padding row, i.e. `is_pad::next == 0`.
+    - The multiplication `a * b + c == d` MUST be assigned correctly, where `c == 0`.
+    - At every intermediate step in the exponentiation trace, the `exponentiation` MUST be equal to `d`, which is the result of the multiplication.
+    - The multiplication `2 * q + is_odd == exponent` MUST be correct.
+
+- For every step except the last step where `is_odd == 1`:
+    - The second multiplicand in the multiplication `b` MUST be equal to `base`.
+    - `exponent::next == exponent::cur - 1`, which implies:
+        - `exponent_lo::next == exponent_lo::cur - 1`.
+	- `exponent_hi::next == exponent_hi::cur`.
```
ChihChengLiang, line 53: Are the low and high 128 bits?
If so, then this constraint is incorrect if we have `is_odd = 1`, `exponent_lo::cur = 0`, and  `exponent_hi::cur = 1`?
The exponent_lo::next should be 2^128 -1 and the exponent_hi::next would be 0

specs/exp-proof.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Circuit Constraints
+
+- For every step except the last step, validate that:
+    - `base` MUST be the same across subsequent steps.
+    - Multiplication result `d` from the next row MUST be equal to the first multiplicand `a` in the current row. For instance, if we consider `row_0` (`531441 * 3 = 1594323`) and `row_1` (`729 * 729 = 531441`) from the above example, `d_1` is `531441` and `a_0` is also `531441`.
+    - `identifier` MUST be the same across subsequent steps.
+
+- For every step, validate that:
+    - `is_last` MUST be boolean.
```
ChihChengLiang, line 40: What does `is_last` mean?
If I have two instances to prove:

1. 3^13
2. 5^17

Do we set `is_last =1` at the end each of the instances 3^13 and 5^17? Or do we do so only at the end of the last instance 5^17?

specs/tables.md
```diff
@@ -356,3 +356,72 @@ The table below lists all of copy pairs supported in the copy table:
...
+| $rwc       | 1       | 3          | 0          | 0          | 0          | 2           | 0           | 2            | 9                 | 0                 |
+
+For `exponent == 13`, i.e. scenario #4 we do two lookups:
+1. Lookup to first row:
+```
+Row(identifier=rwc, is_last=0, base_limbs=[3, 0, 0, 0], exponent_lo_hi=[13, 0], lsb_exponent=13, exponentiation_lo_hi=[1594323, 0])
+```
+2. Lookup to last row:
+```
+Row(identifier=rwc, is_last=1, base_limbs=[3, 0, 0, 0], exponent_lo_hi=[2, 0], lsb_exponent=2, exponentiation_lo_hi=[9, 0])
```
ChihChengLiang, line 426: Why do we need this lookup? It seems the first lookup is already enough for the EXP opcode. What would happen if we don't do this lookup?

specs/tables.md
```diff
@@ -356,3 +356,72 @@ The table below lists all of copy pairs supported in the copy table:
...
+Using the above algorithm, `3 ^ 13 == 1594323 (mod 2^256)` is broken down into the following steps:
+```
+3      * 3   = 9
+9      * 3   = 27
+27     * 27  = 729
+729    * 729 = 531441
+531441 * 3   = 1594323
+```
+
+We assign the above steps to the exponentiation table in the reverse order, so that the first step is `531441 * 3 = 1594323`. From here on, the RHS in the above steps is termed as `intermediate_exponentiation`. We define another term `intermediate_exponent` as a value that starts at the integer exponent of the operation, i.e. `13` in the above case, and reduces down to `2` such that:
```
ChihChengLiang, line 386: Why do we have to do the reverse order? It seems possible to do the normal order and add up to the exponent.

ChihChengLiang (commented):
I reviewed the text description parts.
There are many unorganized questions. Feel free to pick just the ones you feel important to answer

roynalnaruto:
@adria0 no blockers. I am updating the specs repo and also the design document. Should push an update this week

roynalnaruto:
@ChihChengLiang @adria0 Apologies I was a little occupied with some other tasks. I will post updates to the design doc as well as the specs PR this week. Will also post answers to all your comments. Thanks for the review :)

specs/exp-proof.md
roynalnaruto: In the case of the exponentiation circuit, `is_last` is set as true on the row representing the last entry in the exponentiation trace. Which should always be `a * a == a^2` (for `a ^ b == d`).

There will be exactly one row per trace where `is_last == 1`. So for `n` traces in the entire circuit, there will be `n` rows where `is_last == 1`.

Example, if we have two exponentiation operations in a single block, namely:
1. `3 * 3 == 27`
2. `4 * 3 == 64`

Then we would have (a simplified version):
| is_first | is_last | base | exponent | exponentiation |
|--------|--------|------|----------|----------------|
| 1           | 0          | 3       | 3              | 27                      |
| 0          | 1           | 3       | 2              | 9                        |
| 1           | 0          | 4       | 3              | 64                     |
| 0          | 1           | 4       | 2              | 16                      |

So it is at the end of each instance. I will add this to the design document to make it clearer :)

specs/tables.md
roynalnaruto: It can be done in the normal order as well. When I started designing this initially, it was more intuitive for me to do a lookup to the first step of the trace, instead of the last step.

It was just a design choice from my side, it can in fact be done the other way as well. What do you think?

specs/exp-proof.md
```diff
@@ -0,0 +1,67 @@
...
+531441 * 3   = 1594323
+```
+
+Moving on, we refer each of the above multiplication steps as:
+```
+a * b + c = d
+```
+while validating the equation using a `MulAddWordsGadget` where `c == 0`.
+
+For a detailed algorithm, please refer the [design document](https://hackmd.io/@rohitnarurkar/BJhpYGiCc).
```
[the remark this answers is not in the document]
roynalnaruto: I'll add it to the repo 👍

specs/tables.md
```diff
@@ -356,3 +356,72 @@ The table below lists all of copy pairs supported in the copy table:
...
+
+Depending on the value of the `exponent` within the exponentiation operation, the `EXP` gadget will either:
+1. Do no lookup if `exponent == 0` since `base ^ 0 == 1 (mod 2^256)`
+2. Do no lookup if `exponent == 1` since `base ^ 1 == base (mod 2^256)`
+3. Do 1 lookup to a row if `exponent == 2` since there is a single step in the exponentiation trace, i.e. `base ^ 2 == base * base (mod 2^256)`, implying that `is_first == is_last == 1` for this row.
+4. Do 2 lookups to 2 different rows if `exponent > 2` since there are more than one steps in the exponentiation trace, i.e. a lookup to `is_last == 0` and a lookup to `is_last == 1`.
+
+Consider `3 ^ 13 == 1594323 (mod 2^256)`. The exponentiation table assignment looks as follows:
+
+| identifier | is_last | base_limb0 | base_limb1 | base_limb2 | base_limb3 | exponent_lo | exponent_hi | lsb_exponent | exponentiation_lo | exponentiation_hi |
```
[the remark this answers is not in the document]
roynalnaruto: The `lsb_exponent` was used in the earlier version (to check odd/even parity of the exponent). I am using a different approach now, so that column is redundant. Will remove it from the docs, thanks for catching that!

specs/tables.md
```diff
@@ -356,3 +356,72 @@ The table below lists all of copy pairs supported in the copy table:
...
+|------------|---------|------------|------------|------------|------------|-------------|-------------|--------------|-------------------|-------------------|
+| $rwc       | 0       | 3          | 0          | 0          | 0          | 13          | 0           | 13           | 1594323           | 0                 |
+| $rwc       | 0       | 3          | 0          | 0          | 0          | 12          | 0           | 12           | 531441            | 0                 |
+| $rwc       | 0       | 3          | 0          | 0          | 0          | 6           | 0           | 6            | 729               | 0                 |
+| $rwc       | 0       | 3          | 0          | 0          | 0          | 3           | 0           | 3            | 27                | 0                 |
+| $rwc       | 1       | 3          | 0          | 0          | 0          | 2           | 0           | 2            | 9                 | 0                 |
+
+For `exponent == 13`, i.e. scenario #4 we do two lookups:
```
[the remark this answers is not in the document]
roynalnaruto: I've listed the 4 scenarios above:
```
Depending on the value of the exponent within the exponentiation operation, the EXP gadget will either:
```
onwards. I'll mark them appropriately so the scenarios become more evident.

specs/opcode/0AEXP.md
```diff
@@ -0,0 +1,59 @@
...
+    - stack_pointer += 1
+    - pc += 1
+    - gas -= static_gas + dynamic_gas, where:
+        - `static_gas = 10`
+        - `dynamic_gas = 50 * byte_size(exponent)`
+3. Lookups:
+    - `base` is at the top of the stack (Read from RW Table)
+    - `exponent` is at the second position of the stack (Read from RW Table)
+    - `exponentiation` is at the top of the stack (Write to RW Table)
+    - if `exponent == 2`:
```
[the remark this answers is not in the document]
roynalnaruto: There are no lookups in case `exponent == 0` or `exponent == 1`. In those cases, we validate that:
1. If `exponent == 0` then `exponentiation == 1`
2. If `exponent == 1` then `exponentiation == base`

Because there are no lookups, I did not mention in the gadget specs. It is mentioned in the `tables.md` though. Will add it here just to make it more verbose.

src/zkevm_specs/exp_circuit.py
```diff
@@ -0,0 +1,103 @@
...
+        # is_last == 0 is following by is_pad == 0.
+        cs.constrain_equal(rows[0].is_last, rows[1].is_pad)
+        # multiplication is assigned correctly
+        _overflow, carry_lo_hi, additional_constraints = mul_add_words(
+            rows[0].a, rows[0].b, rows[0].c, rows[0].d
+        )
+        cs.range_check(carry_lo_hi[0], 9)
+        cs.range_check(carry_lo_hi[1], 9)
+        cs.constrain_equal(additional_constraints[0][0], additional_constraints[0][1])
+        cs.constrain_equal(additional_constraints[1][0], additional_constraints[1][1])
```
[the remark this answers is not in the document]
roynalnaruto: We do the `mul_add_words` operation which returns 2 additional constraints that need to hold.

The util method simply returns those constraints, then I use the constraint system to validate those constraints:
1. `mul_add_words`: https://github.com/scroll-tech/zkevm-specs/blob/feat/exp/src/zkevm_specs/util/arithmetic.py#L135
2. additional constraints: https://github.com/scroll-tech/zkevm-specs/blob/feat/exp/src/zkevm_specs/util/arithmetic.py#L161-L162

For simplicity, I could have used `assert` in the util itself, but then in case of failure it wouldn't catch and report it appropriately (this part is managed by the [constraint system](https://github.com/scroll-tech/zkevm-specs/blob/feat/exp/src/zkevm_specs/util/constraint_system.py))

specs/tables.md
roynalnaruto: We wish to validate that the exponentiation trace finally reduces to `exponent == 2`, i.e. `a * a == a^2 (mod 2^256)`.

In the exponentiation circuit, we can only verify that:
```python
if is_last is 1 then assert exponent == 2
```

But we cannot ensure that such a row occurs. In order to ensure that this row occurs, we do a lookup from the `EXP` gadget.

specs/exp-proof.md
roynalnaruto: If `exponent` is odd, then we won't have `exponent_lo == 0` because
```
exponent == exponent_lo + (2^128 * exponent_hi)
```

And if the exponent is only reducing by 1, i.e. `exponent::next == exponent::cur - 1` then we can say that:
```
(exponent_hi::next == exponent_hi::cur) && (exponent_lo::next == exponent_lo::cur - 1)
```

Is there a case where you can find an anomaly?

roynalnaruto:
@ChihChengLiang I've answered all your questions and updated the documents/specs. Could you review again?

ChihChengLiang:
rereviewing now. Thanks for waiting.

specs/exp-proof.md
ChihChengLiang: That makes perfect sense now!

specs/exp-proof.md
ChihChengLiang: I saw the design doc and it makes sense now.

specs/tables.md
ChihChengLiang: makes sense. Thanks for the explanation.

specs/tables.md
ChihChengLiang: Reverse order sounds great to me.

specs/exp-proof.md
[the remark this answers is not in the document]
ChihChengLiang: We can now reference the doc to the spec path.
```suggestion
For a detailed algorithm, please refer the [design document](specs/exponentiation-circuit.md).
```

specs/exponentiation-circuit.md
```diff
@@ -0,0 +1,254 @@
+# Exponentiation Circuit
```
ChihChengLiang, line 1: Can we indicate this as a design doc in both title and the filename? So that people don't confuse the difference between `spec/exp-proof.md` and `spec/exponentiation-circuit.md`
```suggestion
# Exponentiation Proof Design Document.
```

specs/exp-proof.md
```diff
@@ -0,0 +1,77 @@
...
+        - `exponent` MUST reduce by 1, which means:
+	    - Low 128 bits of `exponent::next` MUST equal low 128 bits of `exponent::cur - 1`.
+	    - High 128 bits of `exponent::next` MUST equal high 128 bits of `exponent::cur`.
+	- `exponent` is odd also means it was a multiplication by `base` operation, that is:
+	    - For each limb, `exp_table.base_limb == mul_gadget.b`.
+    - If parity check if `even`, i.e. `parity_check.r_lo == 0`:
+        - `exponent` MUST reduce to `exponent // 2`, which means:
+	    - Low 128 bits of `exponent::cur` MUST equal low 128 bits of `parity_check`'s multiplication result `d`.
+	        - That is, `exp_table.exponent_lo == parity_check.d_lo`.
+	    - High 128 bits of `exponent::cur` MUST equal high 128 bits of `parity_check's multiplication result `d`.
```
ChihChengLiang, line 59: Missing quote
```suggestion
	    - High 128 bits of `exponent::cur` MUST equal high 128 bits of `parity_check`'s multiplication result `d`.
```

specs/exponentiation-circuit.md
```diff
@@ -0,0 +1,254 @@
...
+We make use of a cell called `most_significant_nonzero_byte_inverse` to store the inverse of the byte at the said index.
+
+### Constraints
+
+* Exactly one cell is turned on from the array `most_significant_nonzero_byte_index`.
+* For instance, if the above turned on index is `index'`, then we have:
+```
+sum(word_bytes[index'..32]) == 0
+```
+which basically means that, all the other bytes from `index'` must be `0` (since in the little-endian representation of word, bytes beyong the byte-size will be `0`).
```
ChihChengLiang, line 250:
```suggestion
which basically means that, all the other bytes from `index'` must be `0` (since in the little-endian representation of word, bytes beyond the byte-size will be `0`).
```

ChihChengLiang (approved):
LGTM. I added some nitpicks.

roynalnaruto:
@ChihChengLiang updated the PR with all suggestions. Shall we merge it now?

PR #258, opened 2022-08-24
public_inputs: update tx fields

pinkiebell:
This removes `GasTipCap` and `GasFeeCap` and adds `CallDataCost` to the transactions fields.

related: https://github.com/privacy-scaling-explorations/zkevm-chain/issues/13


src/zkevm_specs/public_inputs.py
```diff
@@ -195,13 +197,23 @@ def tx_table_value_column(self) -> List[FQ]:
         column.append(FQ(self.nonce))  # Nonce
         column.append(FQ(self.gas))  # Gas
         column.append(FQ(self.gas_price))  # GasPrice
-        column.append(FQ(0))  # GasTipCap
-        column.append(FQ(0))  # GasFeeCap
         column.append(FQ(self.from_addr))  # CallerAddress
         column.append(FQ(self.to_addr))  # CalleeAddress
+        # FIXME: no difference in the case `tx.to = None` and `tx.to = 0x0000000000000000000000000000000000000000`
```
ed255, line 202: table-wise we handle the case `tx.to = None` by setting the `IsCreate = 1` and `CalleeAddress = 0x0000000000000000000000000000000000000000 or ignore`.
But thanks for bringing this up because I had the impression that having `tx.to = None` was the same as `tx.to = 0x0000000000000000000000000000000000000000`, but I checked how the RLP encoding of a tx appears and those two cases are different!  So I think we need to make sure that tall tx types either have a `is_create` bool flag, or that the `to` type can be None.

ed255 (approved):
LGTM! Thanks for updating the discrepancy!

I see that the CI lint check is not passing.  You can fix that by running `make fmt` from the root.

src/zkevm_specs/public_inputs.py
pinkiebell: I addressed that in the latest commit. Note that this changes the table values. Prior to this change, the padding of MAX_TXS always resulted in `is_create = 1` for `Transaction.default()`, now it is zero. I guess this was intended anyway?

ed255: Thanks for the update!  Yeah having 0s in the fields for padding is commonly the intended approach, so thanks for fixing that too :)

PR #274, opened 2022-09-15
Spec for opcode `RETURNDATACOPY`

smtmfft:


src/zkevm_specs/evm/execution/returndatacopy.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+    memory_offset, size = instruction.memory_offset_and_length(memory_offset_word, size_word)
+    next_memory_size, memory_expansion_gas_cost = instruction.memory_expansion_dynamic_length(
+        memory_offset, size
+    )
+    gas_cost = instruction.memory_copier_gas_cost(size, memory_expansion_gas_cost)
+
+    copy_rwc_inc, _ = instruction.copy_lookup(
+        instruction.curr.call_id,
```
han0110, line 28: It should copy from callee's memory instead of itself, and that's the main reason `LastCalleeId` is also stored in call context.

```suggestion
        instruction.call_context_lookup(CallContextFieldTag.LastCalleeId),
```

src/zkevm_specs/evm/execution/returndatacopy.py
```diff
@@ -0,0 +1,47 @@
...
+        instruction.stack_pop(),
+    )
+
+    return_data_length = instruction.call_context_lookup(
+        CallContextFieldTag.LastCalleeReturnDataLength, RW.Read
+    )
+    return_data_offset = instruction.call_context_lookup(
+        CallContextFieldTag.LastCalleeReturnDataOffset, RW.Read
+    )
+
```
han0110, line 20: We need to make sure `offset_word + size_word <= return_data_length` here, Copy circuit doesn't know where is the end of return data.

tests/evm/test_returndatacopy.py
```diff
@@ -0,0 +1,158 @@
...
+    src_data = dict(
+        [
+            (i, CALLEE_MEMORY[i] if i < len(CALLEE_MEMORY) else 0)
+            for i in range(return_data_offset, return_data_offset + return_data_length)
+        ]
+    )
+    copy_circuit = CopyCircuit().copy(
+        randomness,
+        rw_dictionary,
+        CALL_ID,
```
han0110, line 115: Need to create a mock `CALLEE_ID` for the test.

```suggestion
        CALLEE_ID,
```

han0110 (commented):
There seems to be some missing checks, otherwise it looks good!

src/zkevm_specs/evm/execution/returndatacopy.py
smtmfft: Hi, Han! Thanks for review! I chose call_id because I thought the data was already copied by opcode `return` to current call's memory space, correct me if I am wrong.

src/zkevm_specs/evm/execution/returndatacopy.py
smtmfft: That one is missed, circuit did that but spec doesn't,  will fix soon.

src/zkevm_specs/evm/execution/returndatacopy.py
han0110: Not really, so `RETURN` doesn't actually copy data, it's `*CALL*` that could be specified to copy return data to certain memory chunk (e.g. https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L699-L701).

However, the return data is still immutable and accessible until next call/create, and that's the actual source of `RETURNDATACOPY` (the [`interpreter.returnData`](https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L340)).

smtmfft: Got it. will fix soon! Also will fix retundatacopy circuit accordingly later.

han0110 (approved):
LGTM! Nice work!

ChihChengLiang:
Merge for getting two approvals.

PR #276, opened 2022-09-16
CodeHashProof changed to CodeHashMod

miha-stopar:
MPT circuit should support code hash modifications, not just proving that an account has a particular code hash.

By @ed255 :
I think the case where CodeHash goes from default to a non-default value should be supported.  And further more, I think this case can happen when the account leaf already exists.  Here's an example
- Someone precomputes the address that a contract with CREATE2 will have, and sends some balance to that address.  Now we have a new leaf account with balance > 0, and codeHash = default.
- Then CREATE2 is called which will need to update the codeHash of the already existing leaf from default to non-default value


ed255 (approved):
LGTM! Thanks for updating the spec :)

ChihChengLiang (approved):
LGTM

Issue #252, opened 2022-08-19
Implement opcode `STATICCALL` in Spec


silathdiir:
Please help assign this issue to me if no one is working on this opcode. Thanks.

Related Circuit issue: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/699

PR #281, opened 2022-09-23
Spec for opcode `STATICCALL`

silathdiir:
Close https://github.com/privacy-scaling-explorations/zkevm-specs/issues/252

Circuit: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/801


DreamWuGit (approved):
LGTM

src/zkevm_specs/evm/execution/call_staticcall.py
```diff
@@ -173,7 +197,11 @@ def call(instruction: Instruction):
             (CallContextFieldTag.CallDataLength, cd_length),
             (CallContextFieldTag.ReturnDataOffset, rd_offset),
             (CallContextFieldTag.ReturnDataLength, rd_length),
-            (CallContextFieldTag.Value, value.expr()),
+        ]
+        # Value is not used for opcode STATICCALL.
+        if is_call == 1:
+            call_context_lookups.append((CallContextFieldTag.Value, value.expr()))
```
han0110, line 203: We still need to initialize the `CallContextFieldTag.Value` to `0` in for next call when it's `STATICCALL`, otherwise a malicious prover can put any value there.

src/zkevm_specs/evm/execution/call_staticcall.py
```diff
@@ -78,6 +86,9 @@ def call(instruction: Instruction):
     has_value = 1 - instruction.is_zero(value)
     instruction.constrain_zero(has_value * is_static)
 
+    # Constrain value == 0 for opcode STATICCALL.
+    instruction.constrain_zero(has_value * (1 - is_call))
```
han0110, line 90: It seems line 87 is already doing this.

tests/evm/test_call_staticcall.py
```diff
@@ -181,31 +213,42 @@ def test_call(
...
     # fmt: off
     rw_dictionary = (
-        RWDictionary(24)
+        RWDictionary(rw_counter)
         .call_context_read(1, CallContextFieldTag.TxId, 1)
         .call_context_read(1, CallContextFieldTag.RwCounterEndOfReversion, caller_ctx.rw_counter_end_of_reversion)
         .call_context_read(1, CallContextFieldTag.IsPersistent, caller_ctx.is_persistent)
         .call_context_read(1, CallContextFieldTag.CalleeAddress, caller.address)
-        .call_context_read(1, CallContextFieldTag.IsStatic, False)
+        .call_context_read(1, CallContextFieldTag.IsStatic, is_static)
```
han0110, line 227: This `is_static` is actually the context of current call, so it's not always equal to `opcode == STATICCALL`. Would you like to add an extra test matrix `is_static = [True, False]` in `gen_testing_data`?

han0110 (commented):
Overall the modification of `CALL` to also support `STATICCALL` looks good, thanks for the work! I have also added some comments, please take a look.

src/zkevm_specs/evm/execution/call_staticcall.py
silathdiir: Fixed to initialize `CallContextFieldTag.Value` to `0` for opcode `STATICCALL`.

src/zkevm_specs/evm/execution/call_staticcall.py
silathdiir: Deleted this constraint and reserve `constrain_zero(has_value * is_static)` above.

tests/evm/test_call_staticcall.py
silathdiir: It seems that `is_static` of current call is useless and could not be constrained.

I deleted this lookup (seems unnecessary to add `is_statics = [True, False]` in `gen_test_data`?), and add a lookup for `is_static` from `callee` and constrain it as:
```
# Get callee is_static and Verify is_static == 0 for opcode CALL, and is_static == 1 for opcode STATICCALL.
is_static = instruction.call_context_lookup(
  CallContextFieldTag.IsStatic, call_id=callee_call_id
)
instruction.constrain_equal(is_call + is_static, FQ(1))
```
And updated test code.

@han0110 Could you help review if it is correct when you have time? Thanks a lot.

DreamWuGit: @han0110 if current call is static call, still need to constrain its subcall is also static?

han0110: > if current call is static call, still need to constrain its subcall is also static?

yes, so we should initialize `is_static = current_call_context.is_static || opcode = STATICCALL`, and use it to check with `has_value` and constrain the next call context to have `is_static`.

silathdiir: Fixed to initialize and constrain `is_static` as:
```
is_static = instruction.select(
    is_call, instruction.call_context_lookup(CallContextFieldTag.IsStatic), FQ(1)
)

...

# Verify is_static == 1 for opcode STATICCALL.
instruction.constrain_zero((1 - is_call) * (1 - is_static.expr()))
```

In test cases, set `is_static` as:
```
is_call = opcode == Opcode.CALL
is_static = not is_call or stack.value == 0
```

src/zkevm_specs/evm/execution/call_staticcall.py
```diff
@@ -1,37 +1,47 @@
...
     reversion_info = instruction.reversion_info()
     caller_address = instruction.call_context_lookup(CallContextFieldTag.CalleeAddress)
-    is_static = instruction.call_context_lookup(CallContextFieldTag.IsStatic)
+    is_static = instruction.select(
+        is_call, instruction.call_context_lookup(CallContextFieldTag.IsStatic), FQ(1)
+    )
     depth = instruction.call_context_lookup(CallContextFieldTag.Depth)
 
+    # Verify is_static == 1 for opcode STATICCALL.
+    instruction.constrain_zero((1 - is_call) * (1 - is_static.expr()))
```
han0110, line 35: I think this is not necessary, since in line 21 `instruction.pair_select` already constrains so (see also circuit implementation https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/util/math_gadget.rs#L590-L606).

```suggestion
```

han0110 (approved):
Just another nit pick, otherwise we can merge

PR #283, opened 2022-09-26
Reorganize RETURN specs

z2trillion:


specs/opcode/F3RETURN.md
```diff
@@ -75,16 +82,16 @@ expansion, the gas cost is 0.
 1. opId - 0xF3
 2. State transition:
     - rwc and `rw_table` lookups:
-        - Fixed: 1 (`call_context` lookup) + 2 (stack lookup)
+        - Fixed: 3 (`call_context` lookups) + 2 (stack lookup)
```
ed255, line 85: If I understand this correctly, the updates on the lookup numbers match the implementation in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/719 right?
But that also means that the python implementation at https://github.com/scroll-tech/zkevm-specs/blob/feat%2Freturn/src/zkevm_specs/evm/execution/return_.py differs from these numbers?

ed255 (commented):
Overall looks good! Thanks for the update.  Nevertheless I have a question about the python part.

specs/opcode/F3RETURN.md
z2trillion: the number of lookups now isn't changed, because I am using `is_create`, `is_root`, and `call_id` from the constraint builder step.

ed255 (approved):
LGTM!

Issue #284, opened 2022-09-27
Update EndBlock spec


ed255:
During the implementation of the EndBlock circuit I found some issues that required extending the constraints in order to:
- Support TxPadding with fixed Tag column
- Support empty block
- Support full TxTable

The EndBlock spec requires an update to reflect those new requirements.

Issue #279, opened 2022-09-20
Support for empty blocks


ed255:
One of the responsibilities of the EVM Circuit is to make sure that all the transactions that appear in the TxTable (which match the transactions included in the block) are processed.  In order to do that, the current approach is to take the index of the last processed transaction (once we enter the EndBlock state) and use it to make sure there are no more than those transactions in the TxTable (by lookup-querying the padding rows in the TxTable).

But what happens if a block doesn't have any transaction?  In that case we don't have a "last transaction".

From the EVM Circuit, we can detect the case when the lookup-query to padding rows means that all rows are padding.  We could use a IsZero gadget to build a selector that activates the special case of no txs in a block.

ed255:
During the implementation of the EndBlock circuit I've managed to extend the constraints to support the empty block.  I'm working on an update to the spec that matches the circuit I currently have.

Issue #273, opened 2022-09-14
TxTable padding txs approach


ed255:
Currently we have a conflicting spec for the padding areas of the TxTable (this was originally borught up by @davidnevadoc) 

The current TxTable is divided into 2 fixed height parts:
- a. Tx Fields. num_rows = MAX_TXS * NUM_FIELDS_TX
- b. Tx Calldata. num_rows = MAX_CALLDATA

One of the requirements for this table is to be able to get a lower bound on the number of padding txs (out of MAX_TXS, only some of them are active, and the rest are padding, with the signature check disabled).  This is needed so that we can prove that the number of non-padding txs and the number of txs processed in the EVM circuit are the same (so that no tx is skipped).

A simple solution for that (which is currently specified for the EVM circuit) is to define the padding area of part `a` as follows:
- Every row uses Tag = Pad
- TxId starts from 1 and grows sequentially until part `b` is reached. (this is either calculated by the verifier, or constrained in circuit)
This allows verifying that k is a lower bound on the number of padding txs by doing the following lookups:
- `tx_lookup(tx_id = 1, tag = Pad)`
- `tx_lookup(tx_id = k * NUM_FIELDS_TX, tag = Pad)`

Unfortunately this means that the Tag column is not fixed (because it depends on the number of padding txs), which is incompatible with the current spec of the PublicInputs circuit.

It would be good to have a fixed Tag column, while keeping a design that allows verifiying that at last a number of padding txs exist.  In any case, the padding rows should be distinguishable from non-padding rows, so that the EVM circuit never queries a padding row unintentionally.

ed255:
Here's an idea (A):

For padding Txs, leave the tag column filled with all the tags.  A Padding Tx will have the field `CallerAddress = 0`, and the TxID of padding txs will be sequential still (that is, all rows of the same padding tx will have the same tx_id, and the tx_id of each padding tx will continue after the last tx).  The TxCircuit already supports `CallerAddress = 0` for padding (this is what is being used to disable the signature verification).

From the EVM circuit
- 1. At each begin_tx, verify that the `CallerAddress` of the current `tx_id` is `!= 0`.  This guarantees that we're not processing padding txs
[...]

PR #286, opened 2022-09-27
Update EndBlock

ed255:
- Support TxTable padding with fixed Tag column by following approach (A) in https://github.com/privacy-scaling-explorations/zkevm-specs/issues/273 - Update Tx circuit spec accordingly
- Add support for empty block
- Support full TxTable (when there are no padding transactions)

Resolve https://github.com/privacy-scaling-explorations/zkevm-specs/issues/284 Resolve https://github.com/privacy-scaling-explorations/zkevm-specs/issues/279 Resolve https://github.com/privacy-scaling-explorations/zkevm-specs/issues/273

### DIFF
--- a/src/zkevm_specs/evm/execution/end_block.py
+++ b/src/zkevm_specs/evm/execution/end_block.py
@@ -56,7 +56,7 @@ def end_block(instruction: Instruction):
     # Note that rw_counter starts at 1
     is_empty_block = instruction.is_zero(instruction.curr.rw_counter - 1)
     # If the block is not empty, we will do 1 call_context lookup
-    total_rws = is_empty_block * 0 + (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 1)
+    total_rws = (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 1)
 
     if instruction.is_last_step:
         # 1. Constraint total_txs witness values depending on the empty block case.
@@ -96,6 +96,8 @@ def end_block(instruction: Instruction):
         # in the rw_table.
         # We conclude that the number of meaningful entries in the rw_table is total_rws.
 
+        # TODO: Send fixed reward to coinbase according to the chain rules (See
+        # https://github.com/privacy-scaling-explorations/zkevm-specs/issues/290)
     else:
         # Propagate rw_counter and call_id all the way down
         instruction.constrain_step_state_transition(
--- a/src/zkevm_specs/evm/execution/end_block.py
+++ b/src/zkevm_specs/evm/execution/end_block.py
@@ -4,23 +4,30 @@ from ..table import CallContextFieldTag, TxTableRow, TxContextFieldTag
 from typing import Set
 
 # EndBlock is an execution state that constraints the following:
-# 1. Once the EndBlock state is reached, there's no other execution states appearing until the end of the EVM Circuit.  In particular, after the first EndBlock, there will be no new lookups to the rw_table.
-# 2. The number of meaningful entries (non-padding) in the rw_table match the rw_counter after the EndBlock state is processed.
-# 3. The number of txs processed by the EVM Circuit match the number of txs in the TxTable
-#    total_tx = instruction.call_context_lookup(CallContextFieldTag.TxId)
-
+# A. Once the EndBlock state is reached, there's no other execution states appearing until the end of the EVM Circuit.  In particular, after the first EndBlock, there will be no new lookups to the rw_table.
+# B. The number of meaningful entries (non-padding) in the rw_table match the rw_counter after the EndBlock state is processed.
+# C. The number of txs processed by the EVM Circuit match the number of txs in the TxTable
+#
 # As an extra point:
-# 4. We need to prove that the EndBlock state exists
-
-# We prove (1) by constraining the transition rule that after an EndBlock
+# D. We need to prove that at least one EndBlock state exists
+#
+# We prove (A) by constraining the transition rule that after an EndBlock
 # state, only an EndBlock state can follow.
 #
-# We prove (2) and (3) by proving that at least `MAX_COUNT - evm_circuit_count`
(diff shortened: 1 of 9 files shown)


src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -4,69 +4,102 @@
...
+            ]
+        )
+    )
+    last_rw_counter = instruction.curr.rw_counter - 1
+
+    total_rws = last_rw_counter
+    if total_txs.n > 0:
+        # If the block is not empty, we will do a call_context lookup
+        total_rws = last_rw_counter + 1
+
```
han0110, line 62: It seems a little bit difficult to understand when `total_rws` depends on `total_txs`, could we try to define `total_rws` in this way?

```python
is_empty_block = is_zero(instruction.curr.rw_counter)
if is_empty_block:
    total_rws = 0
else:
    total_rws = instruction.curr.rw_counter # extra 1 from tx_id lookup
```

Note currently this makes `total_rws` to be always `instruction.curr.rw_counter`, but in the future we might have more lookup in `EndBlock` (for rewarding block proposer), then this if-else is necessary.

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -4,69 +4,102 @@
...
+        if total_rws == 0:
+            # 1a. total_txs is 0 in empty block
+            instruction.constrain_equal(total_txs, FQ(0))
+        else:
+            # 1b. total_txs matches the tx_id that corresponds to the final step.
+            instruction.constrain_equal(
+                instruction.call_context_lookup(CallContextFieldTag.TxId), total_txs
+            )
+            # total_rws is last_rw_counter + 1.  extra 1 from tx_id lookup.
+            instruction.constrain_equal(total_rws, last_rw_counter + 1)
```
han0110, line 74: With `total_rws` defined as my proposal above, this check seems unnecessary.

han0110 (approved):
LGTM! Nice work and the updated document for `EndBlock` is clean and easy to follow!

src/zkevm_specs/evm/execution/end_block.py
ed255: Thanks for the suggestion, I've applied it in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/286/commits/21e6a522ce2e0fbcdfb78759c4526d108fbf94ba

src/zkevm_specs/evm/execution/end_block.py
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/286/commits/21e6a522ce2e0fbcdfb78759c4526d108fbf94ba

DreamWuGit:
one discussion regarding empty block: I see current `coinbase` reward handles in `end_tx`,  but for empty block, there is no begin/end tx,  how to handle the coinbase reward to miner address ? IIUC, `coinbase` is supposed to handled in end_block.

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -4,69 +4,98 @@
...
+#
+# We prove (C) by showing that the transaction ID after the last one processed
+# corresponds to a padding tx in the TxTable, which enforces that once a
+# padding txs appears in the table, the rest will also be padding.  This way,
+# we know that if a padding tx with tx_id exist, then at most there are tx_id-1
+# non-padding txs.  In case we have exhausted the TxTable with txs, we won't
+# have any padding txs; so we skip the lookup.
 #
-# We prove (4) in the circuit implementation by constraining that the last
+# We prove (D) in the circuit implementation by constraining that the last
```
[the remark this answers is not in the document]
ed255: Yes!  The reason is that if we have an empty block, there will be no transactions processed in the EVM circuit, and we need to prove that the only situation in which this can occur is when the TxTable is full of padding txs.  And we do that in the EndBlock.  Otherwise we may have a TxTable with k transactions, the EVM circuit empty (with no step), and all constraints would pass!

ChihChengLiang (approved):
LGTM! Great PR!
I added a nitpick

ed255:
> one discussion regarding empty block: current `coinbase` reward handles in `end_tx`, but for empty block, there is no begin/end tx, how to handle the coinbase reward to miner address ? IIUC, `coinbase` is supposed to handled in end_block.

I checked the logic of rewards to `coinbase` and I see that there are two kinds:
1. for each tx, the gas_used * (gas_price - base_fee) is rewarded to coinbase
2. for each block, a fixed amount (that depends on the hard fork) is rewarded to coinbase

Currently (1) is being done in `EndTx` and I think it should be kept there.  Point (2) should be done in `EndBlock` as you mention.  There are some details that are not yet clear to me, like rewards associated to uncle and nephew blocks (See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1234.md).  So I would prefer to add a TODO and open an issue to address this later.

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -4,69 +4,98 @@
...
+                for tx_row in instruction.tables.tx_table
+                if tx_row.field_tag == TxContextFieldTag.CallerAddress
+                and tx_row.value.expr() != FQ(0)
+            ]
+        )
+    )
+    # Note that rw_counter starts at 1
+    is_empty_block = instruction.is_zero(instruction.curr.rw_counter - 1)
+    # If the block is not empty, we will do 1 call_context lookup
+    total_rws = is_empty_block * 0 + (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 1)
```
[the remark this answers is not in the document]
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/286/commits/01a498c06831e9248f0c28f05d32e2f66038f7cd

PR #298, opened 2022-10-26
Fix bad indentation in the `EXP` spec doc

adria0:
Changed tab for space, fixed bullet indentation


roynalnaruto (approved):
@adria0 Good catch. The same issue can be seen for the `exponent` is `even` case. Let's fix that in this same PR?

PR #300, opened 2022-10-29
fix: make exp gadget compatible with circuits impl

roynalnaruto:
We define a boolean variable `single_step` where:
- `single_step == 1 if exponent == 2`
- `single_step == 0 if exponent > 2`

We don't need any additional constraints on `single_step` (not even `require_boolean`) because it is simply used to do a lookup to the exponentiation circuit. All necessary constraints on that column (`exp_circuit.exp_table.is_last`) are handled within the exponentiation circuit.

This makes the [specs gadget](https://github.com/scroll-tech/zkevm-specs/blob/fix/exp-circuit-compat/src/zkevm_specs/evm/execution/exp.py#L25-L34) compatible with the [circuits gadget](https://github.com/scroll-tech/zkevm-circuits/blob/feat/exp/zkevm-circuits/src/evm_circuit/execution/exp.rs#L130-L167)

@adria0 This is the last pending PR before we can get https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/700 merged

### DIFF
--- a/src/zkevm_specs/evm/execution/exp.py
+++ b/src/zkevm_specs/evm/execution/exp.py
@@ -22,30 +22,23 @@ def exp(instruction: Instruction):
     else:
         base_limbs = instruction.word_to_64s(base_rlc)
         identifier = FQ(instruction.curr.rw_counter + instruction.rw_counter_offset)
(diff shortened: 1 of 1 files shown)


ChihChengLiang (approved):
LGTM

PR #301, opened 2022-10-31
spec for stack overflow & underflow state

DreamWuGit:


specs/error_state/ErrorStackOverflow_Underflow.md
```diff
@@ -0,0 +1,25 @@
...
+2. Otherwise, it restores caller's context and switch to it.
+
+### Circuit behavior 
+1. in order to get each step(op code)'s `min_stack_pointer` & `max_stack_pointer` value, construct new fixed tag
+called `opcode_stack` which holds the data.
+2. look up `opcode_stack` to retrieve current executing step's stack info(`minStack` & `maxStack`) 
+3. combine stack overflow & underflow circuit into one, need to check `is_stack_overflow` and `is_stack_underflow`
+is bool and only one is true at the mean while
+4. common error handling:
+  - If it's a root call, it transits to `EndTx`, and the call's `IsSuccess` must be false
```
ed255, line 20: Shouldn't the `IsSuccess = false` constraint be applied in both cases? (is root and is not root).  Actually in the python part it seems correct.

src/zkevm_specs/evm/opcode.py
```diff
@@ -377,6 +377,13 @@ def stack_underflow_pairs() -> List[Tuple[Opcode, int]]:
     return pairs
 
 
+def stack_pairs() -> List[Tuple[Opcode, int, int]]:
```
ed255, line 380: This function returns a triplet instead of a pair, so maybe a different name would be more appropriate?

src/zkevm_specs/evm/table.py
```diff
@@ -77,6 +78,13 @@ def table_assignments(self) -> List[FixedTableRow]:
                 FixedTableRow(FQ(self), FQ(code[0]), FQ(code[1]), FQ(0))
                 for code in constant_gas_cost_pairs()
             ]
+        elif self == FixedTableTag.OpcodeStack:
+            return [
+                # flag | op code |  min stack | max stack
+                FixedTableRow(FQ(self), FQ(pair[0]), FQ(pair[1] if pair[1] > 0 else 0), FQ(pair[2]))
```
ed255, line 84: I checked the values from `src/zkevm_specs/evm/opcode.py` in the `OPCODE_INFO_MAP` and I think I don't understand them.
https://github.com/privacy-scaling-explorations/zkevm-specs/blob/d4e50f011f7a6d6d60c24449df75402490f247f1/src/zkevm_specs/evm/opcode.py#L209-L211
According to the spec written in this PR at `specs/error_state/ErrorStackOverflow_Underflow.md` I believe that for `ADD` it should be `min_stack = 2, max_stack = 1024`.

src/zkevm_specs/evm/execution/error_stack.py
```diff
@@ -0,0 +1,55 @@
...
+from ..table import CallContextFieldTag
+from ..execution_state import ExecutionState
+from ..opcode import Opcode
+from ...util import N_BYTES_STACK
+
+
+def stack_error(instruction: Instruction):
+    # retrieve op code associated to stack error
+    opcode = instruction.opcode_lookup(True)
+    # TODO: lookup min or max stack pointer
```
ed255, line 12: Is this a leftover comment?

tests/evm/test_error_stack.py
```diff
@@ -0,0 +1,159 @@
...
+    Block,
+    Transaction,
+    Bytecode,
+    RWDictionary,
+)
+from zkevm_specs.util import rand_fq, RLC
+
+BYTECODE = Bytecode().pop()
+BYTECODE_PUSH = Bytecode().push1(0x10).push1(0x20)
+
```
ed255, line 20: Could you add another test vector that uses `ADD`? (or any other opcode that has more pops than pushes to the stack)

src/zkevm_specs/evm/table.py
ed255: Ok I figured it out.  I was thinking: empty stack -> stack_pointer = 0.  But we're handling the stack like it grows backwards, so empty stack -> stack_pointer = 1024.
It feels strange because when I think `min_stack...` I think about the minimum number of elements that should be in the stack to avoid a stack underflow.

ed255 (commented):
Overall looks good!  I initially got confused because I forgot that we handle the stack growing backwards.  I added a few notes, the main request is to add a test for a stack underflow with `ADD`.

specs/error_state/ErrorStackOverflow_Underflow.md
DreamWuGit: yes, both cases are false, will move it up line, thanks for catching..

src/zkevm_specs/evm/table.py
DreamWuGit: `min_stack` means smallest stack pointer value , for `Opcode.ADD: OpcodeInfo(-1, 1022, GAS_COST_FASTEST)` the negative one (-1) means stack overflow can not happens. in the circuit I just set to 0 (https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/873/files#diff-a2fce07a0cbe579c2eabd385419d9c771a81069ccd4fa7c5d593d175a1f33061) , because stack pointer never less than 0, which ensure stack overflow not happen for ADD.

src/zkevm_specs/evm/execution/error_stack.py
DreamWuGit: yes, will remove it.

specs/error_state/ErrorStackOverflow_Underflow.md
DreamWuGit: updated in [92d101b](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/301/commits/92d101b7ba811f7e71a441ad9272c7cd967ccf6f) thanks!

src/zkevm_specs/evm/execution/error_stack.py
DreamWuGit: remove TODO in [92d101b](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/301/commits/92d101b7ba811f7e71a441ad9272c7cd967ccf6f) thanks !

tests/evm/test_error_stack.py
DreamWuGit: add `ADD` in [e47e611](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/301/commits/e47e611ba0917d0e60b5ddc43db2742f89541b3b) thanks !

src/zkevm_specs/evm/opcode.py
DreamWuGit: renamed to `stack_bounds` .

ed255 (approved):
LGTM! Thanks for addressing the comments :)

ed255:
@silathdiir could you do a second review to this PR before we merge it please?

silathdiir (approved):

silathdiir:
> @silathdiir could you do a second review to this PR before we merge it please?

Sorry for missing this message. Just reviewed and approved.

PR #303, opened 2022-11-03
Update `BeginTx`

han0110:
This PR aims to synchronize `BeginTx` to current circuit implementation with an extra `IsSuccess` call context write lookup.

It also fix `test_public_inputs.py` to have correct type and remove duplicated test (not sure why my `make test` fails with the wrong type definition but CI doesn't).


ed255 (approved):
LGTM!

DreamWuGit (approved):
LGTM

PR #307, opened 2022-11-08
Add `SuperCircuit` spec

davidnevadoc:
Add explanation of the Super Circuit aggregation strategy.


specs/super_circuit.md
```diff
@@ -0,0 +1,39 @@
...
+ - [x] RW Table
+ - [x] MPT Table
+ - [x] Keccak Table
+ - [x] Tx Table
+ - [x] Bytecode Table
+ - [x] Copy Table
+ - [x] Block Table
+
+The following diagram shows the relation between this circuits and tables:
+![](./super_circuit.png)
```
ed255, line 29: - Missing lookup from Copy Circuit to RW Table
- I think the Copy Circuit doesn't do lookups to the Keccak Table
- Since you added the future Block Circuit, I think the diagram will be more accurate if the EVM Circuit does a lookup to the Block Table (instead of contains)

ed255 (commented):
Looks great!!! I've added a note about the diagram with some points :)

leolara:
Perhaps at the diragram, when it says commit, ti shoudl say commitment?

ed255 (approved):
LGTM!  The diagram is really nice!

specs/super_circuit.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Note**: There are some smaller sub-circuits, like the ECDSA circuit and opcode related
```
DreamWuGit, line 31: I think the tx circuit should look up signatures in ecdsa circuit table to verify signature's correctness.

specs/super_circuit.md
```diff
@@ -0,0 +1,39 @@
...
+sub-circuits do not interact with other tables and circuits in a way that affects the Super
+Circuits, so they are not shown in the diagram for clarity.
+
+Putting together all sub-circuits in one allows them to easily make lookups to whatever table is
+necessary without the need of having them copied. The main drawback is that the Super Circuit proof
+is very large, and therefore its verification its costly. To alleviate this problem a new circuit is
+introduced: the Root Circuit, whose function is to verify the Super Circuit proof, generating a much
```
DreamWuGit, line 38: I am not clear about the Root circuit. is it verification circuit of recursion ? aggregate multi super circuit?

DreamWuGit (approved):
LGTM &Leave two comments !

specs/super_circuit.md
ed255: Exactly; currently the TxCircuit uses halo2wrong to verify correct ECDSA signatures, and the connection is made via copy constraints.

AFAIK the only signature verification needed are for the transactions and for the `ecrecover` precompile.  Once we implement the precompile we may have an ECDSA signature verification circuit that exposes a table where other circuits can look up; but it's not clear yet if we'll choose this direction.  A detail about `ecrecover` that is different than the signature verification of transactions is that it must support invalid signatures.

ChihChengLiang:
@davidnevadoc is this ready to merge?

PR #315, opened 2022-11-17
fix a link

zhiqiangxu:


ChihChengLiang:
Thanks for the fix @zhiqiangxu! I found it needs ../ to work and fix it with #316.

zhiqiangxu:
> Thanks for the fix @zhiqiangxu! I found it needs ../ to work and fix it with #316.

I think both work.

PR #337, opened 2022-12-08
Add cumulative gas  check in `EndBlock`

### LINKED ISSUE #207: constrain cumulativeGasUsed not exceed gas limit in block
since we have `cumulativeGasUsed` data per tx in RW table now for receipt importing,   we can constrain `cumulativeGasUsed` of all txs  not exceed gas limit in the block, proper place to handle is in `end_block` gadget !

davidnevadoc:
Close #207

### DIFF
--- a/src/zkevm_specs/evm/execution/end_block.py
+++ b/src/zkevm_specs/evm/execution/end_block.py
@@ -74,7 +74,7 @@ def end_block(instruction: Instruction):
     # Note that rw_counter starts at 1
     is_empty_block = instruction.is_zero(instruction.curr.rw_counter - 1)
     # If the block is not empty, we will do 1 call_context lookup
-    total_rws = (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 4)
+    total_rws = (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 2)
 
     if instruction.is_last_step:
         # 1. Constraint total_txs witness values depending on the empty block case.
@@ -91,8 +91,6 @@ def end_block(instruction: Instruction):
             cumulative_gas = instruction.tx_receipt_read(
                 total_txs,
                 TxReceiptFieldTag.CumulativeGasUsed,
-                # +1 to rw_counter to skip lookup to PostStateOrStatus
-                instruction.curr.rw_counter + instruction.rw_counter_offset + 1,
             )
             limit_exceeded, _ = instruction.compare(gas_limit, cumulative_gas, N_BYTES_GAS)
             instruction.constrain_equal(limit_exceeded, FQ(0))
--- a/tests/evm/test_end_block.py
+++ b/tests/evm/test_end_block.py
@@ -59,20 +59,10 @@ def test_end_block(
                     value=FQ(tx.id),
                 )
             )
-            # append receipt
+            # append CumlativeGasUsed
             rw_rows.append(
                 RWTableRow(
                     FQ(23),
-                    FQ(RW.Read),
-                    FQ(RWTableTag.TxReceipt),
-                    FQ(0),
-                    FQ(TxReceiptFieldTag.PostStateOrStatus),
-                    value=FQ(0),
-                )
-            )
-            rw_rows.append(
-                RWTableRow(
-                    FQ(24),
                     FQ(RW.Read),
                     key0=FQ(RWTableTag.TxReceipt),
                     key1=FQ(tx.id),
@@ -80,16 +70,6 @@ def test_end_block(
                     key3=FQ(TxReceiptFieldTag.CumulativeGasUsed),
                     key4=FQ(0),
                     value=FQ(cumulative_gas),
-                )
-            )
-            rw_rows.append(
-                RWTableRow(
-                    FQ(25),
-                    FQ(RW.Read),
-                    FQ(RWTableTag.TxReceipt),
-                    FQ(0),
-                    FQ(TxReceiptFieldTag.LogLength),
-                    value=FQ(0),
                 )
             )
 


src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -1,12 +1,24 @@
...
-# C. The number of txs processed by the EVM Circuit match the number of txs in the TxTable
+# A. Once the EndBlock state is reached, there's no other execution states
+# appearing until the end of the EVM Circuit.  In particular, after the first
+# EndBlock, there will be no new lookups to the rw_table.
+#
+# B. The number of meaningful entries (non-padding) in the rw_table match the
+# rw_counter after the EndBlock state is processed.
+#
+# C. The number of txs processed by the EVM Circuit match the number of txs in
+# the TxTable
```
ed255, line 21: Since this comment serves as spec/doc for this execution state, could you add a point `D` mentioning the gas check?

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -59,6 +71,7 @@ def end_block(instruction: Instruction):
     total_rws = (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 1)
 
     if instruction.is_last_step:
+        # tx_id = instruction.call_context_lookup(CallContextFieldTag.TxId)
```
ed255, line 74: `tx_id` should be `total_txs` right?  So this line is unnecessary?

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -68,6 +81,16 @@ def end_block(instruction: Instruction):
             instruction.constrain_equal(
                 instruction.call_context_lookup(CallContextFieldTag.TxId), total_txs
             )
+            # 4. Verify that CumulativeGasUsed does not exceed the block gas limit.
+            gas_limit = instruction.block_context_lookup(BlockContextFieldTag.GasLimit)
+            cumulative_gas = instruction.tx_receipt_read(
+                total_txs,
+                TxReceiptFieldTag.CumulativeGasUsed,
+                # +1 to rw_counter to skip lookup to PostStateOrStatus
+                instruction.curr.rw_counter + instruction.rw_counter_offset + 1,
```
ed255, line 90: I don't follow this value.  I think we can pass `None` here so that the logic is of `Instruction.rw_lookup` applied:
```python
            rw_counter = self.curr.rw_counter + self.rw_counter_offset
            self.rw_counter_offset += 1
```

The EndBlock performs a list of lookups to rw_table which correspond to some new rw operations (that will have contiguous rw_counters).

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -80,17 +103,16 @@ def end_block(instruction: Instruction):
...
+            # uses a unique tx_id, we know that at least there are total_tx
+            # meaningful txs in the tx_table. We conclude that the number of
             # meaningful txs in the tx_table is total_tx.
 
         # 3. Verify rw_counter counts to the same number of meaningful rows in
         # rw_table to ensure there is no malicious insertion.
         # Verify that there are at most total_rws meaningful entries in the rw_table
         instruction.rw_table_start_lookup(FQ(1))
-        instruction.rw_table_start_lookup(max_rws - total_rws)
+        # instruction.rw_table_start_lookup(max_rws - total_rws)
```
ed255, line 115: Why is this commented?

src/zkevm_specs/evm/execution/end_block.py
```diff
@@ -59,6 +71,7 @@ def end_block(instruction: Instruction):
     total_rws = (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 1)
```
ed255, line 71:
```suggestion
    total_rws = (1 - is_empty_block) * (instruction.curr.rw_counter - 1 + 2)
```
Since now we're doing an extra rw_lookup (for the TxReceiptFieldTag.CumulativeGasUsed) I think this should be updated.  And the comment as well, it should mention "1 call_context and 1 tx_receipt lookup".

tests/evm/test_end_block.py
```diff
@@ -52,6 +59,40 @@ def test_end_block(is_last_step: bool, empty_block: bool, max_txs: int):
                     value=FQ(tx.id),
                 )
             )
+            # append receipt
```
ed255, line 62: Ah, I now see why you added a `+1` to the rw_counter in the `CumulativeGasUsed` lookup.
If I understand it correctly, you're trying to reuse the rw entries from `end_tx` in the `end_block` right?
If that's the case, this won't always work because we may have an empty block which doesn't contain the `end_tx` state.
So the `end_block` step should have their own rw entries, and they should only be 2: `RWTableTag.CallContext` and `TxReceiptFieldTag.CumulativeGasUsed`.

han0110: I also think only reading `TxReceiptFieldTag.CumulativeGasUsed` should work (so it adds its own RwTable entry). But we might need to additionally check whether the value is zero if it's an empty block, since there is no constraint on the `CumulativeGasUsed` value if it's the first read (the value seems not important now, but when someday we need to compute the block hash in circuit, we'd need to connect this `CumulativeGasUsed` value to block header field somewhere).

han0110 (commented):
Overall looks good to me! Got the similar thought as Edu's and left some further comments.

tests/evm/test_end_block.py
ed255: > I think in the case of having an empty block it will still work because the check is only done for non-empty blocks. However I think your point is still valid. 

Ah you're right, if the block is empty, not even a single `TxReceiptFieldTag.CumulativeGasUsed` rw operation can be made because there's no previous tx!  But as you say, this is not a problem because the circuit only checks this lookup in the non-empty block case.

> Let me know if I got this right: `EndBlock` should add its own RwTable entry with `CumulativeGasUsed` but not the other 2 `TxReceipt` fields. This adds one extra row to the `RwTable` but removes the need to rely on `end_tx` state right? It also removes that magical `+1` I added earlier.

 correct!

> I also think only reading `TxReceiptFieldTag.CumulativeGasUsed` should work (so it adds its own RwTable entry). But we might need to additionally check whether the value is zero if it's an empty block, since there is no constraint on the `CumulativeGasUsed` value if it's the first read (the value seems not important now, but when someday we need to compute the block hash in circuit, we'd need to connect this `CumulativeGasUsed` value to block header field somewhere).

Ah, I think this needs to be resolved by:
1. Adding a new entry to the block table: `GasUsed`
2. In EndTx, if block is empty, lookup to block_table with `(GasUsed, 0)`.  If block is not empty, lookup to block_table with `(GasUsed, cumulative_gas)`

But we can address this in a future PR if you want :)

han0110: > Ah, I think this needs to be resolved by:
> 1. Adding a new entry to the block table: `GasUsed`
> 2. In EndTx, if block is empty, lookup to block_table with `(GasUsed, 0)`.  If block is not empty, lookup to block_table with `(GasUsed, cumulative_gas)`
> 
> But we can address this in a future PR if you want :)

Make sense! I think we should not consider this for now and leave it in future PR, just want to remind that we need to check the value to be zero when it's empty block :)

han0110 (approved):
LGTM!

ed255 (approved):
LGTM!

Issue #151, opened 2022-03-18
Refactor Bytecode circuit to be a finite state machine


han0110:
Currently Bytecode circuit is implemented with access to previous row, and with `is_final` indicator as main constraint flag, which seems not that straightforward to think about all kinds of possible branches.

This note proposes another way to implement it, by treating Bytecode circuit as a finite state machine just like EVM circuit, but with only 2 state: `Length` and  `Byte`. The pseudo code looks like:

```python3
Row = namedtuple(
    'Row',
    [
        'tag',
        'length', # renamed from hash_length
        'index',
        'value', # renamed from byte
        'is_code',
        'push_data_size', # renamed from byte_push_size
        'push_data_left',
        'hash',
        'value_rlc', # renamed from hash_rlc
    ],
)

curr, next = Row(), Row()

assert curr.tag in [Length, Byte]

# first row
if q_is_first:
    assert curr.tag == Length

# all except last row
if not q_is_last:

    if curr.tag == Length:
        assert curr.index == 0
        assert curr.value == curr.length

        if next.tag == Byte:
            assert next.length == curr.length
            assert next.index == 0
            assert next.is_code == 1
            assert next.hash == curr.hash
            assert next.value_rlc == 0
            # don't care next.value
            # don't care next.push_data_left

        if next.tag == Length:
            assert curr.length == 0
            assert curr.hash == EMPTY_HASH
            # don't care next.length
            # don't care next.index
            # don't care next.value
            # don't care next.is_code
            # don't care next.push_data_left
            # don't care next.hash
            # don't care next.value_rlc

    if curr.tag == Byte:
        assert push_data_size_table_lookup(curr.value, curr.push_data_size)
        assert curr.is_code == (curr.push_data_left == 0)
        if curr.is_code:
            assert next.push_data_left == curr.push_data_size
        else:
            assert next.push_data_left == curr.push_data_left - 1

        if next.tag == Byte:
            assert next.length == curr.length
            assert next.index == curr.index + 1
            assert next.hash == curr.hash
            assert next.value_rlc == curr.value_rlc * randomness + curr.value
            # don't care next.value

        if next.tag == Length:
            assert curr.index + 1 == curr.length
            assert keccak256_table_lookup(curr.hash, curr.length, curr.value_rlc)
            # don't care next.length
            # don't care next.index
            # don't care next.value
            # don't care next.hash
            # don't care next.value_rlc

# last row
if q_is_last:
[...]

PR #340, opened 2022-12-13
Refactor bytecode circuit

leolara:
Closes https://github.com/privacy-scaling-explorations/zkevm-specs/issues/151

### DIFF
--- a/specs/bytecode-proof.md
+++ b/specs/bytecode-proof.md
@@ -75,7 +75,7 @@ All circuit constraints are based on the current row (`cur`) and the `next` row.
 
 First of all if `cur.q_first` or `cur.q_last` are `1`, then `cur.tag == Header`.
 
-We should have the following contrains based on `cur.tag` and `next.tag` (state transition), for all rows except the last one (`cur.q_last == 1`).
+We should have the following constraint based on `cur.tag` and `next.tag` (state transition), for all rows except the last one (`cur.q_last == 1`).
 
 To enable lookup all `cur.tag == Header` rows should have:
 
--- a/specs/bytecode-proof.md
+++ b/specs/bytecode-proof.md
@@ -95,7 +95,7 @@ else:
     assert next.push_data_left == cur.push_data_left - 1
 ```
 
-This way we make sure is_code and next.push_data_left has the right values.
+This way we make sure is_code and next.push_data_left have the right values.
 
 ### cur.tag == Header and next.tag == Header
 
--- a/specs/bytecode-proof.md
+++ b/specs/bytecode-proof.md
@@ -4,24 +4,29 @@ The bytecode proof helps the EVM proof by making the bytecode (identified by its
 
 ## Circuit Layout
 
-| Column                | Description                                                        |
-| --------------------- | ------------------------------------------------------------------ |
-| `q_first`             | `1` on the first row, else `0`                                     |
-| `q_last`              | `1` on the last row, else `0`                                      |
-| `hash`                | The keccak hash of the bytecode                                    |
-| `tag`                 | Tag indicates whether `value` is a byte or length of the bytecode  |
-| `index`               | The position of the byte in the bytecode                           |
-| `value`               | The byte data for the current position, or length of the bytecode  |
-| `is_code`             | `1` if the byte is code, `0` if the byte is PUSH data              |
-| `push_data_left`      | The number of PUSH data bytes that still follow the current row    |
-| `hash_rlc`            | The accumulator containing the current and previous bytes          |
-| `hash_length`         | The bytecode length                                                |
-| `byte_push_size`      | The number of bytes pushed for the current byte                    |
-| `is_final`            | `1` if the current byte is the last byte of the bytecode, else `0` |
-| `padding`             | `1` if the current row is padding, else `0`                        |
-| `push_data_left_inv`  | The inverse of `push_data_left` (`IsZeroChip` helper)              |
-| `push_table.byte`     | Push Table: A byte value                                           |
-| `push_table.push_size`| Push Table: The number of bytes pushed for this byte as opcode     |
+The collumn `tag` (advice) makes the circuit behave as a state machine, selecting different constrains depending on the current and next row value. The `tag` collumn can have two different values: Header, Byte. A row of `tag==Header` preceeds a series of `tag==Byte` rows that contain a complete bytecode sequence. The row `tag==Header` contains the length of the bytecode and the hash of the bytecode, each `tag==Byte` contains one byte of the bytecode, bytecode hash, its lenght and other for the push data.
+
+
+| Column                | Description                                                         |
+| --------------------- | --------------------------------------------------------------------|
+| `q_first` (fixed)     | `1` on the first row, else `0`                                      |
+| `q_last` (fixed)      | `1` on the last row, else `0`                                       |
+| `hash`                | The keccak hash of the bytecode                                     |
+| `index`               | The position of the byte in the bytecode starting from 0            |
+| `value`               | Value for this row bytecode byte, and the length in Header rows.    |
+| `is_code`             | `1` if the byte is code, `0` if the byte is PUSH data               |
+| `push_data_left`      | The number of PUSH data bytes that still follow the current row     |
+| `value_rlc`           | The accumulator containing the current and previous bytes RLC       |
+| `length`              | The bytecode length, that could be 0 for empty byrecodes and padding|
+| `push_data_size`      | The number of bytes pushed for the current byte                     |
+| `push_table.byte`     | Push Table: A byte value                                            |
+| `push_table.push_size`| Push Table: The number of bytes pushed for this byte as opcode      |
+
+
+After all the bytecodes have been added, the rest of the rows are filled with padding in the form of `tag == Header && length == 0 && value ==0` rows.
+
+Additionally we will need two collumns for IsZeroChip for `length` and `push_data_left`
+
 
 ## Push table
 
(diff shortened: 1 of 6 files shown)


specs/bytecode-proof.md
```diff
@@ -4,24 +4,29 @@ The bytecode proof helps the EVM proof by making the bytecode (identified by its
...
-| `hash_length`         | The bytecode length                                                |
-| `byte_push_size`      | The number of bytes pushed for the current byte                    |
-| `is_final`            | `1` if the current byte is the last byte of the bytecode, else `0` |
-| `padding`             | `1` if the current row is padding, else `0`                        |
-| `push_data_left_inv`  | The inverse of `push_data_left` (`IsZeroChip` helper)              |
-| `push_table.byte`     | Push Table: A byte value                                           |
-| `push_table.push_size`| Push Table: The number of bytes pushed for this byte as opcode     |
+The collumn `tag` (advice) makes the circuit behave as a state machine, selecting different constrains depending on the current and next row value. The `tag` collumn can have two different values: Header, Byte. A row of `tag==Header` preceeds a series of `tag==Byte` rows that contain a complete bytecode sequence. The row `tag==Header` contains the length of the bytecode and the hash of the bytecode, each `tag==Byte` contains one byte of the bytecode, bytecode hash, its lenght and other for the push data.
```
Brechtpd, line 7:
```suggestion
The column `tag` (advice) makes the circuit behave as a state machine, selecting different constraints depending on the current and next row value. The `tag` column can have two different values: Header, Byte. A row of `tag==Header` precedes a series of `tag==Byte` rows that contain a complete bytecode sequence. The row `tag==Header` contains the length of the bytecode and the hash of the bytecode, each `tag==Byte` contains one byte of the bytecode, bytecode hash, its length and other for the push data.
```

specs/bytecode-proof.md
```diff
@@ -4,24 +4,29 @@ The bytecode proof helps the EVM proof by making the bytecode (identified by its
...
+| --------------------- | --------------------------------------------------------------------|
+| `q_first` (fixed)     | `1` on the first row, else `0`                                      |
+| `q_last` (fixed)      | `1` on the last row, else `0`                                       |
+| `hash`                | The keccak hash of the bytecode                                     |
+| `index`               | The position of the byte in the bytecode starting from 0            |
+| `value`               | Value for this row bytecode byte, and the length in Header rows.    |
+| `is_code`             | `1` if the byte is code, `0` if the byte is PUSH data               |
+| `push_data_left`      | The number of PUSH data bytes that still follow the current row     |
+| `value_rlc`           | The accumulator containing the current and previous bytes RLC       |
+| `length`              | The bytecode length, that could be 0 for empty byrecodes and padding|
```
Brechtpd, line 20:
```suggestion
| `length`              | The bytecode length, that could be 0 for empty bytecodes and padding|
```

specs/bytecode-proof.md
```diff
@@ -4,24 +4,29 @@ The bytecode proof helps the EVM proof by making the bytecode (identified by its
...
+| `push_data_size`      | The number of bytes pushed for the current byte                     |
+| `push_table.byte`     | Push Table: A byte value                                            |
+| `push_table.push_size`| Push Table: The number of bytes pushed for this byte as opcode      |
+
+
+After all the bytecodes have been added, the rest of the rows are filled with padding in the form of `tag == Header && length == 0 && value == 0 && hash == EMPTY_HASH` rows.
+
+Additionally we will need two collumns for IsZeroChip for `length` and `push_data_left`
```
Brechtpd, line 28:
```suggestion
Additionally we will need two columns for IsZeroChip for `length` and `push_data_left`
```

specs/bytecode-proof.md
```diff
@@ -35,30 +40,117 @@ Because we do this lookup for each byte, this table is also indirectly used to r
...
+
+```
+first_bytecode.value_rlc := firstbytecode.value
+
+next.value_rlc := cur.value_rlc * r + next.value
+```
+
+For detecting which byte is code and which byte is push data the [Push table](#push-table) is used. This table allows finding out how many bytes an opcode pushes. This is used to set `next.push_data_left` if and only if the current byte is code (the first byte in any bytecode is code).
+
+If a row contains a non-zero value for `push_data_left` we know the current byte is an opcode:
```
Brechtpd, line 62:
```suggestion
If a row contains a zero value for `push_data_left` we know the current byte is an opcode:
```

specs/bytecode-proof.md
```diff
@@ -35,30 +40,117 @@ Because we do this lookup for each byte, this table is also indirectly used to r
...
+first_bytecode.is_code := 1
+cur.is_code := cur.push_data_left == 0
+next.push_data_left := cur.byte_push_size if cur.is_code else cur.push_data_left - 1
+```
+
+The fixed columns `q_first` and `q_last` should be zero for all rows, except the first one where `q_first := 1` and the last one where `q_last := 1`.
+
+## Circuit constrains
+
+All circuit constrains are based on the current row (`cur`) and the `next` row.
```
Brechtpd, line 74:
```suggestion
All circuit constraints are based on the current row (`cur`) and the `next` row.
```

specs/bytecode-proof.md
```diff
@@ -35,30 +40,117 @@ Because we do this lookup for each byte, this table is also indirectly used to r
...
+
+First of all if `cur.q_first` or `cur.q_last` are `1`, then `cur.tag == Header`.
+
+We should have the following contrains based on `cur.tag` and `next.tag` (state transition), for all rows except the last one (`cur.q_last == 1`).
```
Brechtpd, line 78:
```suggestion
We should have the following constraint based on `cur.tag` and `next.tag` (state transition), for all rows except the last one (`cur.q_last == 1`).
```

specs/bytecode-proof.md
```diff
@@ -35,30 +40,117 @@ Because we do this lookup for each byte, this table is also indirectly used to r
...
+```
+assert push_data_size_table_lookup(cur.value, cur.push_data_size)
+assert cur.is_code == (cur.push_data_left == 0)
+if cur.is_code:
+    assert next.push_data_left == cur.push_data_size
+else:
+    assert next.push_data_left == cur.push_data_left - 1
+```
+
+This way we make sure is_code and next.push_data_left has the right values.
```
Brechtpd, line 98:
```suggestion
This way we make sure is_code and next.push_data_left have the right values.
```

src/zkevm_specs/bytecode.py
```diff
@@ -16,104 +16,72 @@ class UnrolledBytecode(NamedTuple):
...
+            else:
+                assert next.push_data_left == cur.push_data_left - 1
+
+            if next.tag == BytecodeFieldTag.Byte:
+                check_bytecode_row_byte_to_byte(cur, next, randomness)
+            if next.tag == BytecodeFieldTag.Header:
+                check_bytecode_row_byte_to_header(cur, keccak_table)
+
+    if cur.q_last == 1:
+        assert cur.tag == BytecodeFieldTag.Header
```
Brechtpd, line 55: This doesn't seem to check that `length` and `hash` are set to the correct empty values (the `to_header` checks only check the current row values, and these checks are disabled on this row), so maybe only rows on not(q_last) be considered valid for lookup?

tests/test_bytecode_circuit.py
```diff
@@ -69,7 +71,7 @@ def test_bytecode_empty():
 
 def test_bytecode_full():
     bytecodes = [unroll(bytes([7] * (2**k - 1)), randomness)]
-    verify(k, bytecodes, randomness, True)
+    verify(k, bytecodes, randomness, False)  # Last row must be tag=Header
```
Brechtpd, line 74: Maybe a more useful test if the test data is modified with the required last row header so the test passes.

tests/test_bytecode_circuit.py
```diff
@@ -85,7 +87,7 @@ def test_bytecode_multiple():
         unroll(bytes([Opcode.ADD, Opcode.PUSH32]), randomness),
         unroll(bytes([Opcode.ADD, Opcode.PUSH32, Opcode.ADD]), randomness),
     ]
-    verify(k, bytecodes, randomness, True)
+    verify(k, bytecodes, randomness, False)  # Push without data must fail
```
Brechtpd, line 90: Should it? I thought this was currently allowed but I could be wrong.

aguzmant103:
Thanks for the review! FYI @leolara

tests/test_bytecode_circuit.py
leolara: Is it allowed? Is this because we should allow invalid bytecode to run in the EVM as it would? The EVM would rise an exception and revert the transaction?

In that case perhaps we need some changes to the spec.

src/zkevm_specs/bytecode.py
leolara: @Brechtpd 

So perhaps we could have at the beginning:

```
if cur.q_first == 1 || cur.q_last== 1:
        assert cur.tag == BytecodeFieldTag.Header
```

And then apply the main rules to every row including last? That way when it loops from last to first it will check that last is empty?

Is this the right approach in your oponion?

Or how do we express that q_last is not valid for lookup?

tests/test_bytecode_circuit.py
leolara: Done

src/zkevm_specs/bytecode.py
Brechtpd: I think the approach you propose here would work perfectly if not for unusable rows. But because we still have unusable rows (because zk is enabled) this wrap around cannot be exploited easily (there are rows with random data in between the first and last rows). I think just having a `q_enabled` column in the circuit that is `1` on all rows with valid data (which with the current code would _exclude_ the `q_last` row) would be the standard way to do it. So I would say the current code works, but just have to be careful to disable lookups on the `q_last` row by disabling `q_enabled` there, so just adding a warning in the spec I think is okay.

tests/test_bytecode_circuit.py
Brechtpd: As far as I know this currently is allowed, future EIPs like EIP-3670 may introduce these kinds of checks but I don't think any checks (except on total bytecode size) are currently being done.

Executing this kind of code will result in a revert (which is totally valid to do), but the bytecode of a smart contract may also just just contain sections that are never used by transactions but still needs to be loaded in.

han0110: I think push without data should be allowed and it's will be padded with `n` zero when EVM really executes this `PUSHn`.

The reason this case fails here is because there will be `next_push_data_left` unassigned in the padding rows, we might need to assign it for the first padding row to pass the `assert next.push_data_left == cur.push_data_size`

specs/bytecode-proof.md
```diff
@@ -35,30 +40,117 @@ Because we do this lookup for each byte, this table is also indirectly used to r
...
+Then it runs over all the bytes of the bytecode in order starting at the byte at position `0`.
+Each following row unrolls a single byte (using `tag = Byte` and `value = the actual byte value`) of the bytecode while also storing its position
 (`index`), the code hash it's part of (`hash`), and if it is code or not
-(`is_code`); using `tag = Byte`
+(`is_code`). Also `push_data_size` is filled to match the push table, and `push_data_left` is computed.
+
+All byte data is accumulated per byte (with one byte per row) into `value_rlc` as follows, where r is a challenge:
+
+```
+first_bytecode.value_rlc := firstbytecode.value
```
han0110, line 55:
```suggestion
first_bytecode.value_rlc := first_bytecode.value
```

han0110 (commented):
The refactor looks really great and codes look well organized! Nice work! Only left 2 comments that might need to be addressed.

tests/test_bytecode_circuit.py
leolara: @han0110 @Brechtpd what I have done (that it seemed more semantic to me but perhaps there is something wrong) is to only assert `assert next.push_data_left ==` when we are in a Byte -> Byte transition, please let me know if there is something wrong with this approach.

src/zkevm_specs/bytecode.py
leolara: @han0110 @Brechtpd as the last thing that Han said is exactly the header to header check, I did that the last row does that check, as that check does not use the next row it should be ok.

leolara:
@han0110 @Brechtpd please check again :-)

tests/test_bytecode_circuit.py
han0110: This approach also looks much better to me!

han0110 (approved):
All LGTM now!

Brechtpd (approved):
LGTM, great work!

src/zkevm_specs/bytecode.py
Brechtpd: > I think it'd be easier to to just make sure the `length = 0` and `hash = EMPTY_HASH` when `q_last`, otherwise it seems we need to add extra computation to the table.

Yeah I guess the lookup table is safe to use without any additional selector as long as we ensure `q_last` is set to `1` at the very last usable row.

leolara:
Thanks @han0110 @Brechtpd @ed255

PR #348, opened 2022-12-24
refacor/CALL and OOG_CALL

KimiWu123:
It's for [CALL & CALL OOG refactor #940](https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/940)

### DIFF
--- a/src/zkevm_specs/evm/execution/callop.py
+++ b/src/zkevm_specs/evm/execution/callop.py
@@ -1,8 +1,4 @@
-from zkevm_specs.evm.util.call_gadget import (
-    common_call_gas_cost,
-    common_call_is_empty_code_hash,
-    common_call_stack_pop,
-)
+from zkevm_specs.evm.util.call_gadget import CallGadget
 from zkevm_specs.util.param import N_BYTES_GAS
 from ...util import (
     FQ,
@@ -43,7 +39,7 @@ def callop(instruction: Instruction):
     # Verify depth is less than 1024
     instruction.range_lookup(depth, 1024)
 
-    call_data = common_call_stack_pop(instruction, FQ(0), is_call + is_callcode)
+    call = CallGadget(instruction, FQ(1), is_call, is_callcode, is_delegatecall)
     # For opcode CALLCODE:
     # - callee_address = caller_address
     #
@@ -52,28 +48,28 @@ def callop(instruction: Instruction):
     # - caller_address = parent_caller_address
     #
     callee_address = instruction.select(
-        is_callcode + is_delegatecall, caller_address, call_data.callee_address
+        is_callcode + is_delegatecall, caller_address, call.callee_address
     )
     caller_address = instruction.select(is_delegatecall, parent_caller_address, caller_address)
 
     # Add `callee_address` to access list
     is_warm_access = instruction.add_account_to_access_list(
-        tx_id, call_data.callee_address, reversion_info
+        tx_id, call.callee_address, reversion_info
     )
 
     # Check not is_static if call has value
-    has_value = 1 - instruction.is_zero(call_data.value)
+    has_value = call.has_value
     instruction.constrain_zero(has_value * is_static)
 
     # Propagate rw_counter_end_of_reversion and is_persistent
     callee_reversion_info = instruction.reversion_info(call_id=callee_call_id)
     instruction.constrain_equal(
         callee_reversion_info.is_persistent,
-        reversion_info.is_persistent * call_data.is_success.expr(),
+        reversion_info.is_persistent * call.is_success.expr(),
     )
-    is_reverted_by_caller = call_data.is_success.expr() == FQ(
-        1
-    ) and reversion_info.is_persistent == FQ(0)
+    is_reverted_by_caller = call.is_success.expr() == FQ(1) and reversion_info.is_persistent == FQ(
+        0
+    )
     if is_reverted_by_caller:
         # Propagate rw_counter_end_of_reversion when callee succeeds but one of callers revert at some point.
         # Note that we subtract it with current caller's reversible_write_counter as callee's endpoint, where caller's
@@ -88,7 +84,7 @@ def callop(instruction: Instruction):
         # transfer to constrain it should be greater than or equal to stack
         # `value`.
         (_, caller_balance), _ = instruction.transfer(
-            caller_address, callee_address, call_data.value, callee_reversion_info
+            caller_address, callee_address, call.value, callee_reversion_info
         )
     elif is_callcode == 1:
         # For CALLCODE opcode, get caller balance to constrain it should be
@@ -99,23 +95,19 @@ def callop(instruction: Instruction):
     # or equal to stack `value`.
     if is_call + is_callcode == 1:
         value_lt_caller_balance, value_eq_caller_balance = instruction.compare_word(
-            call_data.value, caller_balance
+            call.value, caller_balance
         )
         instruction.constrain_zero(1 - value_lt_caller_balance - value_eq_caller_balance)
 
-    callee_exists, is_empty_code_hash, callee_code_hash = common_call_is_empty_code_hash(
-        instruction, call_data.callee_address, FQ(0)
-    )
+    # Load callee account `exists` value from auxilary witness data.
+    callee_exists = instruction.curr.aux_data
 
     # Verify gas cost.
-    gas_cost = common_call_gas_cost(
+    gas_cost = call.gas_cost(
         instruction,
-        has_value,
-        call_data.memory_expansion_gas_cost,
         is_warm_access,
-        FQ(0)
-        if callee_exists == 1
-        else is_call,  # Only CALL opcode could invoke transfer to make empty account into non-empty.
+        callee_exists,
+        is_call,  # Only CALL opcode could invoke transfer to make empty account into non-empty.
     )
     # Apply EIP 150.
     # Note that sufficient gas_left is checked implicitly by constant_divmod.
@@ -123,8 +115,8 @@ def callop(instruction: Instruction):
     one_64th_gas, _ = instruction.constant_divmod(gas_available, FQ(64), N_BYTES_GAS)
     all_but_one_64th_gas = gas_available - one_64th_gas
     callee_gas_left = instruction.select(
-        call_data.is_u64_gas,
-        instruction.min(all_but_one_64th_gas, call_data.gas, N_BYTES_GAS),
+        call.is_u64_gas,
+        instruction.min(all_but_one_64th_gas, call.gas, N_BYTES_GAS),
         all_but_one_64th_gas,
     )
 
@@ -135,9 +127,9 @@ def callop(instruction: Instruction):
         is_precompile, FQ(instruction.next.execution_state in precompile_execution_states())
     )
 
-    if is_empty_code_hash == FQ(1) and is_precompile == FQ(0):
+    if call.is_empty_code_hash == FQ(1) and is_precompile == FQ(0):
         # Make sure call is successful
-        instruction.constrain_equal(call_data.is_success, FQ(1))
+        instruction.constrain_equal(call.is_success, FQ(1))
 
         # Empty return_data
         for (field_tag, expected_value) in [
@@ -162,7 +154,7 @@ def callop(instruction: Instruction):
             program_counter=Transition.delta(1),
             stack_pointer=Transition.delta(stack_pointer_delta),
             gas_left=Transition.delta(has_value * GAS_STIPEND_CALL_WITH_VALUE - gas_cost),
-            memory_size=Transition.to(call_data.next_memory_size),
+            memory_size=Transition.to(call.next_memory_size),
             reversible_write_counter=Transition.delta(3),
             # Always stay same
             call_id=Transition.same(),
@@ -183,7 +175,7 @@ def callop(instruction: Instruction):
                 instruction.curr.stack_pointer + stack_pointer_delta,
             ),
             (CallContextFieldTag.GasLeft, instruction.curr.gas_left - gas_cost - callee_gas_left),
-            (CallContextFieldTag.MemorySize, call_data.next_memory_size),
+            (CallContextFieldTag.MemorySize, call.next_memory_size),
             (
                 CallContextFieldTag.ReversibleWriteCounter,
                 instruction.curr.reversible_write_counter + 1,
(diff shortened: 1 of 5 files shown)


KimiWu123:
@ChihChengLiang , could you help to have a quick review to see if I do it in right direction? Any advice are welcome.

src/zkevm_specs/evm/util/call_gadget.py
```diff
@@ -0,0 +1,127 @@
...
+    callee_address: FQ
+    cd_offset: FQ
+    cd_length: FQ
+    rd_offset: FQ
+    rd_length: FQ
+
+    next_memory_size: FQ
+    memory_expansion_gas_cost: FQ
+
+    def __init__(self, instruction: Instruction, is_oog: FQ, is_call_or_callcode: FQ = FQ(0)):
```
ChihChengLiang, line 36: Instead of a class, is it possible to make the common part of callop and oog_call a function? 

For example,

```
def call_common(instruction: Instruction):
    ...
```
It would be okay to define a struct to hold the output value.

```
class CallStruct:
    is_oog: FQ
    is_success: RLC
    value: RLC
    gas: FQ
    callee_address: FQ
    cd_offset: FQ
    cd_length: FQ
    rd_offset: FQ
    rd_length: FQ
    next_memory_size: FQ
    memory_expansion_gas_cost: FQ
```

The reason is that the Python code serves as the specification of the circuit. Organizing the constraints in the class and methods makes the reader hard to keep track of variables.

ChihChengLiang (commented):
Great work. I added a suggestion on organizing the code in a function.

src/zkevm_specs/evm/execution/callop.py
```diff
@@ -44,57 +43,37 @@ def callop(instruction: Instruction):
...
-    rd_length_rlc = instruction.stack_pop()
-    is_success = instruction.stack_push()
-
-    # Verify is_success is a bool
-    instruction.constrain_bool(is_success)
-
-    # Recomposition of random linear combination to integer
-    code_address = instruction.rlc_to_fq(code_address_rlc, N_BYTES_ACCOUNT_ADDRESS)
-
+    call_data = common_call_stack_pop(instruction, FQ(0), is_call + is_callcode)
```
ChihChengLiang, line 46: The name `call_data` collides with the EVM `calldata`.
What do you think we name it `call` directly here? It would make variable names like `call.callee_address` and `call.value` look natural.

KimiWu123:
**Update** 
After dicussed with @ChihChengLiang , we decided to refactor `zkEVM-circuit` first and then back here to see if we need more modification. The reason behind is that we would like to make the code in `zkEVM-spec` is more consistent with `zkEVM-circuit`

src/zkevm_specs/evm/util/call_gadget.py
KimiWu123: > The reason is that the Python code serves as the specification of the circuit. Organizing the constraints in the class and methods makes the reader hard to keep track of variables.

I changed back to class-base implementation to align with the implementation in `zkEVM-Circuit`([#1023](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1023)). Would like to know your opinion. There are some failed cases and  I'll keep working on fixing them if you're fine with this change.

Btw, I rollback @ed255's change in #354, [here ](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/354/files#diff-10487489ead1ac341b7671be200b7c7b4b63c000a33f05ffb0bcfe63e73d6ff0R131-R133). The original way is more consistent with the implementation in `zkEVM-circuit`. Do you have any concern?

src/zkevm_specs/evm/execution/callop.py
KimiWu123: fixed in [ad22495](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348/commits/ad224959a3e90df0f79898b52888fab521c80f8b)

src/zkevm_specs/evm/execution/oog_call.py
```diff
@@ -22,55 +15,19 @@ def oog_call(instruction: Instruction):
...
-
-    has_value = 1 - instruction.is_zero(value)
-    callee_address = instruction.rlc_to_fq(callee_address_rlc, N_BYTES_ACCOUNT_ADDRESS)
+    call = CallGadget(instruction, FQ(0), FQ(1), FQ(0), FQ(0))
 
     # TODO: handle PrecompiledContract oog cases
 
-    # Add callee to access list
-    is_warm_access = instruction.read_account_to_access_list(tx_id, callee_address)
+    # Add callee to access list``
```
ChihChengLiang, line 23:
```suggestion
    # Add callee to access list
```

src/zkevm_specs/evm/util/call_gadget.py
```diff
@@ -0,0 +1,122 @@
...
+            # Verify is_success is a bool
+            self.instruction.constrain_bool(self.is_success)
+            self.gas = instruction.rlc_to_fq(self.gas_rlc, N_BYTES_GAS)
+            self.is_u64_gas = instruction.is_zero(
+                instruction.sum(self.gas_rlc.le_bytes[N_BYTES_GAS:])
+            )
+        else:
+            self.instruction.constrain_zero(self.is_success)
+            self.gas = FQ(0)
+            self.is_u64_gas = FQ(0)
```
ChihChengLiang, line 73: I don't think we need to constrain `gas` and `is_u64_gas` to zero. 

```suggestion
```

ChihChengLiang (commented):
First pass

src/zkevm_specs/evm/util/call_gadget.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __init__(
+        self,
+        instruction: Instruction,
+        is_success_call: FQ,
+        is_call: FQ,
+        is_callcode: FQ,
+        is_delegatecall: FQ,
+    ):
+        self.instruction = instruction
```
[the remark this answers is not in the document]
KimiWu123: fixed in [53df9ee](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348/commits/53df9ee479bd2b2f70207363595dafb265f91a96)

src/zkevm_specs/evm/execution/oog_call.py
KimiWu123: fixed in [53df9ee](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348/commits/53df9ee479bd2b2f70207363595dafb265f91a96)

src/zkevm_specs/evm/util/call_gadget.py
KimiWu123: fixed in [53df9ee](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348/commits/53df9ee479bd2b2f70207363595dafb265f91a96)

ChihChengLiang (approved):
LGTM

src/zkevm_specs/evm/execution/callop.py
```diff
@@ -171,9 +121,9 @@ def callop(instruction: Instruction):
         is_precompile, FQ(instruction.next.execution_state in precompile_execution_states())
     )
 
-    if is_empty_code_hash == FQ(1) and is_precompile == FQ(0):
+    if call.is_empty_code_hash == FQ(1) and is_precompile == FQ(0):
```
ed255, line 124: Recently the case of \*CALL\* without code execution has been updated due to how non-existing accounts are handled.
Non-existing accounts are now identified by having `code_hash = 0`, which is not the empty hash value.  But those cases should still be handled as no code execution.
So I think it would be better to define:
```python
no_callee_code = call.is_empty_code_hash + call.callee_not_exists
```
And then in this if replace the first check by `no_callee_code == FQ(1)`

I notice that `call.callee_not_exists` is missing from the `CallGadget` in this PR.  I introduced that in a very recent PR in the circuits. See https://github.com/privacy-scaling-explorations/zkevm-circuits/commit/aaaf13ee3020661e048484776bd88686d1659042#diff-08d58ebc59ab6c11bf7c049e859441eed2319a03feb3fdffaddaba73f8a8c829

src/zkevm_specs/evm/util/call_gadget.py
```diff
@@ -0,0 +1,118 @@
...
+        # Verify memory expansion
+        (
+            self.next_memory_size,
+            self.memory_expansion_gas_cost,
+        ) = instruction.memory_expansion_dynamic_length(
+            self.cd_offset,
+            self.cd_length,
+            self.rd_offset,
+            self.rd_length,
+        )
```
ed255, line 89: I noticed that the fields `callee_code_hash` and `is_empty_code_hash` are not set in the `__init__`.  They are set in the `gas_cost` method.  Why is that?  Isn't it better to initialize all the Class fields in `__init__`?  The circuit does it like this;
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/b77613e9b42eacb03ccc894f0b353c374da3f283/zkevm-circuits/src/evm_circuit/util/common_gadget.rs#L523-L525

ed255 (commented):
I've added a few small notes related to the latest circuit implementation, but otherwise looks good!  Please take a look at the two comments :)

src/zkevm_specs/evm/execution/callop.py
KimiWu123: fixed in [a7d329a](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348/commits/a7d329abe49ca42ed8b9882234e04d853206f6a0)

src/zkevm_specs/evm/util/call_gadget.py
KimiWu123: yes, I totally agree that. However, I don't know how to hadle it properly because I need to get `callee_code_hash` first as following,
```
        self.callee_code_hash = instruction.account_read(
            self.callee_address, AccountFieldTag.CodeHash
        )
```

If I moved above code to `__init__`, it breaks the order of rw table. (reading the value of `AccountFieldTag.CodeHash` must be after reading the value of `AccountFieldTag.Balance`, see [here](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/tests/evm/test_callop.py#L345-L355)). Either I change the order in the test or assign an arbitrary value in `__init__`(It will fail if anyone uses these vars before calling `gas_cost`, but I don't think it's a good idea). Do you have any better ideas for this?

ed255: Ah I see!
After this change was applied https://github.com/privacy-scaling-explorations/zkevm-specs/pull/354 it will become impossible to read a `AccountFieldTag.Balance` for non-existing accounts.  This means that the order of lookups for possibly non-existing accounts should always be
1. Read CodeHash
2. If CodeHash != 0, then read other fields as necessary

By impossible what I mean is that it won't be possible to generate a list of MPT updates + proofs that can verify it.  Since we don't have the MPT circuit connected yet, tests will not fail (although we could add some checks to make sure we never read non-codeHash account fields of non-existing accounts).

So I think you should move the callee CodeHash lookup to happen first, and if account exists, then do the callee Balance lookup.  I think this affects the circuits as well.  I see in `error_oog_call.rs` first the callee Balance is read, and then ` call_gadget.gas_cost_expr` is called, which causes the callee CodeHash read.  Actually, I'm not sure why in `error_oog_call` the callee balance is read, there doesn't seem to be any constraint using that value: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/ea6d8e38bcaedbf2d69d0661ad14485c6b02b9cb/zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs#L66-L71

So I think you should change the order of reads in the test for the callee: first read the CodeHash, then if its != 0, read the Balance.

After a second look at the test code you mentioned https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/tests/evm/test_callop.py#L345-L355 I see that the Balance read in there is for the caller which is independent from the callee CodeHash read; which means that the order doesn't matter, but it's still OK to read the callee CodeHash before the caller Balance.

So in summary: for possibly non-existing accounts, first read codeHash, and only if its !=0 read other fields.  For accounts known to be existing in the context (like the caller in the call context), the order of reads is not important.

KimiWu123: the order of CodeHash read was fixed in [7e0fc6e](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348/commits/7e0fc6e7ae1880835beefa735e8fe290338eac15)
.

> Actually, I'm not sure why in error_oog_call the callee balance is read, there doesn't seem to be any constraint using that value:

yep, balance is not necessary in `error_oog_call`, I have removed it.
.

> So I think you should change the order of reads in the test for the callee: first read the CodeHash, then if its != 0, read the Balance.

For `then if its != 0, read the Balance`, I'm not sure I understand it corretly. Do you mean add a condiftion `if callee_bytecode_hash != 0` at [test_callop.py#L349](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/tests/evm/test_callop.py#L349) and test_callop.py#L353? From my understanding, in evm, user's balance is still able to be read even it's a non-existing account (not initialized in MPT). If we add `if callee_bytecode_hash != 0`, I guess it'll panic(or just fail) if we read the balance of a non-existing account. Not sure my understanding is correct and not sure I understand your meaning correctly.

ed255: > For `then if its != 0, read the Balance`, I'm not sure I understand it corretly. Do you mean add a condiftion `if callee_bytecode_hash != 0` at [test_callop.py#L349](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/tests/evm/test_callop.py#L349) and test_callop.py#L353?

I was mistaken on my interpretation of the callop logic, so the `then if its != 0, read the Balance` doesn't apply here, because the Balance is read on the caller, which at this point is guaranteed to exist, so there's no need to make it conditional.  I think there's an edge case on the transfer thought, but more on it later.

> From my understanding, in evm, user's balance is still able to be read even it's a non-existing account (not initialized in MPT).

From the EVM spec, whenever an opcode that reads account fields encounters a non-existing account particular conditions are applied.  Most of the time the behavior is the same as considering the non-existing account as a default account (with nonce=0, balance=0, code_hash=EMPTY_HASH), but that's not always the case.  For example, EXTCODEHASH returns 0 for non-existing accounts (instead of EMPTY_HASH).  This also affects writes, as opcodes that write to accounts may have different gas cost when they create the account (because it was non-existing before).

In our circuits, account read/writes are triggered by the EVM circuit.  Those read/writes are then verified for consistency in the State Circuit, and finally they are collected and matched against MPT proofs to match the state trie transitions.  Our current design of the MPT Circuit supports the proofs listed in this table: https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/tables.md#mpt_table

In particular, the `NonceMod`, `BalanceMod` and `CodeHashMod` don't support a read (valuePrev == value) of non-existing accounts.  They only support reads to existing accounts, and writes to existing accounts or to non-existing accounts (which triggers account creation).

So a Nonce, Balance, CodeHash read proof for non-existing accounts is not supported by the MPT circuit.

The MPT circuit supports `NonExistingAccountProof` which acts like a read and proves that an account doesn't exist.

And in the StateCircuit we map `CodeHash` read of 0 to `NonExistingAccountProof` in the MPT.  We can do this because code_hash will never be 0 in an existing account, so wenever in the EVM/State circuit a CodeHash is read as 0, it means that the account doesn't exist.  And if the account doesn't exist, a following Nonce or Balance read (triggered by the EVM) can't be proved by the MPT Circuit.

This is why every time a Nonce/Balance is read from the EVM circuit, we must make sure that the account exists, otherwise it may be translated to a  `NonceMod` or `BalanceMod` proof in the MPT that is impossible to succeed.  So the idea is, in those cases, have a boolean expression that encodes the account existence, and use it to enable/disable the field lookups.

The way we disable a lookup is by:
1- In all tables always have an all 0 row
2- Multiply all expressions of the lookup by the "selector" expression, so that when it's zero, the lookup is a list of 0s and always suceceds, bypassing the other possibly non-zero values in the original lookup expression.

> If we add `if callee_bytecode_hash != 0`, I guess it'll panic(or just fail) if we read the balance of a non-existing account. Not sure my understanding is correct and not sure I understand your meaning correctly.

I'm not sure I understand this part

ed255: BTW I'm working on a simple check on zkevm-circuits that will guarantee that we don't have invalid account field read/writes: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1084

ed255 (approved):
LGTM! Nice work!

As I mentioned in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/348#discussion_r1073778886 the logic of `if code_hash == 0, skip balance read lookup` doesn't apply in this opcode because the possibley non-existing account is the callee, whereas the balance read lookup is done on the caller.  So the PR looks good to me in the current form :)

src/zkevm_specs/evm/util/call_gadget.py
ed255: I consider this conversation resolved!  But we can continue a discussion to resolve any doubts, etc :)

Issue #364, opened 2023-01-27
Synchronize State Circuit


ashWhiteHat:
# Todo
- [x] out dated api modification https://github.com/privacy-scaling-explorations/zkevm-specs/issues/230
- [x] copy table sync https://github.com/privacy-scaling-explorations/zkevm-specs/issues/377
- [x] rw table sync https://github.com/privacy-scaling-explorations/zkevm-specs/issues/376
- [ ] synchronize prev column https://github.com/privacy-scaling-explorations/zkevm-specs/pull/415
- [ ] create error execution state synchronizing status
- [x] cover test https://github.com/privacy-scaling-explorations/zkevm-specs/pull/389#pullrequestreview-1387569997

## Contraints

|Resource|Constraints|Circuit|Spec|Synchronized|
|---|---|---|---|---|
|Start|1.0|🍗|🍗|✅|
||1.1|🍗|🍗|✅|
||1.2|🍗|🍗|✅|
||1.3|🍗|🍗|✅|
||1.4|🍗|🍗|✅|
||1.5|🍗|🐥|🚧|
|Memory|2.0|🍗|🍗|✅|
||2.1|🐔|🍗|🚧|
||2.2|🍗|🍗|✅|
||2.3|🍗|🍗|✅|
||2.4|🍗|🍗|✅|
||2.5|🍗|🍗|✅|
||2.6|🍗|🐥|🚧|
|Stack|3.0|🍗|🍗|✅|
||3.1|🍗|🍗|✅|
||3.2|🍗|🍗|✅|
||3.3|🍗|🍗|✅|
||3.4|🍗|🍗|✅|
||3.5|🍗|🍗|✅|
||3.6|🍗|🐥|🚧|
|Storage|4.0|🍗|🍗|✅|
||4.1|🍗|🍗|✅|
||4.2|🍗|🐥|✅|
|CallContext|5.0|🍗|🍗|✅|
||5.1|🍗|🍗|✅|
||5.2|🍗|🍗|✅||
||5.3|🍗|🍗|✅|
||5.4|🍗|🍗|✅|
||5.5|🍗|🐥|✅|
|Account Storage|6.0|🍗||✅|
||6.1|🍗|🐥|🚧|
||6.2|🍗|🐥|🚧|
|TxRefund|7.0|🍗|🍗|✅|
||7.1|🍗|🍗|✅|
||7.2|🍗|🍗|✅|
||7.3|🐥|🍗|✅|
[...]

PR #367, opened 2023-02-02
tx refund synchronization

ashWhiteHat:
Part of https://github.com/privacy-scaling-explorations/zkevm-specs/issues/364

I synchronized tx refund constraints with `circuits`.
I would like to check constraints synchronization progress with serial numbers so I defined the tx refund spec and updated comment in state circuit.

What I did are following.
- add initial value to row
- add initial value constraint to state circuit
- give constraints to numbers
- describe tx refund spec and constraints number
- remove account destructed according to https://github.com/privacy-scaling-explorations/zkevm-specs/issues/259#issuecomment-1344003567

The constraint is aligned with https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/state_circuit/constraint_builder.rs#L322

Thank you.


specs/state-proof.md
```diff
@@ -8,15 +8,15 @@ State proof maintains the read-write part of [random accessible data](./evm-proo
...
-- `AccountDestructed`: State of destruction of an account
+1. `Start`: Start of transaction and padding row
+2. `Memory`: Call's memory as a byte array
+3. `Stack`: Call's stack as RLC-encoded word array
+4. `Storage`: Account's storage as key-value mapping
+5. `CallContext`: Context of a Call
+6. `Account`: Account's state (nonce, balance, code hash)
+7. `TxRefund`: Value to refund to the tx sender
+8. `TxAccessListAccount`: State of the account access list
+9. `TxAccessListAccountStorage`: State of the account storage access list
```
ed255, line 19: I found these two missing: `TxLog` and `TxReceipt`

src/zkevm_specs/state_circuit.py
```diff
@@ -80,6 +79,7 @@ class Row(NamedTuple):
                       FQ,FQ,FQ,FQ,FQ,FQ,FQ,FQ]
     value: FQ
     committed_value: FQ
+    initial_value: FQ
```
ed255, line 82: I was under the impression that `committed_value` and `initial_value` are the same concept.

There's an inconsistency in our code/specs.  I think CommittedValue and InitValue are the same.  In the circuit it's "initial_value", in the spec tables it's "InitialValue" and in the python specs it's "committed_value".

I think it makes sense to call the column "initial_value", and the meaning of initial_value for state_trie values is committedValue (because it was the value committed in the previous state trie).

So I think instead of adding a column named `initial_value` we should rename `committed_value` to `initial_value`.

---

The column that may be missing that is discussed in https://github.com/privacy-scaling-explorations/zkevm-specs/issues/202 would be `previous_vale`, but we can figure that out in another PR.

ed255 (commented):
Thanks for the spec synchronization!  I've added two comments, please take a look :)

ashWhiteHat:
Hi @ed255 
Thank you for the review.
I modified changes according to your review.

Thank you!

ed255 (approved):
LGTM! Thanks for taking care of this synchronization between circuits and specs :smiley:

han0110 (approved):
LGTM!

Issue #323, opened 2022-11-23
Precompile: 0x06 ecAdd


ChihChengLiang:
Probably some of the wrong curve circuit logic would be useful here. https://github.com/privacy-scaling-explorations/halo2wrong

KimiWu123:
Scroll team' implementation: https://github.com/scroll-tech/zkevm-circuits/pull/598 and a path, https://github.com/scroll-tech/zkevm-circuits/pull/811

roynalnaruto:
@KimiWu123 @ChihChengLiang  to summarise our implementation for `ecAdd`, `ecMul` and `ecPairing`:
- Inputs to the above precompiles are fed as witness to the [`EccCircuit`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/ecc_circuit.rs)
- We then copy those advices to the [`EccTable`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/table.rs#L2334-L2373)
- We validate calls to those precompiles through:
    - EVM Circuit [`CallOp`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/evm_circuit/execution/callop.rs#L309)
    - [`PrecompileGadget`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/evm_circuit/util/precompile_gadget.rs)
    - And individual verification gadgets, in the above cases:
        - [`EcAddGadget`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/evm_circuit/execution/precompiles/ec_add.rs)
        - [`EcMulGadget`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/evm_circuit/execution/precompiles/ec_mul.rs)
        - [`EcPairingGadget`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/evm_circuit/execution/precompiles/ec_pairing.rs)

KimiWu123:
> @KimiWu123 @ChihChengLiang to summarise our implementation for `ecAdd`, `ecMul` and `ecPairing`:
> 
> * Inputs to the above precompiles are fed as witness to the [`EccCircuit`](https://github.com/scroll-tech/zkevm-circuits/blob/develop/zkevm-circuits/src/ecc_circuit.rs)
[...]

PR #375, opened 2023-02-04
[WIP] bn254Add precompile

ashWhiteHat:
Close https://github.com/privacy-scaling-explorations/zkevm-specs/issues/323

[bn254Add spec](https://www.evm.codes/precompiled#0x06?fork=merge)

I implemented convertor between `py_ecc` and `cloudflare bn254` point.
In precompile bn254Add, it uses `cloudflare bn254` and its coordinate is expressed as mont limbs but we use `py_ecc FQ` on circuit.
It doesn't have mont form so I implemented convertor.

Thank you!


han0110:
Hey Ash, I think having the point addition inline in EVM circuit would be a little bit costly (the height of the step might be much more than current ones), and I have been thinking to introduce another sub-circuit to handle all BN254 operations then EVM circuit could use it directly via lookup, so I opened an issue here https://github.com/privacy-scaling-explorations/zkevm-specs/issues/386, would you like to check and perhaps leave some feedback? If you think the proposal is okay, then we could start to draft the `bn254_table` and implement the precompile by doing lookup to it.

ChihChengLiang:
Completed in #500

PR #381, opened 2023-02-09
Markdown doc for `ErrorOutOfGasSloadSstore` state

silathdiir:
Close https://github.com/privacy-scaling-explorations/zkevm-specs/issues/382

Circuit impl PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1147

For `SLOAD` gas calculation with EIP-2929 could reference [this go-ethereum function](https://github.com/ethereum/go-ethereum/blob/fd4230f695247102c249d5ece2c7f6760350ae6f/core/vm/operations_acl.go#L103
).
And for `SSTORE` gas calculation with EIP-3529 could reference [this go-ethereum function](https://github.com/ethereum/go-ethereum/blob/fd4230f695247102c249d5ece2c7f6760350ae6f/core/vm/operations_acl.go#L27).

### DIFF
--- a/specs/error_state/ErrorOutOfGasSloadSstore.md
+++ b/specs/error_state/ErrorOutOfGasSloadSstore.md
@@ -25,7 +25,7 @@ For this gadget, SSTORE gas cost is calculated with EIP-3529 as:
 if value == value_prev:
   gas_cost = SLOAD_GAS # 100
 else:
-  if value == original_value:
+  if value_prev == original_value:
     if original_value == 0:
       gas_cost = SSTORE_SET_GAS # 20000
     else:
--- a/specs/error_state/ErrorOutOfGasSloadSstore.md
+++ b/specs/error_state/ErrorOutOfGasSloadSstore.md
@@ -1,0 +1,62 @@
+# ErrorOutOfGasSloadSstore state for both SLOAD and SSTORE OOG errors
+
+## Procedure
+
+Handle the corresponding out of gas errors for both `SLOAD` and `SSTORE` opcodes.
+
+### EVM behavior
+
+For the current `go-ethereum` code, the out of gas error may occur for `constant gas` or `dynamic gas`.
+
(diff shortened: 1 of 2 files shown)


han0110 (approved):
LGTM!

miha-stopar (approved):
LGTM!

specs/error_state/ErrorOutOfGasSloadSstore.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+#### SSTORE gas cost
+
+For this gadget, SSTORE gas cost is calculated with EIP-3529 as:
+```
+if value == value_prev:
+  gas_cost = SLOAD_GAS # 100
+else:
+  if value == original_value:
```
KimiWu123, line 28: Here should be `!=`. If it's a `==` which means `original_value = value == 0 & value != value_prev` ==> e.g. 0 --> non-zero --> 0. In this case, gas_cost should be `SLOAD_GAS` only.

KimiWu123 (changes requested):

specs/error_state/ErrorOutOfGasSloadSstore.md
KimiWu123: I think to use `original_eq_prev` [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1147/files#diff-08d58ebc59ab6c11bf7c049e859441eed2319a03feb3fdffaddaba73f8a8c829R719-R724) is incorrect. 

Let's see the case, `value_eq_prev == false && original_eq_prev == false`, the `gas_cost` would be `WARM_ACCESS`. The case means `original_value != value_pre != value` but it doesn't imply `original_value == value` and I can't find any contraints for this equivalence (correct me if I missed it).

So, I think to follow the spec here (using `value != original_value`) is correct.

KimiWu123 (approved):

PR #392, opened 2023-03-03
amend RETURN_FDREVERT specs to add EIP-211 returndatasize constraint

hero78119:
to close issue https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1250


src/zkevm_specs/evm/execution/return_revert.py
```diff
@@ -109,7 +121,8 @@ def return_revert(instruction: Instruction):
             rw_counter=Transition.delta(rwc_delta + 1),
             call_id=Transition.same(),
         )
-    else:
+
+    if not instruction.curr.is_root:
```
hero78119, line 125: this is just a style alignment with specs/circuit to use explicitly negate condition in previous `if` to replace `else` condition

PR #404, opened 2023-03-22
markdown spec for contract address collision error

### LINKED ISSUE #405: handle ErrorContractAddressCollision error state
when create in evm outputs the same existing address, it meets address collision error.

DreamWuGit:
target to resolve #405


KimiWu123:
has been addressed by #466

PR #407, opened 2023-03-31
pse-word-addr-mem: Word-Addressable Memory

naure:
Currently, Execution gadgets MLOAD/MSTORE as well as the Copy Circuit access memory byte-by-byte through the State Circuit. This can be replaced by 32B-word operations to reduce load on the State Circuit.

This is the python implementation of the [spec](https://www.notion.so/scrollzkp/Word-Addressable-Memory-Optimization-7ab2c03cb71b4fa3b8710756ce9c5fcb).

### DIFF
--- a/src/zkevm_specs/evm_circuit/execution/memory.py
+++ b/src/zkevm_specs/evm_circuit/execution/memory.py
@@ -1,3 +1,4 @@
+from zkevm_specs.util.arithmetic import RLC
 from ..instruction import Instruction, Transition
 from ..opcode import Opcode
 from ..table import RW
@@ -5,9 +6,16 @@ from ...util import FQ
 
 
 def memory(instruction: Instruction):
+    X32 = make_X32(instruction.randomness)
+
     opcode = instruction.opcode_lookup(True)
 
     address = instruction.stack_pop()
+    shift = address.int_value % 32
+    slot = address.int_value - shift
+
+    addr_left = FQ(slot)
+    addr_right = FQ(slot + 32)
 
     is_mload = instruction.is_equal(opcode, Opcode.MLOAD)
     is_mstore8 = instruction.is_equal(opcode, Opcode.MSTORE8)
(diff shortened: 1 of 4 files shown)


src/zkevm_specs/evm_circuit/execution/memory.py
```diff
@@ -22,18 +30,32 @@ def memory(instruction: Instruction):
...
+        value_left = instruction.memory_lookup(
+            RW.Write if is_store == FQ(1) else RW.Read, addr_left
+        )
+        value_right = instruction.memory_lookup(
+            RW.Write if is_store == FQ(1) else RW.Read, addr_right
+        )
+
+        # Check consistency of value, value_left, and value_right.
+        mask = make_mask(shift)
+        X = instruction.randomness
```
DreamWuGit, line 46: memory opcodes(mload/mstore/mstore8) don't use copy circuit. I conclude if only refactor these memory opcodes, just evm execution circuit and state circuit require refactor accordingly. copy circuit doesn't need to change.  only when update execution codes(like calldatacopy, log* etc.) requires copy circuit udpates.

src/zkevm_specs/evm_circuit/execution/memory.py
```diff
@@ -43,3 +65,58 @@ def memory(instruction: Instruction):
...
+def check_mask(offset, M):
+    # Compute 2**offset by squaring-and-multiplying.
+    offset_value = 1
+    for i in reversed(range(5)):
+        bit = (offset >> i) & 1
+        offset_value = offset_value * offset_value * (1 + bit)
+
+    # Interpret the mask as a binary number.
+    mask_value = 0
+    for (i, m) in enumerate(M):
```
DreamWuGit, line 79: need to check `M` begin with consecutive 1 explicitly ?

src/zkevm_specs/evm_circuit/execution/memory.py
naure: The copy circuit needs to change because the memory format changed. It must read and write the same way that the memory instructions do, so by word instead of by byte.

This is not included in this PR yet.

src/zkevm_specs/evm_circuit/execution/memory.py
naure: The assert with offset_value below is supposed do that.

CPerezz (commented):
Hey! I also thought about this some time ago. And did again when thinking about ZKWasm.

I like the idea of using masks and therefore be able to read on the same way memory always.
In any case, there are some questions I have:
- How much do we offload the State Circuit vs. we load with constraints the copy?
- I presume in the circuit we also need to prove correct mask gen/application right? That feels like going too much to the binary world which isn't usually super performant. Do you have any intuitions on the costs etc..?


I think if we can estimate a bit the cost differences, will be much easier to decide going for this solution or not. 
In any case, I love the proposal, just need to see some evidence to go for it!

naure:
Thanks @CPerezz ! Did you also review the spec? ([Link](https://scrollzkp.notion.site/Word-Addressable-Memory-Optimization-7ab2c03cb71b4fa3b8710756ce9c5fcb)). I updated it today.

* The load on the State Circuit should be much reduced: `32X for \*COPY\* and MSTORE8`; or `16X for MLOAD and MSTORE`.
* The Copy Circuit should not change much in size. Recall that copies process 1 byte per row, with source and destination in parallel. Since one side is still byte-by-byte (e.g., bytecode), there is space on the memory side to recompose the memory word into bytes. It might still require a little more witness.
* The Execution Circuit looks reasonably efficient. There are no hard binary operations, only some `byte AND boolean`, which is are simple field multiplications. It uses polynomial identities to check all the bytes in a few constraints. A difficulty is to handle the variable `offset`, but that could be solved efficiently with pure algebra. I also made sure to factorize together the common computations of the various operations.

ed255:
Closing this PR because we decided to archive the python implementation and just maintain the markdown specs.
Just having a markdown spec for the Word-Addressable Memory feature will be enough.

PR #410, opened 2023-04-06
markdown spec for oog account access

DreamWuGit:


ed255:
Closing because we already have this specified in https://github.com/privacy-scaling-explorations/zkevm-specs/blob/cb85f8be5337641856ce48dd870ff6167fbca492/specs/error_state/ErrorOutOfGasAccountAccess.md

PR #416, opened 2023-04-27
feat/#412 CREATE/CREATE2 Synchronization

### LINKED ISSUE #412: CREATE/CREATE2 Synchronization
Here is the comment from Han
https://github.com/privacy-scaling-explorations/zkevm-specs/pull/355#pullrequestreview-1392383794

KimiWu123:
closed #412

### DIFF
--- a/src/zkevm_specs/evm_circuit/execution/create.py
+++ b/src/zkevm_specs/evm_circuit/execution/create.py
@@ -44,18 +44,20 @@ def create(instruction: Instruction):
     is_static = instruction.call_context_lookup(CallContextFieldTag.IsStatic)
     reversion_info = instruction.reversion_info()
 
+    has_init_code = size != FQ(0)
+
     # calculate contract address
+    code_hash = instruction.next.code_hash if has_init_code else Word(EMPTY_CODE_HASH)
     contract_address = (
         instruction.generate_contract_address(caller_address, nonce)
         if is_create == 1
-        else instruction.generate_CREAET2_contract_address(
-            caller_address, salt_word, instruction.next.code_hash
-        )
+        else instruction.generate_CREAET2_contract_address(caller_address, salt_word, code_hash)
     )
 
-    # verify the equality of input `size` and length of calldata
-    code_size = instruction.bytecode_length(instruction.next.code_hash)
-    instruction.constrain_equal(code_size, size)
+    if has_init_code:
+        # verify the equality of input `size` and length of calldata
+        code_size = instruction.bytecode_length(instruction.next.code_hash)
+        instruction.constrain_equal(code_size, size)
 
     # verify return contract address
     instruction.constrain_equal(
@@ -99,10 +101,7 @@ def create(instruction: Instruction):
     instruction.transfer(caller_address, contract_address, value_word, callee_reversion_info)
 
     # gas cost of memory expansion
-    (
-        next_memory_size,
-        memory_expansion_gas_cost,
-    ) = instruction.memory_expansion(
+    (next_memory_size, memory_expansion_gas_cost,) = instruction.memory_expansion(
         offset,
         size,
     )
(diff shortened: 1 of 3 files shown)


KimiWu123:
> Handles insufficient balance, nonce overflow, depth overflow in the same gadget

nonce overflow check: https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/src/zkevm_specs/evm_circuit/execution/create.py#L70
Depth overflow check: https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/src/zkevm_specs/evm_circuit/execution/create.py#L67
insufficent balance check (inside transfer gadget): https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/src/zkevm_specs/evm_circuit/execution/create.py#L99


> Skips new context if init code is empty

fixed in [308ef82](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/308ef82d1182449214377fce9ad89fe58ee20590)


> Does keccak table lookup itself without checking bytecod length (which I think we need)

After discussed with Han, we don't need to do this because we had 
1. copy table lookup and 
2. bytecode length lookup.

Those are enough to make sure bytecode hash is consistent.

KimiWu123:
> Currently in spec we just forbid these situation to happen, but perhaps we could just handle these non-reverting cases together to avoid redudant constraints in different execution state.

Handled the non-reverting cases as you mentioned in [17fc20a](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/17fc20a257ebe1f1bb1c3322175a3dae00daaa43). Please help to review the way I implemented makes sense to you or not. As you mentioned `handle these non-reverting cases together`, prefer to have another ticket to review all the op codes and resolve it.

src/zkevm_specs/evm_circuit/execution/create.py
```diff
@@ -130,71 +81,181 @@ def create(instruction: Instruction):
...
-        (CallContextFieldTag.MemorySize, next_memory_size),
-        (
-            CallContextFieldTag.ReversibleWriteCounter,
-            instruction.curr.reversible_write_counter + 1,
-        ),
-    ]:
+
+    # error cases, should end this call and return
+    # is_static should be false
+    if not is_precheck_ok or is_static.expr() == FQ(1):
```
han0110, line 99: I think we should just constrain that `is_static == 0`.

src/zkevm_specs/evm_circuit/execution/create.py
```diff
@@ -130,71 +81,181 @@ def create(instruction: Instruction):
...
+            )
+
+        instruction.constrain_step_state_transition(
+            rw_counter=Transition.delta(instruction.rw_counter_offset),
+            program_counter=Transition.delta(1),
+            stack_pointer=Transition.delta(2 + is_create2),
+            reversible_write_counter=Transition.delta(1),
+            gas_left=Transition.delta(-gas_cost),
+            # Always stay same
+            memory_word_size=Transition.same(),
```
han0110, line 117:
```suggestion
            memory_word_size=Transition.to(next_memory_size),
```

src/zkevm_specs/evm_circuit/execution/create.py
```diff
@@ -130,71 +81,181 @@ def create(instruction: Instruction):
...
-        offset,  # src_addr
-        offset + size,  # src_addr_boundary
-        FQ(0),  # dst_addr
-        size,  # length
-        instruction.curr.rw_counter + instruction.rw_counter_offset,
+    ### Do stack depth, nonce and balance pre-check
+    # ErrDepth constraint
+    is_error_depth, _ = instruction.compare(FQ(CALL_CREATE_DEPTH), depth, N_BYTES_STACK)
+    # ErrNonceUintOverflow constraint
+    is_nonce_overflow, _ = instruction.compare(nonce, nonce_prev, N_BYTES_U64)
```
han0110, line 88: We need to compare `nonce_prev` with <code>2<sup>64</sup>-1</code> here (see https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2d2bfc6ccf179ade1a8d063f9586b93e5283a557/zkevm-circuits/src/evm_circuit/execution/create.rs#L180).

src/zkevm_specs/evm_circuit/execution/create.py
```diff
@@ -130,71 +81,181 @@ def create(instruction: Instruction):
...
+    # ErrInsufficientBalance constraint
+    is_insufficient_balance, _ = instruction.compare_word(Word(balance_prev.expr().n), value_word)
+
+    # pass the pre-check if none of above errors happen
+    is_precheck_ok = (
+        is_error_depth == FQ(0) and is_nonce_overflow == FQ(0) and is_insufficient_balance == FQ(0)
```
han0110, line 94: To have deterministic circuit trace, we also need to constrain the order of these cases like https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2d2bfc6ccf179ade1a8d063f9586b93e5283a557/zkevm-circuits/src/evm_circuit/execution/create.rs#L182-L199

han0110: Just discussed offline and we find the cirucit should be updated to allow these exception to happen at the same time, so no need this check.

KimiWu123: After discussed with Han, it's possible to have nonce overflow, insufficient balance and depth too deep in the same time. So, we don't need to constraint the relationship among these three errors.

src/zkevm_specs/evm_circuit/execution/create.py
KimiWu123: fixed in [22b6eff](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/22b6effbb5dc11a274a9382e089bdc7d439bcb9e)

src/zkevm_specs/evm_circuit/execution/create.py
KimiWu123: fixed in [22b6eff](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/22b6effbb5dc11a274a9382e089bdc7d439bcb9e)

src/zkevm_specs/evm_circuit/execution/create.py
KimiWu123: fixed in [22b6eff](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/22b6effbb5dc11a274a9382e089bdc7d439bcb9e)

src/zkevm_specs/evm_circuit/execution/create.py
```diff
@@ -130,71 +85,181 @@ def create(instruction: Instruction):
...
-        offset,  # src_addr
-        offset + size,  # src_addr_boundary
-        FQ(0),  # dst_addr
-        size,  # length
-        instruction.curr.rw_counter + instruction.rw_counter_offset,
+    ### Do stack depth, nonce and balance pre-check
+    # ErrDepth constraint
+    is_error_depth, _ = instruction.compare(FQ(CALL_CREATE_DEPTH), depth, N_BYTES_STACK)
+    # ErrNonceUintOverflow constraint
+    is_nonce_overflow, _ = instruction.compare(FQ(MAX_U64), nonce, N_BYTES_U64)
```
han0110, line 92: When nonce is added to be `MAX_U64+1`, the `compare` will fail because it's no longer in 8-byte, so we should check `nonce_prev` is less than `MAX_U64` here.

```suggestion
    # ErrNonceUintOverflow constraint
    is_nonce_in_range, _ = instruction.compare(nonce_prev, FQ(MAX_U64), N_BYTES_U64)
```

src/zkevm_specs/evm_circuit/execution/create.py
han0110: I think we can remove `or is_static.expr() == FQ(1)` here

han0110 (commented):
Final 2 nits, other LGTM!

src/zkevm_specs/evm_circuit/execution/create.py
KimiWu123: fixed in [048ab74](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/048ab74e8c19e9d3d997c416f42221c68afe31a1)

src/zkevm_specs/evm_circuit/execution/create.py
KimiWu123: fixed in [048ab74](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/416/commits/048ab74e8c19e9d3d997c416f42221c68afe31a1)

han0110 (approved):

tests/evm/test_create.py
```diff
@@ -209,20 +226,33 @@ def test_create_create2(
...
+
     # CREATE: 33 * 3(push) + 1(CREATE) + 1(mstore) + 33(PUSH32) + 2(PUSH) + 1(RETURN)
     # CREATE2: 33 * 4(push) + 1(CREATE2) + 1(mstore) + 33(PUSH32) + 2(PUSH) + 1(RETURN)
-    next_program_counter = 33 * 4 + 1 + 35 + 1 if is_create2 else 33 * 3 + 1 + 35 + 1
-    assert caller_bytecode.code[next_program_counter] == opcode
+    pc = 33 * 4 + 1 if is_create2 else 33 * 3 + 1
+    if has_init_code:
+        next_program_counter = pc + 1 + 35 + 1 if is_create2 else pc + 1 + 35 + 1
+    else:
+        next_program_counter = pc
```
ChihChengLiang, line 238: I think we can simplify this
```suggestion
    next_program_counter = 33 * 4 + 1 if is_create2 else 33 * 3 + 1
    if has_init_code:
        next_program_counter += 1 + 35 + 1
```

ChihChengLiang (approved):
LGTM. I added a nitpick

Issue #425, opened 2023-05-22
Document how we handle non-existing accounts and account creation (with CodeHash = 0, etc)


ed255:
Related issue https://github.com/privacy-scaling-explorations/zkevm-specs/issues/395

Currently there's no central source of documentation of this, I think it's scattered in comments and issues.

PR #426, opened 2023-05-23
Document Account and Storage rws in RwTable

ed255:
Resolve https://github.com/privacy-scaling-explorations/zkevm-specs/issues/425

I noticed that we already had some documentation in https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/evm-proof.md#account-non-existence so I've added a link to that section in the new paragraph.


hero78119:
Very clear documentation!

Issue #429, opened 2023-05-29
Investigate why mypy type check not work properly.


hero78119:
`mypy .` capture less source files and skip others sources without reasons and error message.

One possible root cause is it failed silently, mostly because some invalid source file/naming exist which make it stop silently

Try upgrade mypy to 1.3.0 `https://pypi.org/project/mypy/1.3.0/`, still not works.

PR #430, opened 2023-05-29
Make sure mypy checks all files

ed255:
I've searched for other projects using mypy encountering similar issues (not all files are checked) and I found that setting the `MYPATH` env var helps.  With this new type errors popped up, which I fixed in this PR.

Resolve https://github.com/privacy-scaling-explorations/zkevm-specs/issues/429

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ lint: ## Check whether the code is formatted correctly
 	flake8 .
 
 type: ## Check the typing of the Python code
-	mypy .
+	MYPATH=src mypy .
 
 test: ## Run tests
 	pytest --doctest-modules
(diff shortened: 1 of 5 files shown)


hero78119:
By my local test, it's still only check 15 source files, and not checking file under `zkevm_specs/evm_circuit/execution/*`

To reproduce,  add new unused import e.g. in `zkevm_specs/evm_circuit/execution/gas.py` `from ...util import Expression`
then `make type` still pass check

hero78119:
> By my local test, it's still only check 15 source files, and not checking file under `zkevm_specs/evm_circuit/execution/*`
> 
> To reproduce, add new unused import e.g. in `zkevm_specs/evm_circuit/execution/gas.py` `from ...util import Expression` then `make type` still pass check

Just verified it works! unused import should be handle by lint haha

hero78119 (approved):
LGTM! thanks for the quick fix!

ChihChengLiang (approved):
LGTM. Thanks for the quick fix

PR #433, opened 2023-06-14
Contract creation via Tx

adria0:
This PR enables the contract creation via transaction.

### DIFF
--- a/src/zkevm_specs/evm_circuit/execution/begin_tx.py
+++ b/src/zkevm_specs/evm_circuit/execution/begin_tx.py
@@ -125,7 +125,7 @@ def begin_tx(instruction: Instruction):
 
             # Get code hash of tx calldata
 
-            copy_rwc_inc, rlc_acc = instruction.copy_lookup(
+            copy_rwc_inc, tx_calldata_rlc = instruction.copy_lookup(
                 tx_id,  # src_id
                 CopyDataTypeTag.TxCalldata,  # src_type
                 call_id,  # dst_id
@@ -140,8 +140,7 @@ def begin_tx(instruction: Instruction):
             assert copy_rwc_inc == FQ.zero()
             # no memory involved, no rw counter incremented
 
-            code_hash = instruction.keccak_lookup(tx_call_data_length, rlc_acc)
-            is_empty_code_hash = instruction.is_equal_word(code_hash, Word(EMPTY_CODE_HASH))
+            code_hash = instruction.keccak_lookup(tx_call_data_length, tx_calldata_rlc)
 
             # Copy tx calldata to bytecode table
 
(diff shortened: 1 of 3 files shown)


src/zkevm_specs/evm_circuit/execution/begin_tx.py
```diff
@@ -91,16 +104,94 @@ def begin_tx(instruction: Instruction):
...
+                rw_counter=Transition.delta(9), call_id=Transition.to(call_id)
+            )
+        else:
+            # Expected behabeur
+            # - If initcode does not RETRUN, contract is created empty and value transferred
+            # - If initcode is invalid bytecode or reverts, contract is not created and value not transferred
+
+            # Get code hash of tx calldata
+
+            copy_rwc_inc, rlc_acc = instruction.copy_lookup(
```
ed255, line 128: I would suggest renaming `rlc_acc` to `tx_calldata_rlc` to make it more clear.

src/zkevm_specs/evm_circuit/execution/begin_tx.py
```diff
@@ -91,16 +104,94 @@ def begin_tx(instruction: Instruction):
...
+                FQ.zero(),  # dst_addr
+                tx_call_data_length,  # length
+                instruction.curr.rw_counter + instruction.rw_counter_offset,
+            )
+
+            assert copy_rwc_inc == FQ.zero()
+            # no memory involved, no rw counter incremented
+
+            code_hash = instruction.keccak_lookup(tx_call_data_length, rlc_acc)
+            is_empty_code_hash = instruction.is_equal_word(code_hash, Word(EMPTY_CODE_HASH))
```
ed255, line 144: You could also reuse the `tx_call_data_length == 0` check for this, to save some constraints and cells.

src/zkevm_specs/evm_circuit/table.py
```diff
@@ -764,6 +764,12 @@ def lookup(
     ]
 
     if len(matched_rows) == 0:
+        print("LOOKUP FAILURE")
+        print("TABLE: ", table_name)
+        print("QUERY: ", query)
+        print("CURRENT VALUES")
+        for row in table:
+            print("-->", row)
```
ed255, line 772: leftover?

ed255 (approved):
LGTM!  I've added a few small naming suggestions and small simplification, and also saw some leftover prints.  Since these are very small things I'm approving already the PR, please take a look and reply to my requests :)

Thanks for keeping the specs updated and complete!

src/zkevm_specs/evm_circuit/execution/begin_tx.py
```diff
@@ -91,16 +104,94 @@ def begin_tx(instruction: Instruction):
...
+            # Setup next call's context
+
+            for tag, word_or_value in [
+                (CallContextFieldTag.Depth, FQ(1)),
+                (CallContextFieldTag.CallerAddress, tx_caller_address_word),
+                (CallContextFieldTag.CalleeAddress, contract_address_word),
+                (CallContextFieldTag.CallDataOffset, FQ(0)),
+                (CallContextFieldTag.CallDataLength, tx_call_data_length),
+                (CallContextFieldTag.Value, tx_value),
+                (CallContextFieldTag.IsStatic, FQ(True)),
```
KimiWu123, line 171: `IsStatic` should be `False`.

src/zkevm_specs/evm_circuit/execution/begin_tx.py
KimiWu123: I guess this constraint is not necessary because `tx_call_data_length` must be a non-zero value in this `else` condition. Since `calldata` is non-zero data, the `code_hash` would not be equal to `EMPTY_CODE_HASH`.

KimiWu123 (commented):
first pass, have some questions and a minor naming suggestion

KimiWu123 (approved):
LGTM! Nice work!!

PR #439, opened 2023-06-28
refactor `mul_word_by_u64` gadget

TomTaehoonKim:
# Changes
- Added zero equality constraint for `quotient_hi`

- Fix to return with `product_bytes` which was there but unused

# Description
As mentioned in https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/util/math_gadget/mul_word_u64.rs#L15-L16, `mul_word_by_u64` gadget assumes that there will be no overflow.

And this is implicitly checked in [the code below](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/util/math_gadget/mul_word_u64.rs#L50-L53).

However, there is no such constraint inside the `mul_word_by_u64` gadget, but instead it returns the carry and checks outside the gadget. I think this is an inconsistent (with the zkevm-circuits code) implementation, so I made the fix.

# Note
Sorry for the weird branch name. I had the same motivation, but at first, I made a wrong way of change.

### DIFF
--- a/src/zkevm_specs/evm_circuit/instruction.py
+++ b/src/zkevm_specs/evm_circuit/instruction.py
@@ -567,7 +567,7 @@ class Instruction:
 
         return Word((diff_lo, diff_hi)), FQ(borrow_hi)
 
-    def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Tuple[Word, FQ]:
+    def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Word:
         multiplicand_lo, multiplicand_hi = multiplicand.to_lo_hi()
 
         quotient_lo, product_lo = divmod((multiplicand_lo * multiplier.expr()).n, 1 << 128)
@@ -575,9 +575,9 @@ class Instruction:
             (multiplicand_hi * multiplier.expr() + quotient_lo).n, 1 << 128
         )
 
-        product_bytes = product_lo.to_bytes(16, "little") + product_hi.to_bytes(16, "little")
+        self.constrain_zero(FQ(quotient_hi))
 
-        return Word((FQ(product_lo), FQ(product_hi))), FQ(quotient_hi)
+        return Word((FQ(product_lo), FQ(product_hi)))
 
     def mul_add_words(self, a: Word, b: Word, c: Word, d: Word) -> FQ:
         """
--- a/src/zkevm_specs/evm_circuit/execution/begin_tx.py
+++ b/src/zkevm_specs/evm_circuit/execution/begin_tx.py
@@ -52,8 +52,7 @@ def begin_tx(instruction: Instruction):
     # Calculate gas fee
     tx_gas = instruction.tx_context_lookup(tx_id, TxContextFieldTag.Gas)
     tx_gas_price = instruction.tx_gas_price(tx_id)
-    gas_fee, carry = instruction.mul_word_by_u64(tx_gas_price, tx_gas)
-    instruction.constrain_zero(carry)
+    gas_fee = instruction.mul_word_by_u64(tx_gas_price, tx_gas)
 
     # intrinsic gas
     # G_0 = sum([G_txdatazero if CallData[i] == 0 else G_txdatanonzero for i in len(CallData)]) +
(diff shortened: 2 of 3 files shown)


src/zkevm_specs/evm_circuit/instruction.py
```diff
@@ -575,9 +575,11 @@ def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Tuple[W
             (multiplicand_hi * multiplier.expr() + quotient_lo).n, 1 << 128
         )
 
+        self.constrain_zero(FQ(quotient_hi))
+
         product_bytes = product_lo.to_bytes(16, "little") + product_hi.to_bytes(16, "little")
```
ChihChengLiang, line 580: This is now dead code.
```suggestion
```

src/zkevm_specs/evm_circuit/instruction.py
```diff
@@ -575,9 +575,11 @@ def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Tuple[W
...
 
-        return Word((FQ(product_lo), FQ(product_hi))), FQ(quotient_hi)
+        return Word(product_bytes)
```
ChihChengLiang, line 582: The signature of the Word is two cells.
```suggestion
        return Word((FQ(product_lo), FQ(product_hi)))
```

src/zkevm_specs/evm_circuit/instruction.py
```diff
@@ -575,9 +575,11 @@ def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Tuple[W
             (multiplicand_hi * multiplier.expr() + quotient_lo).n, 1 << 128
```
ChihChengLiang, line 575: Can we also fix the function signature?
```diff
-def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Tuple[Word, FQ]:
+def mul_word_by_u64(self, multiplicand: Word, multiplier: Expression) -> Word:
```

ChihChengLiang (approved):
@TomTaehoonKim Thank you so much for the contribution!
Constraining the carry bits to zero makes so much sense. That's a great refactor.
I added some comments on function signatures.

src/zkevm_specs/evm_circuit/instruction.py
TomTaehoonKim: Applied.

src/zkevm_specs/evm_circuit/instruction.py
TomTaehoonKim: Oh, I see. Applied.

src/zkevm_specs/evm_circuit/instruction.py
TomTaehoonKim: Applied.

KimiWu123 (approved):
LGTM! Nice refactor!

src/zkevm_specs/evm_circuit/instruction.py
```diff
@@ -567,17 +567,17 @@ def sub_word(self, minuend: Word, subtrahend: Word) -> Tuple[Word, FQ]:
...
         quotient_lo, product_lo = divmod((multiplicand_lo * multiplier.expr()).n, 1 << 128)
         quotient_hi, product_hi = divmod(
             (multiplicand_hi * multiplier.expr() + quotient_lo).n, 1 << 128
         )
 
-        product_bytes = product_lo.to_bytes(16, "little") + product_hi.to_bytes(16, "little")
+        self.constrain_zero(FQ(quotient_hi))
 
-        return Word((FQ(product_lo), FQ(product_hi))), FQ(quotient_hi)
+        return Word(FQ(product_lo), FQ(product_hi))
```
ChihChengLiang, line 580: I think the test is failing because the function signature is incorrect here.
The first argument is a tuple.
After the argument is fixed, the tests can pass on my local side.
```suggestion
        return Word((FQ(product_lo), FQ(product_hi)))
```

TomTaehoonKim:
@KimiWu123 In my local machine, all tests succeed. Can you try rerunning the test please?

src/zkevm_specs/evm_circuit/instruction.py
TomTaehoonKim: @ChihChengLiang You're correct! Thanks a lot.

@KimiWu123 Fixed!

KimiWu123 (approved):

PR #462, opened 2023-07-10
feat/ implement ErrorWriteProtection error state

### LINKED ISSUE #441: Implement `ErrorWriteProtection` error state
see #365 for more details

KimiWu123:
fixed #441


tests/evm/test_error_write_protection.py
```diff
@@ -0,0 +1,131 @@
...
+    # fixed values
+    callee_address = 0xCAFECAFE
+    gas = 1000
+
+    caller_context = CallContext()
+    caller_bytecode_hash = Word(bytecode.hash())
+
+    is_call = False
+    if opcode is Opcode.CALL:
+        is_call = True
```
ChihChengLiang, line 63: nitpick
```suggestion
    is_call = (opcode is Opcode.CALL)
```

ChihChengLiang (approved):
The spec looks great to me!
I added some description improvements and nitpicks.

specs/error_state/ErrorWriteProtection.md
```diff
@@ -0,0 +1,19 @@
+# ErrorWriteProtection state
+
+## Procedure
+there are some op codes which modify state. there are `[SSTORE, CREATE, CREATE2, CALL, SELFDESTRUCT, LOG0, LOG1, LOG2, LOG3, LOG4]` when execution call context is read only (static call), these op codes running will encounter write protection error. See [EIP-214](https://eips.ethereum.org/EIPS/eip-214)
+
+### EVM behavior
+In above op codes which modify state, `CALL` is somewhat special. For non call op codes (SSTORE, CREATE, etc) first check if running in read only call context. If yes, throw write protection error. For `CALL` op code, it will also check `value` is not zero. Only both non zero value & read only call context, then throw write protection error.
```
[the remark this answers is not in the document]
KimiWu123: > A write protection error is thrown if all the following conditions are met:

The statement `all the following conditions` is not correct. For non-CALL opcodes, it's condition is running in read-only context only. For CALL opcode, it needs to meet both conditions.

I had a fix, [655c871](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/462/commits/655c871f03d8cccb19c80cdb48e44d95e1c72ce2)
or 
adding a `If` in the second condition
```
A write protection error is thrown if all the following conditions are met:
- State modifying opcodes run in a read-only context.
- If the opcode is `CALL` and the `value` argument is non-zero.
```

Which one do you think better?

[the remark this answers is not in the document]
ChihChengLiang: Good idea! Adding if it sounds better.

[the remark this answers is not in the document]
KimiWu123: fixed in [ca71c36](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/462/commits/ca71c36dfaaeab266833e7919c966c2896e3b5eb)

src/zkevm_specs/evm_circuit/execution/error_write_protection.py
```diff
@@ -0,0 +1,73 @@
...
+        + is_log2
+        + is_log3
+        + is_log4,
+        FQ(1),
+    )
+
+    # Spec 2.
+    # current call must be an internal call
+    is_root = instruction.call_context_lookup(CallContextFieldTag.IsRoot)
+    instruction.constrain_equal(is_root, FQ(0))
```
ed255, line 60: Is there a way to have a Root call with static=true?  If that's not possible, maybe this constraint is redundant?

src/zkevm_specs/evm_circuit/execution/error_write_protection.py
```diff
@@ -0,0 +1,73 @@
...
+    # current call context must be readonly
+    is_static = instruction.call_context_lookup(CallContextFieldTag.IsStatic)
+    instruction.constrain_equal(is_static, FQ(1))
+
+    # Spec 3. for `CALL` op code, `value` must be  non-zero
+    if is_call == FQ(1):
+        instruction.stack_pop()
+        instruction.stack_pop()
+        value = instruction.stack_pop()
```
ed255, line 69: You could use `instruction.stack_lookup` here to directly lookup at the 3rd position in the stack (like a peek), and this way you skip 2 extra lookups.

ed255 (approved):
Overall looks good! I've added to minor comments for possible optimizations.  The stack lookup should be quite straightforward.  The one about skipping the root call check may be more tricky to reason about.

src/zkevm_specs/evm_circuit/execution/error_write_protection.py
KimiWu123: fixed in [e81d764](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/462/commits/e81d764ed77cd2a557903a0936f11a5a465f29bc)

src/zkevm_specs/evm_circuit/execution/error_write_protection.py
KimiWu123: I'm not sure I understand it fully. 

It's not possible bcs if it's a root call and static=true, then this case must be handled by `STATICCALL` opcode. If the root call with any one of `[SSTORE, CREATE, CREATE2, CALL, SELFDESTRUCT, LOG0, LOG1, LOG2, LOG3, LOG4]` opcodes, then  `static` must be false and wiil be handled by `[SSTORE, CREATE...]` instead of `error_write_protection`  in the call context .
Aka. any call contexts with static=true at least having a `STATICCALL` ahead. Is that correct?

KimiWu123: fixed in [672f4f9](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/462/commits/672f4f9b9c3e9be3676620e2b6ac7f85e42efb70)

ed255: > Aka. any call contexts with static=true at least having a `STATICCALL` ahead. Is that correct?

I think so!  And since static=true needs a parent `STATICCALL`, then it can't be root (because a root call has no parent callop opcode)

Issue #463, opened 2023-07-11
Fully implement pushn


rrtoledo:
Currently PUSHn is not fully implemented, the case "push out of code range" is not taken care. It should first lookup the bytecode's length and have the data selectors summing to min(length - program_counter, n).

Linked to https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/633

PR #464, opened 2023-07-11
Fully implement PUSHn

rrtoledo:
Linked to issue #463 

------------ Objective
We intend here to support PUSHn with bytecode length inferior to n.

------------ Rational
In pushn, "selectors" are switches to enable or disable lookup arguments on the bytcode bytes as well as to check whether these are non-null if needs be. They directly depend on "n" of PUSHn.
We add to these "selectors", "counters" which have similar role but are a subset of them depending on the bytecode length. We added these extra switches to be able to easily check the various identities at the risk of some redundancy.

To support these changes, modifications in stop.py also had to be done. More precisely, the check "is_out_of_range had to be extended to support small bytecodes.

### DIFF
--- a/src/zkevm_specs/evm_circuit/execution/push.py
+++ b/src/zkevm_specs/evm_circuit/execution/push.py
@@ -1,3 +1,4 @@
+from ...util import N_BYTES_PROGRAM_COUNTER
 from ..instruction import Instruction, Transition
 from ..opcode import Opcode
 
@@ -5,15 +6,19 @@ from ..opcode import Opcode
 def push(instruction: Instruction):
     opcode = instruction.opcode_lookup(True)
     num_pushed = opcode - Opcode.PUSH1 + 1
-    num_additional_pushed = num_pushed - 1
+    code_length = instruction.bytecode_length(instruction.curr.code_hash)
+    code_length_left = code_length - instruction.curr.program_counter - 1
+    is_out_of_bound, _ = instruction.compare(code_length_left, num_pushed, N_BYTES_PROGRAM_COUNTER)
+    num_padding = is_out_of_bound * (num_pushed - code_length_left)
 
     value = instruction.stack_push()
     value_le_bytes = value.to_le_bytes()
-    selectors = instruction.continuous_selectors(num_additional_pushed, 31)
+    is_pushed = instruction.continuous_selectors(num_pushed, 32)
+    is_padding = instruction.continuous_selectors(num_padding, 32)
 
     for idx in range(32):
         index = instruction.curr.program_counter + num_pushed - idx
-        if idx == 0 or selectors[idx - 1] == 1:
+        if is_pushed[idx] * (1 - is_padding[idx]) == 1:
             instruction.constrain_equal(
                 value_le_bytes[idx], instruction.opcode_lookup_at(index, False)
             )
--- a/src/zkevm_specs/evm_circuit/execution/stop.py
+++ b/src/zkevm_specs/evm_circuit/execution/stop.py
@@ -1,4 +1,4 @@
-from ...util import FQ
+from ...util import FQ, N_BYTES_PROGRAM_COUNTER
 from ..instruction import Instruction, Transition
 from ..table import CallContextFieldTag
 from ..execution_state import ExecutionState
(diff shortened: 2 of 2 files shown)


src/zkevm_specs/evm_circuit/execution/push.py
```diff
@@ -6,19 +7,40 @@ def push(instruction: Instruction):
...
         index = instruction.curr.program_counter + num_pushed - idx
-        if idx == 0 or selectors[idx - 1] == 1:
+        if (idx == 0 and not is_out_of_bound) or (idx > 0 and counters[idx - 1] == 1):
             instruction.constrain_equal(
                 value_le_bytes[idx], instruction.opcode_lookup_at(index, False)
             )
-        else:
-            instruction.constrain_zero(value_le_bytes[idx])
+    for idx in range(1, 31):
+        instruction.constrain_zero((1 - counters[idx]) * value_le_bytes[idx + 1])
```
han0110, line 43: Previously `value_le_bytes[idx]` is always constraint by `opcode_lookup_at`, so we could skip the `constrain_zero` for it, but with padding I think we need to check every byte is zero if it's not constraint by `opcode_lookup_at`.

Here it seems `value_le_bytes[0..2]` are always skipped right?

src/zkevm_specs/evm_circuit/execution/stop.py
```diff
@@ -7,9 +7,13 @@
 def stop(instruction: Instruction):
     # Note when transition to STOP, program_counter can only be increased by 1,
     # (JUMP* will always transit to JUMPDEST, then to STOP if any) so when opcode
-    # fetching is out of range, the program_counter must be equal to code_length.
+    # fetching is out of range, the program_counter must be greater than code_length.
```
han0110, line 10:
```suggestion
    # fetching is out of range, the program_counter must be greater or equal than code_length.
```

han0110 (commented):
Overall LGTM! But I feel there is some part under constraint.

Also I feel previous assumption about "first byte always has value" is making things more difficult to understand, and also considering `PUSH0` support in the future, I try to redo it in this way:

```python3
from ...util import N_BYTES_PROGRAM_COUNTER
from ..instruction import Instruction, Transition
from ..opcode import Opcode


def push(instruction: Instruction):
    num_pushed = opcode - Opcode.PUSH1 + 1
    code_length = instruction.bytecode_length(instruction.curr.code_hash)
    code_length_left = code_length - instruction.curr.program_counter - 1
    is_out_of_bound, _ = instruction.compare(code_length_left, num_pushed, N_BYTES_PROGRAM_COUNTER)
    num_padding = is_out_of_bound * (num_pushed - code_length_left)

    value = instruction.stack_push()
    value_le_bytes = value.to_le_bytes()
    is_pushed = instruction.continuous_selectors(num_pushed, 32)
    is_padding = instruction.continuous_selectors(num_padding, 32)

    for idx in range(32):
        index = instruction.curr.program_counter + num_pushed - idx
        if is_pushed[idx] * (1 - is_padding[idx]) == 1:
            instruction.constrain_equal(
                value_le_bytes[idx], instruction.opcode_lookup_at(index, False)
            )
        else:
            instruction.constrain_zero(value_le_bytes[idx])

    instruction.step_state_transition_in_same_context(
        opcode,
        rw_counter=Transition.delta(1),
        program_counter=Transition.delta(1 + num_pushed),
        stack_pointer=Transition.delta(-1),
    )
```

It mainly splits the `selector` into 2 part: `is_pushed` and `is_padding`, where the former represents little-endian bits of `num_pushed`, and the latter represents little-endian bits of `num_padding`, and we could have `selector` as `is_pushed and (not is_padding)` to iterate on `value_le_bytes` to perform `opcode_lookup_at` or `constrain_zero` check. What do you think about this approach?

src/zkevm_specs/evm_circuit/execution/push.py
rrtoledo: That's much better indeed!
I was trying not to increase the degree of the identity but it turned everything more complicated and I even made an error with the values skipped as you saw.

rrtoledo: Updated!

han0110: I see I see, I think it makes complete sense to not increase the degree, there must be smarter way to generate `is_pushed and (not is_padding)` I feel. Also in circuit side the constraint builder will help us to cap it at some value by using some extra intermediate cells, so I think it might be okay here to write in this way.

han0110 (approved):
LGTM! Nice work!

ed255 (approved):
LGTM!

I just realized that on a PUSHN that consumes data past the bytecode length, the actual pushed value can't be used (because the next opcode is a STOP).  So when `is_out_of_bound = True` it doesn't really matter which value we push in the stack, because it will never be read.

Issue #435, opened 2023-06-15
Specify `PUSH0` instruction


ed255:
This is a new opcode from the Shanghai fork. https://eips.ethereum.org/EIPS/eip-3855

KimiWu123:
Scroll's implementation: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1361

PR #471, opened 2023-07-23
Implement `PUSH0` instruction

silathdiir:
Close https://github.com/privacy-scaling-explorations/zkevm-specs/issues/435

zkevm-circuits PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1463

### DIFF
--- a/src/zkevm_specs/evm_circuit/execution/push.py
+++ b/src/zkevm_specs/evm_circuit/execution/push.py
@@ -5,8 +5,6 @@ from ..opcode import Opcode
 
 def push(instruction: Instruction):
     opcode = instruction.opcode_lookup(True)
-    is_push0 = instruction.is_equal(opcode, Opcode.PUSH0)
-
     num_pushed = opcode - Opcode.PUSH0
     code_length = instruction.bytecode_length(instruction.curr.code_hash)
     code_length_left = code_length - instruction.curr.program_counter - 1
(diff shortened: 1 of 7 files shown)


src/zkevm_specs/evm_circuit/execution/push.py
```diff
@@ -16,9 +18,12 @@ def push(instruction: Instruction):
     is_pushed = instruction.continuous_selectors(num_pushed, 32)
     is_padding = instruction.continuous_selectors(num_padding, 32)
 
+    # Constrain stack write value must be zero for PUSH0.
+    instruction.constrain_zero(is_push0 * instruction.sum(value_le_bytes))
```
ed255, line 22: Why is this necessary? When the opcode is PUSH0, then `num_pushed = 0`.
This means that the `is_pushed` selectors will be all zeroes (because `sum(is_pushed) = num_pushed`, from `instruction.continuous_selectors(num_pushed, 32)`.
If is_pushed is all zeroes, then each `value_le_bytes` is 0 from the loop below (which is checking all indexes of value_le_bytes: from 0 to 31).

Perhaps this was assuming that the loop was from 1 to 31?

silathdiir: I see. I try to review this code again, the `value` has already constrained to zero of each byte for `PUSH0`.

I delete the redundant constraint in commit 01fd687. Thanks.

ed255 (approved):
LGTM! Thanks for working on this :)

rrtoledo:
LGTM!

PR #476, opened 2023-07-27
Feat/#447 implement error out of gas account access state error

### LINKED ISSUE #447: Implement `ErrorOutOfGasAccountAccess` state error
### Descirption
For opcodes BALANCE, EXTCODESIZE, EXTCODEHASH, which possibly touches an extra account

### Circuit
https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1349

KimiWu123:
closed #447

### DIFF
--- a/specs/error_state/ErrorOutOfGasAccountAccess.md
+++ b/specs/error_state/ErrorOutOfGasAccountAccess.md
@@ -1,0 +1,41 @@
+# ErrorOutOfGasAccountAccess state
+
+## Procedure
+
+Handle the corresponding out of gas errors for `BALANCE`, `EXTCODESIZE`, and `EXTCODEHASH` opcodes which possibly touches an extra account.
+
+### EVM behavior
+
+For the current `go-ethereum` code, the out of gas error may occur for `constant gas` or `dynamic gas`. The gas cost is calculated as:
+
+```
+gas_cost = constant_gas + dynamic_gas
+```
+
+The constant gas costs are the same for `BALANCE`, `EXTCODESIZE`, and `EXTCODEHASH`.
+
+```
+constant_gas = 0
+```
+
+The calculation of the dynamic gas costs of `BALANCE`, `EXTCODESIZE`, and `EXTCODEHASH` is the same. It depends on cold and warm access.
+
+```
+if is_warm:
+    dynamic_gas = 100
+else:
+    dynamic_gas = 2600
+```
+
+### Constraints
+
+1. Current opcode is one of `BALANCE`, `EXTCODESIZE`, and `EXTCODEHASH`.
+2. Constrain `gas_left < gas_cost`.
+3. Current call must fail.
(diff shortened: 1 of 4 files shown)


specs/error_state/ErrorOutOfGasAccountAccess.md
```diff
@@ -0,0 +1,41 @@
+# ErrorOutOfGasAccountAccess state
+
+## Procedure
+
+Handle the corresponding out of gas errors for `BALANCE`, `EXTCODESIZE`, and `EXTCODEHASH` opcodes which possibly touches an extra account.
```
miha-stopar, line 5: Not sure what "possibly touches" mean? Does it refer only to `EXTCODESIZE`? Are there some special constraints to be added when it touches an extra account?

miha-stopar (approved):
LGTM! I just have one question.

specs/error_state/ErrorOutOfGasAccountAccess.md
KimiWu123: My understanding is all of these three opcodes could touch an external account and we handle an account access [here](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/476/files#diff-67ee79e787a7b6bbf6a48ab2f1dbf05f79b5d5fe69fe615b238277f59eff9f6dR32-R33), 
```rust
    is_warm = instruction.read_account_to_access_list(tx_id, address)
    gas_cost = GAS_COST_WARM_ACCESS if is_warm == FQ(1) else GAS_COST_ACCOUNT_COLD_ACCESS
```

If these opcode access a non-external account, it must be a warm access. So the above snippet is to take care `possibly touches an extra account`.

miha-stopar: Thanks Kimi. I didn't know this refers to the warm access. I think it's somehow misleading, because it's not very clear what "extra" means. I think the part "which possibly touches an extra account" could be simply removed.

Also, I don't think it needs to be warm access in some specific cases - it can be warm or cold, just the price is different then, no?

KimiWu123: I was thinking to revise to "which possibly touches an `external` account". Will it be better?

> Also, I don't think it needs to be warm access in some specific cases

I didn't get it. Could you elaborate more?

> it can be warm or cold, just the price is different then, no?

Yes, just the price differences.

miha-stopar: > I was thinking to revise to "which possibly touches an `external` account". Will it be better?

Hm, I just don't understand what is `external` account?

> 
> > Also, I don't think it needs to be warm access in some specific cases
> 
> I didn't get it. Could you elaborate more?

I was referring to "If these opcode access a non-external account, it must be a warm access."

> 
> > it can be warm or cold, just the price is different then, no?
> 
> Yes, just the price differences.

PR #482, opened 2023-08-14
chore(tables.md): Update an invalid link

hsyodyssey:


hsyodyssey:
Also, the link to tx.py is invalid, but it's clear to me exactly which file this would point to

KimiWu123 (approved):

ChihChengLiang:
> Also, the link to tx.py is invalid, but it's not clear to me which exact file this would point to

I think tx.py is now tx_circuit.py.

PR #485, opened 2023-08-30
Update state-proof.md typos and abbreviation

ltfschoen:


KimiWu123 (approved):
Nice catch and thanks for your contribution! I'm wondering if you could help to make a paragraph into a single line? Every paragraph was broken into several lines and it looks strange to me.

ltfschoen:
> Nice catch and thanks for your contribution! I'm wondering if you could help to make a paragraph into a single line? Every paragraph was broken into several lines and it looks strange to me.

the whole document was already breaking every paragraph into several lines because the original author may have wanted a limited amount of characters per line. my PR just maintains the character limit per line.
if i make the paragraph that i modified into a single line instead of the way it was before, then that wouldn't be consistent with the rest of the document.
the preview is shown here https://github.com/privacy-scaling-explorations/zkevm-specs/pull/485/files?short_path=383bd0a#diff-383bd0a8138e9b40964a4046a8b879413643629705529274f64ba6270fb350ea

ChihChengLiang (approved):
LGTM. Thanks for the fix!
I think the line break is fine. I'll merge the typo fix as it is.
We can talk about fixing the line break in a consistent way in a new issue.

PR #487, opened 2023-09-06
feat/#460 EIP-4895 Validators withdraw from Beacon chain - part 1

KimiWu123:
This PR is part of #460.

I'll separate #460 into two parts, 
1. spec verification
2. python code implementation

This PR is trying to fix the first part: spec verification

Issue #460, opened 2023-07-07
[Shanghai] EIP-4895 Beacon chain push withdrawals as operations


KimiWu123:
[EIP-4895: Beacon chain push withdrawals as operations](https://eips.ethereum.org/EIPS/eip-4895)

KimiWu123:
### Some notes from EIP doc:
This new field for `Withdrawals` is encoded after the existing fields in the execution payload structure and is considered part of the execution payload’s body.
```
execution_payload_body_rlp = RLP([transactions, [], withdrawals])
```
---

This commitment is constructed identically to the transactions root in the existing execution payload header by inserting each withdrawal into a Merkle-Patricia trie keyed by index in the list of withdrawals.
```
assert execution_payload_header.withdrawals_root == compute_trie_root_from_indexed_data(execution_payload.withdrawals)
```
---

`Withdrawals` provide key information from the consensus layer, ([source](https://eips.ethereum.org/EIPS/eip-4895#system-level-operation-withdrawal)):
1. a monotonically increasing `index`, starting from `0`, as a `uint64` value that increments by 1 per withdrawal to uniquely identify each withdrawal
2. the `validator_index` of the validator, as a `uint64` value, on the consensus layer the withdrawal corresponds to
3. a recipient for the withdrawn ether address as a 20-byte value
4. a nonzero amount of ether given in `Gwei` (1e9 wei) as a `uint64` value.

#### Notes
- Recall that the amount is given in units of `Gwei` so a conversion to units of wei must be performed when working with account balances in the execution state.
- This balance change is unconditional and **MUST** not fail.
- This operation has no associated gas costs.

### Code snippets from geth
`Withdrawals` defined in struct [`Block`](https://github.com/ethereum/go-ethereum/blob/b8adb4cb0c4989d138506531ef1966793b658c54/core/types/block.go#L197)
`WithdrawalHash` define in struct [`Header`](https://github.com/ethereum/go-ethereum/blob/b8adb4cb0c4989d138506531ef1966793b658c54/core/types/block.go#L86)

Process withdrawals after all transaction being processed (https://github.com/ethereum/go-ethereum/blob/b8adb4cb0c4989d138506531ef1966793b658c54/core/state_processor.go#L97-L102)

[...]

PR #488, opened 2023-09-12
Feat/#460 EIP-4895 withdrawal from beacon chain - part2

KimiWu123:
the 2nd part of #460.

see part 1, #487

### DIFF
--- a/specs/tables.md
+++ b/specs/tables.md
@@ -41,14 +41,14 @@ NOTE:
 Proved by the withdrawal circuit.
 
 This circuit 
-| 0 Withdrawal ID | 1 Validator ID | 2 Address      | 3 Amount      | 4 MPT root     |
-| -----------     | -------------  | -------------- | ------------- | -------------- |
-| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value{Lo,Hi} | $value{Lo,Hi}  |
+| 0 Withdrawal ID | 1 Validator ID | 2 Address      | 3 Amount      |
+| -----------     | -------------  | -------------- | ------------- |
+| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value{Lo,Hi} |
 
 NOTE:
-- `Amount` is in Gwei, 64 bits
 - `WithdrawalID` is increased monotonically, 64 bits.
 - `ValidatorID`, 64 bits
+- `Amount` is in Gwei, 64 bits
 
 ## `rw_table`
 
--- a/specs/tables.md
+++ b/specs/tables.md
@@ -43,7 +43,7 @@ Proved by the withdrawal circuit.
 This circuit 
 | 0 Withdrawal ID | 1 Validator ID | 2 Address      | 3 Amount      |
 | -----------     | -------------  | -------------- | ------------- |
-| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value{Lo,Hi} |
+| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value |
 
 NOTE:
 - `WithdrawalID` is increased monotonically, 64 bits.
--- a/specs/withdrawal-proof.md
+++ b/specs/withdrawal-proof.md
@@ -23,9 +23,9 @@ For every withdrawal defined as the parameters `(withdrawal_index, validator_ind
 
 1. `withdrawalsData: bytes = rlp([withdrawal_index, validator_index, address, amount])`
 2. `withdrawalDataHash: word = keccak(withdrawalsData)`
-2. `withdrawalsRoot: word = mpt(withdrawalDataHash)`
-3. `withdrawal_index`, `validator_index` and `amount` are all `uint64` values.
-4. `amount_wei = amount * 1e9` and increases validator's balance by `amount_wei`
+3. `withdrawalsRoot: word = mpt(withdrawalDataHash)`
+4. `withdrawal_index`, `validator_index` and `amount` are all `uint64` values.
+5. `amount_wei = amount * 1e9` and increases validator's balance by `amount_wei`
 
 - The rlp encoding of withdrawal parameters will be done using a custom rlp encoding gadget, isolated from the rlp encoding used by the MPT circuit.
 - The keccak hash of rlp encoded data verification will be done in the keccak circuit; the withdrawal circuit will do a single lookup to the keccak table.
--- a/specs/withdrawal-proof.md
+++ b/specs/withdrawal-proof.md
@@ -22,11 +22,13 @@ execution_payload_body_rlp = RLP([transactions, [], withdrawals])
 For every withdrawal defined as the parameters `(withdrawal_index, validator_index, address, amount)` and the circuit verifies the followings:
 
 1. `withdrawalsData: bytes = rlp([withdrawal_index, validator_index, address, amount])`
-2. `withdrawalsRoot: word = mpt(withdrawalsData)`
+2. `withdrawalDataHash: word = keccak(withdrawalsData)`
+2. `withdrawalsRoot: word = mpt(withdrawalDataHash)`
 3. `withdrawal_index`, `validator_index` and `amount` are all `uint64` values.
 4. `amount_wei = amount * 1e9` and increases validator's balance by `amount_wei`
 
-- The rlp encoding of withdrawal parameters will be done using a custom rlp encoding gadget,  isolated from the rlp encoding used by the MPT circuit.
+- The rlp encoding of withdrawal parameters will be done using a custom rlp encoding gadget, isolated from the rlp encoding used by the MPT circuit.
+- The keccak hash of rlp encoded data verification will be done in the keccak circuit; the withdrawal circuit will do a single lookup to the keccak table.
 - The MPT root verification will be done in MPT circuit; the withdrawal circuit do a lookup to the MPT table.
 
 From this information the circuit builds the WithdrawalTable:
@@ -56,8 +58,8 @@ For each withdrawal, the withdrawal circuit also must prepare the key and value 
 
 Each MPT update uses the following parameters:
 
-- Key = `rlp(withdrawal_index)`
-- Value = `rlp([withdrawal_index, validator_index, address, amount])`
+- Key = `withdrawal_index`
+- Value = `keccak(rlp([withdrawal_index, validator_index, address, amount]))`
 - ValuePrev = `0`
 
 NOTE: The MPT proof used for the Withdrawal Trie doesn't need deletion support.
--- a/src/zkevm_specs/evm_circuit/execution/end_block.py
+++ b/src/zkevm_specs/evm_circuit/execution/end_block.py
@@ -20,14 +20,14 @@ from typing import Set
 # C. The number of txs processed by the EVM Circuit match the number of txs in
 # the TxTable
 #
-# D. The number of balance updates due to withdrawal operation in rw_table
-# match the number of withdrawals in the WithdrawalTable
-#
 # As an extra point:
-# E. We need to prove that at least one EndBlock state exists
+# D. We need to prove that at least one EndBlock state exists
 #
 # Also:
-# F. We need to prove that CumulativeGasCost does not exceed the gas limit
+# E. We need to prove that CumulativeGasCost does not exceed the gas limit
+#
+# F. The number of balance updates due to withdrawal operation in rw_table
+# match the number of withdrawals in the WithdrawalTable
 #
 # We prove (A) by constraining the transition rule that after an EndBlock
 # state, only an EndBlock state can follow.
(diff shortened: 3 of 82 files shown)


src/zkevm_specs/evm_circuit/execution/end_block.py
```diff
@@ -85,6 +89,18 @@ def end_block(instruction: Instruction):
...
+    # a valid withdrawal should have non-zero amount
+    total_withdrawals = FQ(
+        len(
+            [
+                wd_row
+                for wd_row in instruction.tables.withdrawal_table
+                if wd_row.amount.expr() != FQ(0)
+            ]
+        )
+    )
```
ChihChengLiang, line 102: I looked into how withdrawal is processed in a consensus client and an execution client. And I think we can safely assume the amount is non-zero. If the withdrawal contains zero amount, then the withdrawal shouldn't pass the consensus client check.

- In consensus client. 
  - It checked the withdrawal root from execution payload is the same as the one in the beacon state.
  - It checks the withdrawals all have amounts greater than 0 (Using [greater than sign](https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#is_fully_withdrawable_validator) in both fully and partially withdrawal cases)

```suggestion
    total_withdrawals = FQ(len(instruction.tables.withdrawal_table))
```

specs/tables.md
```diff
@@ -41,14 +41,14 @@ NOTE:
 Proved by the withdrawal circuit.
 
 This circuit 
-| 0 Withdrawal ID | 1 Validator ID | 2 Address      | 3 Amount      | 4 MPT root     |
-| -----------     | -------------  | -------------- | ------------- | -------------- |
-| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value{Lo,Hi} | $value{Lo,Hi}  |
+| 0 Withdrawal ID | 1 Validator ID | 2 Address      | 3 Amount      |
+| -----------     | -------------  | -------------- | ------------- |
+| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value{Lo,Hi} |
```
ChihChengLiang, line 46: Maybe we dont need a full word for the amount if it is u64
```suggestion
| $WithdrawalID   | $ValidatorID   | $value{Lo,Hi}  | $value |
```

ChihChengLiang (commented):
First round of review

src/zkevm_specs/evm_circuit/execution/end_block.py
```diff
@@ -93,14 +109,32 @@ def end_block(instruction: Instruction):
...
+                    instruction.add_balance(wd.address, [Word(int(wd.amount.expr().n) * int(1e9))])
+                else:
+                    padding_wds += 1
+
+            # 3a. If max_withdrawals == total_withdrawals, we know we have covered all withdrawals
+            # from the withdrawal_table.
+            # If not, we need to check the reset of withdrawals in the table are padding.
+            instruction.constrain_equal(FQ(padding_wds), FQ(max_withdrawals - total_withdrawals))
+
             # 1b. total_txs matches the tx_id that corresponds to the final step.
```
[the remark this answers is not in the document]
KimiWu123: totally agree, the reason why I was doing this way is bcs of this [snippet](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/files#diff-4a935dd7a04687eca5b4d53bfe7993c5d18e0f60a32c7ac0b7bcc4b8d6fb4058L47-L65). If I followed the order 1b, 2 and 3a, then I need to move [this piece](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/files#diff-4a935dd7a04687eca5b4d53bfe7993c5d18e0f60a32c7ac0b7bcc4b8d6fb4058R58-R79) into above snippet and it looks not good. However, you're right your suggestion make code more readable and beside, all the withdrawal operations should be in the end of rw table.

src/zkevm_specs/evm_circuit/execution/end_block.py
KimiWu123: The purpose of checking non-zero `amount` is to tell valid withdrawals and padding withdrawals. Do you think we need to check another field or adding more comment for this?

src/zkevm_specs/evm_circuit/execution/end_block.py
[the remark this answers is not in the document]
KimiWu123: fixed in [c6eca86](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/c6eca865b56259cdba3510742566ae9ae383f097)

specs/tables.md
KimiWu123: updated code in [035586d](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/035586d031230558551ad74baf198fa9169497e8)

src/zkevm_specs/evm_circuit/execution/end_block.py
ChihChengLiang: We should create a tag to indicate if it is a valid withdraw or padding.

KimiWu123: Wondering what is the criteria of having a tag for padding rows? From [tx_circuit implementation](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/src/zkevm_specs/tx_circuit.py#L206), it only checks `address == 0` to determine it's a padding row or not. I just followed tx_circuit implementation.

specs/tables.md
```diff
@@ -41,14 +41,14 @@ NOTE:
 Proved by the withdrawal circuit.
 
 This circuit 
-| 0 Withdrawal ID | 1 Validator ID | 2 Address      | 3 Amount      | 4 MPT root     |
```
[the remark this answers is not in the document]
KimiWu123: From our discussion, we'll keep mpt root, right?

src/zkevm_specs/evm_circuit/execution/end_block.py
hero78119: In tx table there are 3 kinds of transactions
1. valid txs
2. invalid txs (check validity of `balance`, `nonce`, and `intrinsic gas`)
3. padding transaction (address == 0)

case 2 is to handle L2 specific scenarios to introduce extra cost when transaction is invalid. In withdrawl we might safely assume all withdrawal transactions passed from validators are valid therefore we just check padding by amount = 0, since withdrawl is "pushed" from consensus client -> EVM, and L2 who used the same codebase should follow similar protocols.
I think we can follow @KimiWu123 current approach first to avoid introduce extra tag to indicate invalid withdraw. 

However I was wondering is there better field to differentiate padding, e.g. withdrawal_index == 0 or validator_index == 0 other than amount?

specs/tables.md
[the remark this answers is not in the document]
hero78119: Just to double check: Here mpt root is meant `intermediate_mpt_root` right, and keep intermediate_mpt_root used it during lookup to assure MPT insertion sequential? 

I saw all MPT lookup in withdrawl root only provide `key` information, is it enough to assure MPT root insertion maniputation are indeed [insert wd1, insert wd2, ..., insert wdN] without involving extra node in MPT table ?

hero78119 (commented):
First round pass, overall LGTM! just raise few questions

specs/tables.md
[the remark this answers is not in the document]
ChihChengLiang: For @KimiWu123,
> From our discussion, we'll keep mpt root, right?

We don't need to keep mpt roots in the table, because we don't need them in the EVM circuit. 

We do need to keep mpt roots in the withdrawal circuit. Let me summarize our discussion for future reference.

Say we have a tree of 3 withdrawals `{0: wd0, 1: wd1, 2: wd2}` and its root `withdrawal_root`.

The withdrawal circuit should prove the validity of the withdrawal table and the valid table should have `[wd0, wd1, wd2]`.

It is not enough to show `membership_proof(withdrawal_root, 0, wd0)`, `membership_proof(withdrawal_root, 1, wd1)`, and `membership_proof(withdrawal_root, 2, wd2)` in the withdrawal circuit, because prover might skip the last membership proof in the circuit and remove `wd2` from the withdrawal table. 

The sound withdrawal circuit should build the tree incrementally. It starts with empty root `{}`, then insert wd0 to get `{0: wd0}`, then wd1 for `{0: wd0, 1: wd1}`, and finally `{0: wd0, 1: wd1, 2: wd2}`. It shows that the final tree has the tree root `withdrawal_root`. By constructing the circuit this way, a malicious prover can not skip any withdrawal data in the circuit and in the table.

ChihChengLiang (approved):
I added some comments.
Overall LGTM.

src/zkevm_specs/evm_circuit/execution/end_block.py
KimiWu123: > However I was wondering is there better field to differentiate padding, e.g. withdrawal_index == 0 or validator_index == 0 other than amount?

@hero78119 , I was also thinking to use `withdrawal_index == 0`, but `0` is not a invalid index, the index of first withdawal is `0`. If we don't veriy the history of withdrawals, `0` looks fine with me or we check`withdrawal_index == 0 && amount == 0`? Maybe follow tx circuit to check `address == 0`? any of them is fine with me, I don't have any preference.

tests/evm/test_end_block.py
```diff
@@ -12,38 +13,47 @@
...
-    (True, True, 1, 0, True),
-    (True, False, 1, int(15e6), True),
-    (True, False, 1, int(15e6) + 1, False),
+    # (is_last_step, empty_block, max_txs, max_withdrawals, cumulative_gas, success)
+    # (False, False, 2, 5, 0, True),
+    (True, False, 2, 5, 0, True),
+    # (True, False, 1, 2, 0, True),
+    # (True, True, 1, 5, 0, True),
+    # (True, False, 1, 5, int(15e6), True),
+    # (True, False, 1, 2, int(15e6) + 1, False),
```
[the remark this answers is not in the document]
KimiWu123: fixed in [83688b2](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/83688b279d9743c0bb0092160ec11095bd97eb64)

specs/withdrawal-proof.md
```diff
@@ -22,11 +22,13 @@ execution_payload_body_rlp = RLP([transactions, [], withdrawals])
 For every withdrawal defined as the parameters `(withdrawal_index, validator_index, address, amount)` and the circuit verifies the followings:
 
 1. `withdrawalsData: bytes = rlp([withdrawal_index, validator_index, address, amount])`
-2. `withdrawalsRoot: word = mpt(withdrawalsData)`
+2. `withdrawalDataHash: word = keccak(withdrawalsData)`
```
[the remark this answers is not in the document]
KimiWu123: Nice catch! To make it simple I just followed the sequence 1, 2, 3... .
fixed in [ae2fdfa](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/ae2fdfadc2100f18d9d9d0a103d342a512f22aaf)

KimiWu123:
>Just to double check: Here mpt root is meant intermediate_mpt_root right, and keep intermediate_mpt_root used it during lookup to assure MPT insertion sequential?
I saw all MPT lookup in withdrawl root only provide key information, is it enough to assure MPT root insertion maniputation are indeed [insert wd1, insert wd2, ..., insert wdN] without involving extra node in MPT table ?

Yes, it's intermediate mpt root. TBH, I'm not sure but the MPT lookup here is just a workaround since MPT circuit hasn't been merged. Besides, from MPT Rust circuit implementation, the input is the same, `address` ,`storageKey`, `prev_root`, `root`, `prev_value` and `value`. I guess it's enough.

src/zkevm_specs/evm_circuit/execution/end_block.py
KimiWu123: > The approach now makes sense to me. Can we add a comment like "we use amount=0 for padding withdrawals" for clarity?

fixed in [272535e](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/272535eac60f968a161580733890782ff7416c7b)

KimiWu123:
@ChihChengLiang , I made some changes, adding `withdrawal_root` into block_table and check final `withdrawal_root` in withdrawal circuit as we discussed yesterday. Could you help review it again? Those changes are in [fd176ee](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/fd176ee61ef55458c6615ac4286d230028120de5) and [c6714f8](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/488/commits/c6714f804e665ddc476f5ec1268ae48f5fc3a262).

ChihChengLiang:
@KimiWu123 The block_table lookup for withdrawal_root make sense to me. 👍🏽

hero78119 (approved):
LGTM! Thanks @KimiWu123 effort :)

Issue #489, opened 2023-09-14
We're using the wrong field


ed255:
While doing a debug session for PyChiquito we noticed that the field class we're using in the zkevm-specs is the base field of BN256, but it should be the scalar field instead:
```
>>> from py_ecc import bn128
>>> FQ = bn128.FQ
>>> (FQ(-1).n + 1)
21888242871839275222246405745257275088696311157297823662689037894645226208583
```

For reference https://docs.rs/ark-bn254/latest/ark_bn254/
This is the field we're using in Rust https://github.com/privacy-scaling-explorations/halo2curves/blob/6e2ff3853c8fe91300650a733100640dacf313e6/src/bn256/fr.rs#L39 which is the scalar field of BN256
```
>>> 0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001
21888242871839275222246405745257275088548364400416034343698204186575808495617
```

We should fix the python field so that it matches the field where the circuits are evaluated, mainly to avoid any confusion.

ChihChengLiang:
The correct variable is `bn128.curve_order`

```python
>>> from py_ecc import bn128
>>> bn128.curve_order
21888242871839275222246405745257275088548364400416034343698204186575808495617
>>> 
```

It looks like we need an one-line fix at
https://github.com/privacy-scaling-explorations/zkevm-specs/blob/f4d8c8e7c1f77ef4479963ab89f01fd1db18a81a/src/zkevm_specs/util/arithmetic.py#L27

```diff
-class FQ(bn128.FQ):
+class FQ(bn128.curve_order):
```

KimiWu123:
Hi @ed255 and @ChihChengLiang, we have a contributor submitted https://github.com/privacy-scaling-explorations/zkevm-specs/pull/497 for this issue already. I just want to ensure non of you two are working on this. If so, I'll assign this issue to him.

ed255:
> Hi @ed255 and @ChihChengLiang, we have a contributor submitted #497 for this issue already. I just want to ensure non of you two are working on this. If so, I'll assign this issue to him.

I'm not working on this :)

BTW I think the solution that CC shared will not work.
When doing `class FQ(bn128.FQ)` we are creating a class `FQ` with  parent class `bn128.FQ`
But `bn128.curve_order` is not a class!  It's a number, so we can't use it as parent class.

[...]

PR #494, opened 2023-10-16
Sync BufferReaderGadget

NOOMA-42:
**Summary: **
Solve #440

PR #497, opened 2023-10-24
Fix the bug mentioned in #489.

a2468834:


KimiWu123:
HI @a2468834 , thanks for your contribution. 

However, we have some process when sumbitting a PR. 
1. Contributors need to claim a issue and wait for assignment first. You could see an example here, https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1513#issuecomment-1694575688. This is to avoid the issue was taken already to reduce duplicated work.
2. There is no explict rules for PR title but we usally have a short desciption of what this PR is for. Here is another example, https://github.com/privacy-scaling-explorations/zkevm-specs/pull/494.
3. Lastly. it's better to write some description in PR. If the PR is just a simple fix and well-described in the issue, at least write down "closed #issue no.", "fixed #issue no." (resolved is fine as well). You can find it in #494 as well. By doing so, reviewer can know which issue you're addressing and the most important thing is GitHub will link this PR to the issue which makes issues easier to follow and manage.

a2468834:
Thank you @KimiWu123 for inform me of PR rule. 👍
I will read the links and submit another PR again.

PR #508, opened 2023-12-17
Feat/#498 integrate precompiles into callop

### LINKED ISSUE #498: Integrate precompiles into callop
Currently, our precompile gadgets are all individual verification gadgets, but we need to integrate these precompile gadgets into `callop` (bcs precompiles are part of `callop`). Could refer to Scroll's implementation: https://github.com/scroll-tech/zkevm-circuits/blob/2e06a0f4d089c2c01f2f62f44dc4a2ed354f4d28/zkevm-circuits/src/evm_circuit/execution/callop.rs#L315-L523

KimiWu123:
closed #498

### DIFF
--- a/src/zkevm_specs/evm_circuit/execution/callop.py
+++ b/src/zkevm_specs/evm_circuit/execution/callop.py
@@ -235,7 +235,6 @@ def callop(instruction: Instruction):
         PrecompileGadget(
             instruction, call.callee_address, FQ(precompile_return_length), call.cd_length
         )
-
     else:  # precheck is ok and callee has code
         # Save caller's call state
         for field_tag, expected_value in [
--- a/src/zkevm_specs/evm_circuit/execution/callop.py
+++ b/src/zkevm_specs/evm_circuit/execution/callop.py
@@ -171,7 +171,7 @@ def callop(instruction: Instruction):
             (CallContextFieldTag.ReturnDataLength, call.rd_length),
         ]:
             instruction.constrain_equal_word(
-                instruction.call_context_lookup_word(field_tag, RW.Write),
+                instruction.call_context_lookup_word(field_tag, RW.Write, callee_call_id),
                 WordOrValue(expected_value),
             )
 
@@ -193,7 +193,7 @@ def callop(instruction: Instruction):
             (CallContextFieldTag.LastCalleeReturnDataLength, FQ(precompile_return_length)),
         ]:
             instruction.constrain_equal(
-                instruction.call_context_lookup(field_tag, RW.Write, callee_call_id),
+                instruction.call_context_lookup(field_tag, RW.Write),
                 expected_value,
             )
 
--- a/src/zkevm_specs/evm_circuit/execution/callop.py
+++ b/src/zkevm_specs/evm_circuit/execution/callop.py
@@ -206,39 +206,6 @@ def callop(instruction: Instruction):
                 expected_value,
             )
 
-        ### copy table lookup here
-        ### is to rlc input and output to have an easy way to verify data in PrecompileGadget
-
-        # RLC precompile input from memory
-        input_copy_rwc_inc = input_rlc = FQ.zero()
-        if precompile_input_len != FQ(0):
-            input_copy_rwc_inc, input_rlc = instruction.copy_lookup(
-                instruction.curr.call_id,
-                CopyDataTypeTag.Memory,
-                callee_call_id,
-                CopyDataTypeTag.RlcAcc,
-                call.cd_offset,
-                call.cd_offset + precompile_input_len,
-                FQ.zero(),
-                precompile_input_len,
-                input_rwc,
-            )
-
-        # RLC precompile output from memory
-        output_copy_rwc_inc = output_rlc = FQ.zero()
-        if call.is_success == FQ.one() and precompile_return_length != FQ.zero():
-            output_copy_rwc_inc, output_rlc = instruction.copy_lookup(
-                callee_call_id,
-                CopyDataTypeTag.Memory,
-                callee_call_id,
-                CopyDataTypeTag.RlcAcc,
-                FQ.zero(),
-                precompile_return_length,
-                FQ.zero(),
-                precompile_return_length,
-                output_rwc,
-            )
-
         # Verify data copy from precompiles
         return_copy_rwc_inc = FQ.zero()
         if call.is_success == FQ.one() and precompile_return_length != FQ.zero():
@@ -258,12 +225,7 @@ def callop(instruction: Instruction):
         # Give gas stipend if value is not zero
         callee_gas_left += has_value * GAS_STIPEND_CALL_WITH_VALUE
 
-        rwc = (
-            instruction.rw_counter_offset
-            + input_copy_rwc_inc
-            + output_copy_rwc_inc
-            + return_copy_rwc_inc
-        )
+        rwc = instruction.rw_counter_offset + return_copy_rwc_inc
         instruction.step_state_transition_to_new_context(
             rw_counter=Transition.delta(rwc),
             call_id=Transition.to(callee_call_id),
@@ -275,7 +237,7 @@ def callop(instruction: Instruction):
             log_id=Transition.same(),
         )
 
-        PrecompileGadget(instruction, callee_address, input_rlc, output_rlc)
+        PrecompileGadget(instruction, callee_address, precompile_return_length, call.cd_length)
 
     else:  # precheck is ok and callee has code
         # Save caller's call state
--- a/src/zkevm_specs/evm_circuit/execution/callop.py
+++ b/src/zkevm_specs/evm_circuit/execution/callop.py
@@ -1,8 +1,7 @@
-from zkevm_specs.evm_circuit.precompile import Precompile
 from zkevm_specs.evm_circuit.util.call_gadget import CallGadget
 from zkevm_specs.evm_circuit.util.precompile_gadget import PrecompileGadget
 from zkevm_specs.util.hash import EMPTY_CODE_HASH
-from zkevm_specs.util.param import N_BYTES_GAS, N_BYTES_STACK
+from zkevm_specs.util.param import N_BYTES_GAS, N_BYTES_MEMORY_WORD_SIZE, N_BYTES_STACK
 from ...util import FQ, GAS_STIPEND_CALL_WITH_VALUE, Word, WordOrValue
 from ..instruction import Instruction, Transition
 from ..opcode import Opcode
@@ -117,11 +116,8 @@ def callop(instruction: Instruction):
     )
 
     # Make sure the state transition to ExecutionState for precompile if and
-    # only if the callee address is one of precompiles
-    is_zero_address = instruction.is_zero(callee_address)
-    # +1 is for convenience, we don't need to care the 2nd return (which is 'eq' case)
-    is_within_precompiles_addr, _ = instruction.compare(callee_address, Precompile.len() + 1, 2)
-    is_precompile = is_zero_address == FQ.zero() and is_within_precompiles_addr == FQ.one()
+    # only if the callee address is one of precompile
+    is_precompile = instruction.precompile(callee_address)
     instruction.constrain_equal(
         is_precompile, FQ(instruction.next.execution_state in precompile_execution_states())
     )
@@ -156,13 +152,8 @@ def callop(instruction: Instruction):
         )
     # precompiles call
     elif is_precheck_ok and is_precompile == FQ.one():
-        precompile_return_length: FQ = instruction.curr.aux_data[0]
-        precompile_input_len: FQ = instruction.curr.aux_data[1]
-        input_rwc: FQ = instruction.curr.aux_data[5]
-        output_rwc: FQ = instruction.curr.aux_data[6]
-        return_data_rwc: FQ = instruction.curr.aux_data[6]
-
-        min_rd_copy_size = min(precompile_return_length.n, call.rd_length)
+        precompile_return_length = instruction.curr.aux_data[0]
+        min_rd_copy_size = min(precompile_return_length, call.rd_length.n)
 
         # precompiles have on code
         instruction.constrain_equal(no_callee_code, FQ.one())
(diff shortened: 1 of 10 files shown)


KimiWu123:
Hi @roynalnaruto , I'm wondering if you have time to review this. I'm not confident that I understand your implementation fully. It would be nice if you could help reivew this or give a quick overview if you're busy. 

There are two main difference in my PR, 
1. removed RLC copy table lookup. I don't see why we need it bcs we're using word lo/hi as inputs/outputs for individual precompiles. 
2. the constraints in `PrecompileGadget`. I followed your upstream PR #1628 but not sure how to add constraints for EC_ADD, EC_MUL and EC_PAIRING.

KimiWu123:
@roynalnaruto , do you have time to give me some feedback? Thanks. 😅

src/zkevm_specs/evm_circuit/util/precompile_gadget.py
```diff
@@ -0,0 +1,36 @@
...
+        instruction.constrain_equal(instruction.precompile(callee_addr), FQ.one())
+
+        ### precompiles' specific constraints
+        precompile = Precompile(callee_addr)
+        if precompile == Precompile.DATACOPY:
+            # input length is the same as return data length
+            instruction.constrain_equal(precompile_return_len, calldata_len)
+        elif precompile == Precompile.ECRECOVER:
+            # input length is 128 bytes
+            instruction.constrain_equal(calldata_len, FQ(128))
```
ChihChengLiang, line 26: It might be problematic to constrain equal calldata len here.
The precopmile runs successfully when the calldata len is 128 bytes.
But EVM allows us to fill in other calldata len and then fail the precompile run.

For example, in this [playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code=%27jFirsNplace_parameters%20in%20memoryZ456e9aea5e197a1f1af7a3e85a3212fa4049a3ba34c2289b4c860fc0b0c64ef3whash~Y~28wvX2YZ9242685bf161793cc25603c231bc2f568eb630ea16aa137d2664ac8038825608wrX4YZ4f8ae3bd7535248d0bd448298cc2e2071e56992d0774dc340c368ae950852adawsX6YqqjDo_call~32JSizeX80JOffsetX8VSize~VOffset~1waddressW4QFFFFFFFFwgasqSTATICCALLqqjPut_resulNalonKon_stackqPOPX80qMLOAD%27~W1%20w%20jq%5Cnj%2F%2F%20_%20thKZW32QY0qMSTOREX~0xWqPUSHV0wargsQ%200xNt%20Ke%20Jwret%01JKNQVWXYZ_jqw~_) I tried to change the argsSize from 0x80 to 0x7f and the precompile fails. That suggests the wrong call data length is still allowed in EVM.

It might take much effort to fix the issue in this PR; we can consider fixing it later.

ChihChengLiang (commented):
Great work implementing the spec for the precompiles call. I left some comments.

src/zkevm_specs/evm_circuit/util/precompile_gadget.py
KimiWu123: TBH, I'm not pretty sure what constraints we need here. From [Scroll's PR](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1628/files#diff-a5cfbb05b952ef55b7e2fbb8b88a437aa6fd29b9cc5fac042011cb645541e831R54-R63), it only checks  the equality of `precompile_return_len` and `calldata_len`. I doubt that we need any constraints for `ecRecover` or we need to constrain the data consistency between the inputs in each precompile and the data in `callop`. (In Scroll codebase, they're using rlced data and they [check rlced data consistency in precompile_gadget](https://github.com/scroll-tech/zkevm-circuits/blob/57a2dd4c4f8b0443a4cc9ee43c71be649c937ffc/zkevm-circuits/src/evm_circuit/util/precompile_gadget.rs#L58-L80)).

An issue created, https://github.com/privacy-scaling-explorations/zkevm-specs/issues/509

src/zkevm_specs/evm_circuit/execution/callop.py
```diff
@@ -147,7 +150,92 @@ def callop(instruction: Instruction):
...
+            (CallContextFieldTag.LastCalleeReturnDataOffset, FQ.zero()),
+            (CallContextFieldTag.LastCalleeReturnDataLength, FQ(precompile_return_length)),
+        ]:
+            instruction.constrain_equal(
+                instruction.call_context_lookup(field_tag, RW.Write, callee_call_id),
+                expected_value,
+            )
+
+        # Verify data copy from precompiles
+        return_copy_rwc_inc = FQ.zero()
```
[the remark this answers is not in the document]
KimiWu123: Yep, I didn't forget this part and I added the copy of input_data and output_data. However, I still feel it's not right so I removed it.

My arguments, 
A. We didn't use any rlced data in our precompiles (e.g. `ecRecover`, `datacopy` ...etc). We use `word` type for our input, output and returned data directly in our precompiles.
B. If we add the copy of `input_data` and `output_data`, we should have some constraints for those rlced data (For example, constrain the equality of rlced data in `precompile_gadget` or in every precompile). But it doesn't make sense to me because we didn't use rlced data in our implementation. 

A question, do we need to constrain the equality of the data in `callop` and the data in individual precompile gadget? I guess the answer is yes. But how do we constrain the data consistency if it's non-fixed length? I don't have any solutions for now.

Conclusion,
1. keep the copy of input_data and output_data and add data consistency in `precompile_gadget` and every precompiles.
2. we don't do the copy since we didn't use rlc here but we need to figure out a way to constrain data consistency

Does it make sense to you?

src/zkevm_specs/evm_circuit/execution/callop.py
```diff
@@ -147,7 +150,92 @@ def callop(instruction: Instruction):
...
+            (CallContextFieldTag.IsSuccess, call.is_success),
+            (CallContextFieldTag.CalleeAddress, callee_address_word),
+            (CallContextFieldTag.CallerId, instruction.curr.call_id),
+            (CallContextFieldTag.CallDataOffset, call.cd_offset),
+            (CallContextFieldTag.CallDataLength, call.cd_length),
+            (CallContextFieldTag.ReturnDataOffset, call.rd_offset),
+            (CallContextFieldTag.ReturnDataLength, call.rd_length),
+        ]:
+            instruction.constrain_equal_word(
+                instruction.call_context_lookup_word(field_tag, RW.Write),
```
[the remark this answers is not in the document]
KimiWu123: Nice catch! I set wrong call id for both caller and callee and fixed in [46f951e](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/508/commits/46f951ebea16fb708a59abccbfa7bd27a035d382)

src/zkevm_specs/evm_circuit/execution/callop.py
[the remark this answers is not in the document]
miha-stopar: Don't we need to ensure that the output is properly copied to the return data buffer?

I think the only way to ensure the data consistency if it's not fixed length is to use RLC.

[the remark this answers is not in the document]
ChihChengLiang: I may have missed something, but I see no reason why the following doesn't work.

1. Copy with RLC to constrain the
  1.  Caller's memory for input === precompile input RLC
  2. Caller's memory for output === precompile output RLC
2. Parse the precopmile input RLC with words if the input is a word. We perform the same parsing for the precompile output RLC.

src/zkevm_specs/evm_circuit/util/precompile_gadget.py
KimiWu123: fixed in [3cf679c](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/508/commits/3cf679c68644488840d756cd3dadac1f2ce1f293), but I guess the return length you were saying is `32` instead of 128, right.

src/zkevm_specs/evm_circuit/execution/callop.py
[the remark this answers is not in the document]
KimiWu123: @ChihChengLiang, it makes sense to me. Could you check [14a38d4](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/508/commits/14a38d42df1fd14fa7745949d4a4741c7f9f2cc1) to see if it fits your point 4?

src/zkevm_specs/evm_circuit/util/precompile_gadget.py
miha-stopar: Ah yes!

src/zkevm_specs/evm_circuit/execution/callop.py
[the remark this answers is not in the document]
ChihChengLiang: @KimiWu123 [b5816af](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/508/commits/b5816afe36977ad1b665848e5e47107f2254e5c8) The fix here is exactly what I was looking for.
The parsing in [f3eee6f](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/508/commits/f3eee6f6f45b079af55251945d59e20a0d517b58) looks correct to me too!
Thanks for the fix!

ChihChengLiang (approved):
LGTM! Thanks for the epic work!

src/zkevm_specs/evm_circuit/execution/precompiles/ecrecover.py
```diff
@@ -26,9 +34,25 @@ def ecRecover(instruction: Instruction):
...
     is_recovered = FQ(instruction.is_zero(recovered_addr) != FQ(1))
 
+    # Verify input and output
+    input_bytes = bytearray(b"")
+    input_bytes.extend(msg_hash.int_value().to_bytes(32, "little"))
+    input_bytes.extend(sig_v.int_value().to_bytes(32, "little"))
+    input_bytes.extend(sig_r.int_value().to_bytes(32, "little"))
+    input_bytes.extend(sig_s.int_value().to_bytes(32, "little"))
+    input_rlc = RLC(bytes(reversed(input_bytes)), keccak_randomness, n_bytes=128).expr()
+    instruction.constrain_equal(rlc_data.input_rlc, input_rlc)
```
miha-stopar: Aren't these two values actually the same (I mean not equal, but the same)? Because `rlc_data.input_rlc` is computed when [instantiating the auxiliary data](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/2e5f9263b74ced37ad3d9aea02874ed90bb85bb4/tests/evm/precompiles/test_ecRecover.py#L86) and `input_rlc` is computed accessing the same values in `aux_data`?

KimiWu123: input_rlc = RLC(msg_hash, sig_v, sig_r, sig_s)

For me, data assignment in testing is like what `assign_exec_step` does at proving time in our circuit. So, yes, it looks the same (but not always the same). 
At proving time, a prover assigns
- msg_hash
- sig_v, sig_r, sig_s
- RLC(msg_hash, sig_v, sig_r, sig_s) (a.k.a `input_rlc` here, this field is to verify data consistency between calls)

In verification logic here, we have to gaurantee msg_hash, sig_v, sig_r and sig_s are the same pairs (sig_v, sig_r and sig_s are coming from the signature of msg_hash). However, a malious signer could sign msg_hash2 and have sig_v2, sig_r2 and sig_s2. It still can pass all the constraints if we don't have calculate input_rlc here. I'm assuming `rlc_data.input_rlc` is coming from previous step (we can pass data around between previous step and next step in our circuit, but it seems not doable in our spec), so that's why `rlc_data.input_rlc` and `input_rlc` look "the same".

Does it make sense to you? Or do you think we need to copy `rlc_data.input_rlc` from copy_table?

miha-stopar: I think I am missing something and would like to fix my understanding :). My understanding was that we need to ensure that `instruction.curr.aux_data` is the same as used in `callop.py` and that this should be done by checking the RLC, so to have a lookup call into `copy_table` (RLC field)  here in `ecrecover.py`. Maybe this is not the case?

I agree on the assignment part, putting a link to the assignment was probably misleading from my side. What it seems to me here is that we have `aux_data: PrecompileAuxData = instruction.curr.aux_data[0]` and then both values, `aux_data.input_rlc` and `input_rlc`, are computed using `aux_data`. Or is it `aux_data.input_rlc` checked somewhere else?

KimiWu123: > My understanding was that we need to ensure that `instruction.curr.aux_data` is the same as used in `callop.py` and that this should be done by checking the RLC, so to have a lookup call into `copy_table` (RLC field) here in `ecrecover.py`. Maybe this is not the case?

Yes, that's exactly what I want to do.

> I agree on the assignment part, putting a link to the assignment was probably misleading from my side. What it seems to me here is that we have `aux_data: PrecompileAuxData = instruction.curr.aux_data[0]` and then both values, `aux_data.input_rlc` and `input_rlc`, are computed using `aux_data`. Or is it `aux_data.input_rlc` checked somewhere else?

No, `aux_data.input_rlc` was not check in other place.

In my implementation, I treated `aux_data` as witness inputs, a place I can assign my witnesses. Those data (e.g. sig_r, sig_v...etc) can't be retrieved from stack or memory like what other opcode gagdets implemented. 

I might not explain it very well, but you can check [Scroll's impl.](https://github.com/scroll-tech/zkevm-circuits/blob/57a2dd4c4f8b0443a4cc9ee43c71be649c937ffc/zkevm-circuits/src/evm_circuit/execution/precompiles/ecrecover.rs#L121-L133) The `msg_hash_raw` was converted into rlc formate and compared with `msg_hash_keccak_rlc`. In the assignment, you can see both of them come from the same source, [here](https://github.com/scroll-tech/zkevm-circuits/blob/57a2dd4c4f8b0443a4cc9ee43c71be649c937ffc/zkevm-circuits/src/evm_circuit/execution/precompiles/ecrecover.rs#L372-L380) and [here](https://github.com/scroll-tech/zkevm-circuits/blob/57a2dd4c4f8b0443a4cc9ee43c71be649c937ffc/zkevm-circuits/src/evm_circuit/execution/precompiles/ecrecover.rs#L404-L412).

miha-stopar (approved):
Great work!

PR #522, opened 2024-04-09
MCOPY spec

curryrasul:


curryrasul:
It's not directly related to `MCOPY`, but I'm working on this, and looking through `RETURNDATACOPY`, which we discussed before @ChihChengLiang 
It is said, that `RETURNDATACOPY` copies from memory to memory: https://github.com/privacy-scaling-explorations/zkevm-specs/blob/4f272c2b6ddfaf2b0815d1b93a6b32e38f5d8fb8/specs/tables.md?plain=1#L276-L279

I read about `RETURNDATACOPY`, it's said that it copies from so-called `return data buffer`, that acts like a temporary place, where's the return value from contract-call is stored. 
Is it still memory? Or maybe it doesn't belong to memory , stack or storage?

specs/opcode/94MCOPY.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Constraints
+
+1. OpcodeId check
+
+- opId == OpcodeId(0x5E)
+
+2. State Transition
+
+- rw_counter += 3 stack_reads + ???
```
curryrasul, line 21: Not sure here, need help ...

specs/opcode/94MCOPY.md
```diff
@@ -0,0 +1,40 @@
...
+- stack_pointer += 3
```
curryrasul, line 22: I think it should be 3 here, at least because it's 3 in `RETURNDATACOPY` or in `CALLDATACOPY`.
But in the (WIP) scroll impl of MCOPY it's 4:
https://github.com/scroll-tech/zkevm-circuits/pull/1209/files#diff-8f22d99f71cb4ed2a7ed91d2d25ace314f20d903a8e32af87036b12c4d9f4b27R94

zemse:
> I read about RETURNDATACOPY, it's said that it copies from so-called return data buffer, that acts like a temporary place, where's the return value from contract-call is stored.
Is it still memory? Or maybe it doesn't belong to memory , stack or storage?

I was confused about returndata not present in [CopyDataType](https://github.com/zemse/zkevm-circuits/blob/b309ed57c6ae05a03f3f74316e98e167b81844af/bus-mapping/src/circuit_input_builder/execution.rs#L212), but it seems returndatacopy copies from the memory of [child call's memory](https://github.com/zemse/zkevm-circuits/blob/b309ed57c6ae05a03f3f74316e98e167b81844af/zkevm-circuits/src/evm_circuit/execution/returndatacopy.rs#L121) to the current call's memory. So this memory to memory copy of returndatacopy makes sense to me now and mcopy is pretty much similar apart from it is a memory to memory copy for the same call.

specs/opcode/94MCOPY.md
rrtoledo: Except for the stack reads I do not see other RW since you'd use the copy table afterwards. This is confirmed in Scroll WIP implementation

`//rw_counter: Transition::Delta(3.expr()),`

specs/opcode/94MCOPY.md
rrtoledo: I see 3 on scroll too though?

`            stack_pointer: Transition::Delta(3.expr()),`

curryrasul: It was 4 before, probably a typo; Thanks!

